PR #55, opened 2023-02-10
Added WebHook examples for first workflow tests

matt-aitken:
When you do your first test of a WebHook it will use an example JSON payload that we have added.


matt-aitken:
Also fixed some of the GitHub webhook schemas as the examples were failing validation

PR #69, opened 2023-02-23
feat: added OpenTelemetry instrumentation

nirga:
Instruments telemetries from all 3 backend services. 
I'm basing the instrumentation on [Traceloop's image](https://github.com/traceloop/jest-opentelemetry/blob/main/packages/instrument-opentelemetry/src/tracing.ts) which output more data per span that can be used to generate tests later on:
* HTTP Request Headers
* HTTP Body
* Propagation of `trace-loop-id` which is in use by tests


ericallam:
Closing this for now but will need to pick it back up for v2

PR #73, opened 2023-02-28
Updated local instructions

jessie-young:


matt-aitken:
This is no longer relevant in the new version

PR #77, opened 2023-03-06
New integration system: support for webhooks

matt-aitken:
The new integrations system now has support for webhooks.

It currently supports
- Subscribing to automatic webhooks (where the service has an API for registering a webhook).
- Receiving webhooks and sending them on to the requested destination with built-in resiliency (we retry sending them with exponential back-off)
- Generating the Trigger.dev SDK

Typeform webhooks have been added as part of this PR as a proof of concept.

Coming soon
- Support for API webhooks that required manual registration (where a user needs to go fill in a form with a webhook URL).
- Support for "generic webhooks", which doesn't use an integration.


apps/integrations/src/core/webhook/credentials.ts
```diff
@@ -0,0 +1,85 @@
...
+    return cachedClient as InstanceType<typeof Nango>;
+  }
+
+  cachedClient = new Nango({
+    host: process.env.NANGO_HOST,
+    secretKey: process.env.NANGO_SECRET_KEY,
+  });
+  return cachedClient as InstanceType<typeof Nango>;
+}
+getNangoClient();
```
matt-aitken, line 19: @ericallam I had to do this because Nango doesn't have CJS at the moment. I've requested it.

PR #170, opened 2023-07-14
chore: allow job loader to use JobPresenter class

### LINKED ISSUE #155: [TRI-573] Move job loading to the jobs loader
By fetching jobs too high up in the route hierarchy, navigating to the job list page does not pick up on new jobs created since the last time the jobs were fetched. This is because of how remix works and so we need to move all fetching logic to the correct places

Currently, jobs are fetched in the ProjectPresenter and should be moved to their own presenter (JobListPresenter). The route that currently fetches the jobs is here:

[https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/routes/\_app.orgs.$organizationSlug.projects.$projectParam/route.tsx](https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/routes/\_app.orgs.$organizationSlug.projects.$projectParam/route.tsx)

But they should be moved to here:

[https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/routes/\_app.orgs.$organizationSlug.projects.$projectParam.\_index/route.tsx](https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/routes/\_app.orgs.$organizationSlug.projects.$projectParam.\_index/route.tsx)

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-573](https://linear.app/triggerdotdev/issue/TRI-573/move-job-loading-to-the-jobs-loader)</sub>

kaf-lamed-beyt:
fixes #155 

@ericallam, kindly take a look at this and let me know if I'm on the right track. I also included a Docker setup guide for people like me who may want to contribute to Trigger in the future and could be blocked by the Docker setup.

/claim #155

### DIFF
--- a/apps/webapp/app/presenters/JobPresenter.server.ts
+++ b/apps/webapp/app/presenters/JobPresenter.server.ts
@@ -1,0 +1,22 @@
+import { PrismaClient, prisma } from "~/db.server";
+import { Job } from "~/models/job.server";
+
+export class JobsListPresenter {
+  #prismaClient: PrismaClient;
+
+  constructor(prismaClient: PrismaClient = prisma) {
+    this.#prismaClient = prismaClient;
+  }
+
+  public async getProjectJob(
+    projectSlug: string,
+    jobId: string
+  ): Promise<Job | null> {
+    return this.#prismaClient.job.findFirst({
+      where: {
+        projectId: projectSlug,
+        id: jobId,
+      },
+    });
+  }
+}
--- a/apps/webapp/app/presenters/JobPresenter.server.ts
+++ b/apps/webapp/app/presenters/JobPresenter.server.ts
@@ -12,15 +12,31 @@ export class JobsListPresenter {
   public async call({
     slug,
     userId,
-  }: {
-    slug: string;
-    userId: string;
-  }): Promise<Job[]> {
-    return this.#prismaClient.job.findMany({
+  }: Pick<Job, "slug"> & { userId: User["id"] }) {
+    const project = await this.#prismaClient.project.findFirst({
       where: {
-        projectId: slug,
-        id: userId,
+        slug,
+        organization: {
+          members: {
+            some: {
+              userId,
+            },
+          },
+        },
+      },
+      select: {
+        jobs: {
+          where: {
+            internal: false,
+          },
+        },
       },
     });
+
+    if (!project) {
+      return undefined;
+    }
+
+    return project.jobs.filter((job) => !job.internal);
   }
 }
--- a/apps/webapp/app/presenters/JobPresenter.server.ts
+++ b/apps/webapp/app/presenters/JobPresenter.server.ts
@@ -1,3 +1,4 @@
+import { User } from "~/models/user.server";
 import { PrismaClient, prisma } from "~/db.server";
 import { Job } from "~/models/job.server";
 
@@ -8,14 +9,17 @@ export class JobsListPresenter {
     this.#prismaClient = prismaClient;
   }
 
-  public async getProjectJob(
-    projectSlug: string,
-    jobId: string
-  ): Promise<Job | null> {
-    return this.#prismaClient.job.findFirst({
+  public async call({
+    slug,
+    userId,
+  }: {
+    slug: string;
+    userId: string;
+  }): Promise<Job[]> {
+    return this.#prismaClient.job.findMany({
       where: {
-        projectId: projectSlug,
-        id: jobId,
+        projectId: slug,
+        id: userId,
       },
     });
   }
--- a/apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam._index/route.tsx
+++ b/apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam._index/route.tsx
@@ -27,6 +27,46 @@ import {
   trimTrailingSlash,
 } from "~/utils/pathBuilder";
 import { BreadcrumbLink } from "~/components/navigation/NavBar";
+import { requireUserId } from "~/services/session.server";
+import { LoaderArgs } from "@remix-run/server-runtime";
+import invariant from "tiny-invariant";
+import { typedjson } from "remix-typedjson";
+import { analytics } from "~/services/analytics.server";
+import { JobsListPresenter } from "~/presenters/JobPresenter.server";
+
+export const loader = async ({ request, params }: LoaderArgs) => {
+  const { projectParam, jobId } = params;
+  const userId = await requireUserId(request);
+  invariant(projectParam, "projectParam not found");
+  invariant(jobId, "jobId not found");
+
+  try {
+    const jobsListPresenter = new JobsListPresenter();
+    const jobs = await jobsListPresenter.getProjectJob(projectParam, jobId);
+
+    if (!jobs) {
+      throw new Response("Not found", {
+        status: 404,
+        statusText: `Project with this ${projectParam} was not found in your Organization`,
+      });
+    }
+
+    return typedjson({
+      jobs,
+    });
+  } catch (error) {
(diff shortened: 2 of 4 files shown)


ericallam:
Wow, those docker instructions are great! Going to start a review shortly 👍

apps/webapp/app/presenters/JobPresenter.server.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Job } from "~/models/job.server";
+
+export class JobsListPresenter {
+  #prismaClient: PrismaClient;
+
+  constructor(prismaClient: PrismaClient = prisma) {
+    this.#prismaClient = prismaClient;
+  }
+
+  public async getProjectJob(
```
ericallam: This presenter should return a list of Jobs for the project. The params passed in here would probably need to be both the project slug and the user id (see here: https://github.com/triggerdotdev/trigger.dev/blob/2d722b94bd448e39506f0d6c9fff3c83e373ef2b/apps/webapp/app/presenters/ProjectPresenter.server.ts#L19)

This should also be renamed `call` and do basically the same query for jobs as here: 

https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/presenters/ProjectPresenter.server.ts#L33

But without all the project stuff (so calling `prisma.job.findMany`)

kaf-lamed-beyt:
Ay! Thanks for the feedback! I look forward to the review.

ericallam (changes requested):

apps/webapp/app/presenters/JobPresenter.server.ts
kaf-lamed-beyt: Ah, Nice! I thought so too. But, I was skeptical because I thought the presenter would get just a project that matches the current job id.

kaf-lamed-beyt: Kindly take a look now @ericallam

ericallam (changes requested):
Can you try and actually run the webapp to make sure this works as intended?

kaf-lamed-beyt:
I've actually tried testing it before. But, I do not know what to test for exactly as I got stuck at the point where I entered my email address.

When I clicked on the "email address" option as a sign-in method, I entered my email address and clicked the button. But I did not get any link in the terminal that'll allow me login.

Here's an image of what I'm seeing in my terminal 👇🏼 

image

And when I try to copy the last URL in the terminal and paste it into my browser, I get this response

image

Perhaps there's something I'm missing.

ericallam:
Make sure you don't have a value set for the RESEND_API_KEY environment variable

kaf-lamed-beyt:
I'll ensure this, now.

kaf-lamed-beyt:
I created a test-project that does not have any job running. Does this feel right?

image

apps/webapp/app/presenters/JobPresenter.server.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  public async call({
+    slug,
+    userId,
+  }: {
+    slug: string;
+    userId: string;
+  }): Promise<Job[]> {
+    return this.#prismaClient.job.findMany({
```
ericallam: You are querying here for many jobs but you are trying to find the job by the user ID. The query here should return the same exact type that is being returned from the ProjectPresenter.call, scoped to just the jobs.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam._index/route.tsx
```diff
@@ -27,6 +27,49 @@ import {
...
+import invariant from "tiny-invariant";
+import { typedjson } from "remix-typedjson";
+import { analytics } from "~/services/analytics.server";
+import { JobsListPresenter } from "~/presenters/JobPresenter.server";
+
+export const loader = async ({ request, params }: LoaderArgs) => {
+  const { projectParam, jobId } = params;
+  const userId = await requireUserId(request);
+  invariant(projectParam, "projectParam not found");
+  invariant(jobId, "jobId not found");
```
ericallam: There is no jobId in the params in this route

kaf-lamed-beyt: Yay! I figured it out! 

There shouldn't be a `jobId` param. 

Formerly it was something like this below: 

```tsx
export const loader = async ({ request, params }: LoaderArgs) => {
  const { projectParam, jobId } = params;
  const userId = await requireUserId(request);
  invariant(projectParam, "projectParam not found");
  invariant(jobId, "jobId not found");
```

But, I changed it to this:

```tsx
export const loader = async ({ request, params }: LoaderArgs) => {
  const { projectParam } = params;
  const userId = await requireUserId(request);
  invariant(projectParam, "projectParam not found");
  invariant(userId, "jobId not found");
```  
 Now the UI renders correctly.
  
  
image

kaf-lamed-beyt: Am I on the right track?

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam._index/route.tsx
```diff
@@ -27,6 +27,49 @@ import {
...
+import invariant from "tiny-invariant";
+import { typedjson } from "remix-typedjson";
+import { analytics } from "~/services/analytics.server";
+import { JobsListPresenter } from "~/presenters/JobPresenter.server";
+
+export const loader = async ({ request, params }: LoaderArgs) => {
+  const { projectParam } = params;
+  const userId = await requireUserId(request);
+  invariant(projectParam, "projectParam not found");
+  invariant(userId, "jobId not found");
```
ericallam: You can remove this `invariant`, since `requireUserId` always returns a string

apps/webapp/app/presenters/JobPresenter.server.ts
```diff
@@ -0,0 +1,26 @@
...
+      where: {
+        projectId: slug,
+        id: userId,
```
ericallam: This is still not right

> You are querying here for many jobs but you are trying to find the job by the user ID. The query here should return the same exact type that is being returned from the ProjectPresenter.call, scoped to just the jobs.

ericallam (changes requested):

apps/webapp/app/presenters/JobPresenter.server.ts
kaf-lamed-beyt: Dang! I must have missed this. I'll fix now.

kaf-lamed-beyt: Just to be clear again, I'll have to follow this pattern from the ProjectPresenter and use `findFirst`, yeah?

```ts
 public async call({
    userId,
    slug,
  }: Pick<Project, "slug"> & {
    userId: User["id"];
  }) {
 ```

Comment (kaf-lamed-beyt) [reply apps/webapp/app/presenters/JobPresenter.server.ts]: For now, I'm only returning jobs that are not internal. 

```ts
 return project.jobs.filter((job) => !job.internal);
```  
Is it okay this way?

ericallam:
@kaf-lamed-beyt I appreciate your work on this PR but I'm going to close this now as I think we're not really making great progress and this might be better handled by someone else in the community. I'm going to tip you so as to make sure there are no hard feelings. It would also be great if you could open a separate PR for your documentation updates as I would really love to get those in the repo.

/tip $50

kaf-lamed-beyt:
Sorry that I couldn't get a headway. 😞 

I'll open that PR now and call your attention to it.

Issue #166, opened 2023-07-13
[TRI-654] Add an `update` sub-command the @trigger.dev/cli that updates all @trigger.dev/* packages


ericallam:
It's annoying currently updated `@trigger.dev/*` packages to the latest ones so we should have a command that could be run like so:

```
$ npx @trigger.dev/cli update
```

This should find all the `@trigger.dev/*` packages in the package.json of the current project and use the NPM registry API to figure out if there are newer versions. It should display which ones are newer and ask for confirmation before updating the package.json and re-installing dependencies.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-654](https://linear.app/triggerdotdev/issue/TRI-654/add-an-update-sub-command-the-triggerdevcli-that-updates-all)</sub>

zodman:
/attempt #166

what about integrate the npm-check-updates like a dependency ?

rishi-raj-jain:
@ericallam Can I work on this one?

PR #174, opened 2023-07-15
[TRI-654] Add an update sub-command the @trigger.dev/cli that updates all @trigger.dev/* packages

omernave:
/claim #166


packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,98 @@
+import axios from "axios";
+import inquirer from "inquirer";
+import semver from "semver";
+import { readFileSync, writeFileSync } from 'fs';
+import { logger } from "../utils/logger.js";
+import { execSync } from "child_process";
+
+export async function updateCommand() {
+    try {
+        const installedPackages = JSON.parse(execSync('npm list --depth=0 --json').toString());
```
ericallam: Can you do this in a package-manager agnostic way to support `npm`, `yarn`, and `pnpm`? You can see we do something similar here:

https://github.com/triggerdotdev/trigger.dev/blob/2d722b94bd448e39506f0d6c9fff3c83e373ef2b/packages/cli/src/utils/installDependencies.ts#L7

Also could you use a promise version of `exec` instead of `execSync`? If there is one in the Node.js standard library then don't worry about it.

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,98 @@
...
+        const triggerDevPackages = Object.keys(installedPackages.dependencies).filter(pkg => pkg.startsWith('axios'));
```
ericallam: I think this should probably be `pkg.startsWith("@trigger.dev/")`, right?

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (Object.keys(updates).length === 0) {
+            console.log('No updates available.');
+            return;
+        }
+
+        const confirm = await confirmUpdate(updates);
+
+        if (confirm) {
+            const updatedPackageJson = updatePackageJson(updates);
+            writeFileSync('package.json', JSON.stringify(updatedPackageJson, null, 2));
```
ericallam: Can you use `writeJSONFile` from the `utils/fileSystem` file?

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,98 @@
...
+            logger.info('package.json updated.');
+
+            // Run 'npm install' to install the updated packages
+            logger.info('Running npm install...');
+            execSync('npm install', { stdio: 'inherit' });
```
ericallam: You can you the `installDependencies` linked above for this

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,98 @@
...
+            logger.info('Dependencies updated.');
+        } else {
+            logger.info('Update canceled.');
+        }
+    } catch (error) {
+        logger.error('An error occurred:', error);
+    }
+}
+
+async function getLatestVersions(packageNames: string[]): Promise<{ [packageName: string]: string }> {
```
ericallam: Some of our packages aren't currently using the "latest" tag (like `@trigger.dev/sdk`, `@trigger.dev/github`, `@trigger.dev/slack` and `@trigger.dev/resend`) but instead are using the `next` tag. Could this code be updated to update packages that are tagged with `next` to either:

1. The more latest `next` version, if it exists
2. The more latest `latest` version, if it exists and is considered "later" in semver standards

ericallam (changes requested):

packages/cli/src/commands/update.ts
omernave: switched to execa

packages/cli/src/commands/update.ts
omernave: fixed

packages/cli/src/commands/update.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: switched to exec

packages/cli/src/commands/update.ts
omernave: Can you review the current logic to see that i understood correctly?

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,128 @@
...
+        const packageName = packageNames[index];
+        if (packageName) {
+            const nextVersion = response.data['dist-tags']?.next;
+            const latestVersion = response.data['dist-tags'].latest;
+
+            if (nextVersion && semver.gt(nextVersion, latestVersion)) {
+                versions[packageName] = nextVersion;
+            } else {
+                versions[packageName] = latestVersion;
+            }
```
ericallam, line 144: This is really close, but we also need to check and make sure the existing package version is using the `next` tag. 

I think the pseudocode would be something like

```
If installed package is using the next tag
  Use the newest version of either the published next tag or the published latest tag
If installed package is using the latest tag
  use the newest version of only the published latest tag
```

 
We never want to go from `latest` to `next`, but if `latest` is newer than `next`, we should go from `next` to `latest`.

ericallam (changes requested):

ericallam:
Could you also add a patch changeset for the `@trigger.dev/cli` package for the changes in this PR:

From the root of the repo run:

```sh
pnpm run changeset:add
```

Select the `@trigger.dev/cli` package only and follow the prompts to make it a `patch` changeset.

packages/cli/src/commands/update.ts
omernave: so if I understand correctly I should check in the package.json and if I see "@trigger.dev/*":"next", I should updated to the most recent version even if its latest, otherwise use the latest. Is that correct?

ericallam: Yea to see if the installed version is tagged with `next` you would match against the version string for the string `"next"`, probably just with `const isNext = version.includes("next")` or something similar

ericallam: @omernave does what I said above make sense? Keen to get this PR merged as it's so close and it'd be a really great addition

omernave: @ericallam updated the pr based on the last comment, can you take a look?

packages/cli/package-lock.json
ericallam, line 1: Can you delete this file please? We're using a pnpm monorepo so there's no need for package-lock.json

packages/cli/src/cli/index.ts
```diff
@@ -79,6 +80,17 @@ program
     await createIntegrationCommand(path, options);
   });
 
+program
+  .command("update")
+  .description("Update the Trigger.dev CLI to the latest version")
```
ericallam: This should be "Updates all @trigger.dev to their latest compatible versions"

packages/cli/src/cli/index.ts
```diff
@@ -79,6 +80,17 @@ program
...
+  .argument(
+      "[path]",
+      "The path where you would like the Trigger.dev packages to be updated",
```
ericallam: "The path to the directory that contains the package.json file"

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,134 @@
...
+        });
+    });
+}
+
+async function getLatestVersions(packageNames: string[], packageVersion: { [name: string]: string}): Promise<{ [packageName: string]: string }> {
+    const registryUrl = 'https://registry.npmjs.org';
+
+    const requests = packageNames.map(packageName =>
+        axios.get(`${registryUrl}/${packageName}`, {
```
ericallam: Could you replace axios here with `node-fetch`, which we are already using in the project? And please remove the `axios` dependency 🙏

packages/cli/src/commands/update.ts
ericallam: Yup, this logic looks correct now, nice one!

ericallam (changes requested):
The logic for updated looks solid now, added just a few more review comments. Can you also add a changeset to this PR (should be a `patch` change on just the `@trigger.dev/cli` package. To add a changeset run this command in the root of the repo:

```sh
pnpm run changeset:add
```

Then select just the `@trigger.dev/cli` command, skip the "major" and "minor" updates, and then add a description for the patch change. This will produce a file that you should then commit to this PR.

packages/cli/package-lock.json
omernave: done

packages/cli/src/cli/index.ts
omernave: done

packages/cli/src/cli/index.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: done

omernave:
Made the requested changes and ran the changeset command

ericallam (approved):

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,136 @@
...
+            logger.info('Dependencies updated.');
+        } else {
+            logger.info('Update canceled.');
+        }
+    } catch (error) {
+        logger.error('An error occurred:', error);
+    }
+}
+
+async function getInstalledPackages(): Promise<any> {
```
ericallam: This function doesn't use the resolvedPath, which means it won't be listing the correct dependencies. I also tried running this locally and it failed because `pnpm list --depth=0 --json` returned an array, not an object:

```js
[
  {
    name: 'triggerdotdev',
    version: '0.0.1',
    path: '/Users/eric/code/triggerdotdev/trigger.dev',
    private: true,
    dependencies: { '@changesets/cli': [Object], 'node-fetch': [Object] },
    devDependencies: {
      '@manypkg/cli': [Object],
      '@tailwindcss/forms': [Object],
      '@tailwindcss/typography': [Object],
      '@trigger.dev/cli': [Object],
      autoprefixer: [Object],
      'eslint-config-custom': [Object],
      postcss: [Object],
      prettier: [Object],
      tailwindcss: [Object],
      tsx: [Object],
      turbo: [Object],
      vite: [Object],
      'vite-tsconfig-paths': [Object],
      vitest: [Object]
    },
    unsavedDependencies: {
      prisma: [Object],
      '@prisma/engines': [Object],
      '@prisma/client': [Object],
      '@prisma/engines-version': [Object]
    }
  }
]
```

ericallam (changes requested):

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,144 @@
...
+    const manager = getUserPkgManager();
+    return new Promise((resolve, reject) => {
+        exec(`${manager} list --depth=0 --json`, {
+            cwd: resolvedPath
+        }, (error, stdout) => {
+            if (error) {
+                reject(error);
+            } else {
+                const installedPackages = JSON.parse(stdout.toString());
+                switch (manager) {
```
ericallam, line 74: This logic looks incorrect. Here's what I get when I run the list command using all three different package managers we support (pnpm, npm, and yarn) on a freshly created Next.js project

- `pnpm list --depth=0 --json` - https://jsonhero.io/j/Q9suKMhpFluT/editor
- `yarn list --depth=0 --json` - https://jsonhero.io/j/YG9GzRBVQ3Zj/editor
- `npm list --depth=0 --json` - https://jsonhero.io/j/FQEbMqFzH4xz/editor





```json

**Review by ericallam — CHANGES_REQUESTED** [review]

File: packages/cli/src/commands/update.ts [hunk 12]
```diff
@@ -0,0 +1,176 @@
...
+        if (Object.keys(updates).length === 0) {
+            console.log('No updates available.');
+            return;
+        }
+
+        const confirm = await confirmUpdate(updates);
+
+        if (confirm) {
+            const updatedPackageJson = updatePackageJson(updates);
+            await writeJSONFile('package.json', updatedPackageJson);
```
ericallam: This is not respecting the `resolvedPath` so the wrong `package.json` will get written to if `path` isn't `.`. Need to do:

```ts
import pathModule from "path";

await writeJSONFile(pathModule.join(resolvedPath, 'package.json'), updatedPackageJson);
```

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,176 @@
...
+            logger.info('package.json updated.');
+
+            // Run 'npm install' to install the updated packages
+            logger.info('Running npm install...');
```
ericallam: `installDependencies` already logs this out so could you please remove?

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,176 @@
...
+            await installDependencies(resolvedPath);
+            logger.info('Dependencies updated.');
```
ericallam: You can remove this log message as well

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,176 @@
...
+            } else {
+                versions[packageName] = latestVersion;
+            }
+        }
+        return versions;
+    }, {});
+}
+
+function updatePackageJson(updates: { [packageName: string]: string }) {
+    logger.info(JSON.stringify(updates));
```
ericallam: Can you remove this log message?

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,176 @@
...
+    const packageJson = JSON.parse(readFileSync('package.json', 'utf8'));
```
ericallam: This reading has two issues, one the package.json can be incorrect and needs to be built using the resolvedPath. The other is we already have `readJSONFile` and `readJSONFileSync` utils, so please use one of those.

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,176 @@
...
+    Object.entries(updates).forEach(([packageName, version]) => {
+        packageJson.dependencies[packageName] = `^${version}`;
+    });
+
+    return packageJson;
+}
+
+async function confirmUpdate(updates: { [packageName: string]: string }) {
+    logger.info('The following updates are available:');
+    console.table(updates);
```
ericallam, line 162: Love the use of the table 👍

ericallam (changes requested):

packages/cli/src/commands/update.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: done

packages/cli/src/commands/update.ts
omernave: done

ericallam:
@omernave happy to accept a renewed PR for this but if you do want to give it another try, please document in the PR the steps you took to test the change to make sure it works. Unfortunately we just don't have the bandwidth to review and test PRs if they haven't already been tested by the submitter.  I know you did a lot of work on this and we really appreciate it.

/tip $50

PR #176, opened 2023-07-15
feat: added defineJob method in TriggerClient

akhill10:


packages/trigger-sdk/src/triggerClient.ts
```diff
@@ -861,6 +862,15 @@ export class TriggerClient {
       },
     };
   }
+
+  defineJob(
```
ericallam: This looks great, but I think `defineJob` needs to be generic so the type of the job returned includes the actual Trigger type (we're eventually going to support direct job invocation, so need to make sure the Trigger type is preserved and not just `any`) Otherwise this whole PR is 👍

ericallam (changes requested):

packages/trigger-sdk/src/triggerClient.ts
akhill10: Yeah, agree. Done

ericallam:
Can you add a changeset for a "patch" update to the `@trigger.dev/sdk` package and include it in this PR? 

To add a change, run this from the root of the repo:

```sh
pnpm run changeset:add
```

Then select `@trigger.dev/sdk` only and follow the steps for creating a "patch" changeset

ericallam (approved):

PR #177, opened 2023-07-15
feat: added isRetry in context.run

### LINKED ISSUE #154: [TRI-644] On a Run retry add info to `context.run`
When this button is pressed it retries the existing run.

Linear-hosted image
[Image source](https://github.com/triggerdotdev/trigger.dev/assets/10635986/8e7abfb8-a962-42a2-9387-30796b3142b5)

In this case, we should add some metadata to `context.run` so you can identify that it is a rerun and potentially how many attempts have been made as well.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-644](https://linear.app/triggerdotdev/issue/TRI-644/on-a-run-retry-add-info-to-contextrun)</sub>

akhill10:
fixes #154 

- Added `isRetry` in context.run object
- Added migration for `isRetry` column for EventRecord and JobRun tables
- Setting `isRetry` as `true` when it is `reRunService.call()` is executed  

/claim #154


ericallam (changes requested):
I probably needed to be more clear in this issue as this `isRetry` boolean should only be set when the "Retry job run" button is pressed which runs the `ContinueRunService` service, not the "Run again" button which calls the `ReRunService`. I actually think this means that `isRetry` shouldn't be on the `JobRun` or `EventRecord` tables, but instead on the `JobRunExecution` table is `isRetry` set to true here:

https://github.com/triggerdotdev/trigger.dev/blob/2d722b94bd448e39506f0d6c9fff3c83e373ef2b/apps/webapp/app/services/runs/continueRun.server.ts#L72

And then using the `isRetry` from the JobRunExecution record here to set `context` of the run:

https://github.com/triggerdotdev/trigger.dev/blob/2d722b94bd448e39506f0d6c9fff3c83e373ef2b/apps/webapp/app/services/runs/performRunExecution.server.ts#L266

ericallam:
Can you add a changeset for a "patch" update to the @trigger.dev/sdk package and include it in this PR? Everything else looks good to go!

ericallam (approved):

PR #175, opened 2023-07-15
chore: allow CLI to use `endpointId` if it is available in package.json

kaf-lamed-beyt:
fixes #165 

@ericallam kindly take a look at this and let me know if I'm on the right track.

/claim #165

PR #182, opened 2023-07-17
chore(docs): docker-setup guide

kaf-lamed-beyt:
@ericallam, Kindly take a look. 

One more thing, I'd love to be assigned to good first issues so I can become more familiar with the codebase as time goes on. Which ones can you recommend?


kaf-lamed-beyt:
Hi @ericallam,

In case you missed this and the latest update here 👉🏽 #175

PR #185, opened 2023-07-17
feat: 🎸 move get project jobs to lower level in app hierarchy

SSHari:
PR for #155 (TIR-573).

## Changes

While the requested change mentions the Job list page, I noticed that we grab a project's jobs in several parts of the application:
1. Job list page
image

2. Job page breadcrumb
image

3. Integration jobs page
image
image

The changes are a bit more involved, but I went ahead and added a new `JobListPresenter` and updated _all_ of the relevant parts of the app above that consume project jobs.

## Testing

1. I spun up the trigger.dev app locally
2. I created a new NextJS app locally
3. I followed the [trigger dev quickstart guide](https://trigger.dev/docs/documentation/quickstart) to set up a new integration (this sets up a new integration with 1 example job)
4. I navigated to the job details page for the example job
5. While both the trigger.dev and NextJS app were running, I added a new job to my NextJS app
6. I navigated back to the jobs list page and made sure the new job loaded
7. I added a new job to my NextJS app
8. I navigated to the job details page of one of the jobs and made sure the new job showed up in the Jobs breadcrumb section
10. I added a new job to my NextJS app with a dummy integration (so it shows up on the integrations page)
image

11. I added a new job to my NextJS app with the same dummy integration
12. I navigated to the job details page for the dummy integration and made sure that the new job loaded

Bounty: /claim #155


apps/webapp/app/components/navigation/JobsMenu.tsx
```diff
@@ -18,6 +19,7 @@ export function JobsMenu({ matches }: { matches: RouteMatch[] }) {
   const [isOpen, setIsOpen] = useState(false);
   const organization = useOrganization(matches);
   const project = useProject(matches);
+  const projectJobs = useJobs();
```
matt-aitken: You should pass the `matches` into `useJobs()`. 

This is done because when switching between pages we were getting React errors about the number of hooks called changing. We call `useMatches` once in the `Breadcrumb` component, always. Then pass those matches to everything else. Seems to fix this issue…

SSHari: Good catch. I'll update the code with the suggested change. Thanks!

ericallam:
Can you run this (in the root of the repo):

```sh
pnpm run typecheck --filter webapp
```

And make sure that there are no TS errors. I'm getting the following errors when running it on your changes:

```
webapp:typecheck: app/components/jobs/JobsTable.tsx(84,42): error TS7006: Parameter 'integration' implicitly has an 'any' type.
webapp:typecheck: app/components/jobs/JobsTable.tsx(120,48): error TS7006: Parameter 'property' implicitly has an 'any' type.
webapp:typecheck: app/components/jobs/JobsTable.tsx(120,58): error TS7006: Parameter 'index' implicitly has an 'any' type.
webapp:typecheck: app/components/jobs/JobsTable.tsx(132,48): error TS7006: Parameter 'property' implicitly has an 'any' type.
webapp:typecheck: app/components/jobs/JobsTable.tsx(132,58): error TS7006: Parameter 'index' implicitly has an 'any' type.
webapp:typecheck: app/hooks/useFilterJobs.ts(14,34): error TS7006: Parameter 'integration' implicitly has an 'any' type.
webapp:typecheck: app/hooks/useFilterJobs.ts(21,32): error TS7006: Parameter 'property' implicitly has an 'any' type.
webapp:typecheck: app/hooks/useProject.tsx(9,41): error TS2339: Property 'jobs' does not exist on type '{ id: string; slug: string; name: string; organizationId: string; createdAt: Date; updatedAt: Date; hasInactiveExternalTriggers: boolean; hasUnconfiguredIntegrations: boolean; environments: { ...; }[]; }'.
```

SSHari:
@ericallam I've gone ahead and updated the types accordingly 👍

apps/webapp/app/presenters/JobListPresenter.server.ts
```diff
@@ -0,0 +1,182 @@
...
+    this.#prismaClient = prismaClient;
+  }
+
+  public async call({
+    userId,
+    slug,
+  }: Pick<Project, "slug"> & {
+    userId: User["id"];
+  }) {
+    const project = await this.#prismaClient.project.findFirst({
```
ericallam: I think this would be better if it was a call to `this.#prismaClient.job.findMany` that included a `where` clause that made this the jobs returned belongs to the project passed in.

apps/webapp/app/hooks/useJobs.tsx
```diff
@@ -0,0 +1,26 @@
...
+import type { loader } from "~/routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam/route";
+import { RouteMatch } from "@remix-run/react";
+import { useTypedMatchesData } from "./useTypedMatchData";
+
+export type ProjectJob = UseDataFunctionReturn<
+  typeof loader
+>["projectJobs"][number];
+
+export const jobsMatchId =
+  "routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam";
```
ericallam, line 12: This doesn't seem like the correct match ID, shouldn't it be pointing to the index route? e.g. `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs._index`

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.integrations_.$clientParam._index/route.tsx
```diff
@@ -29,15 +26,25 @@ export const loader = async ({ request, params }: LoaderArgs) => {
...
+    integrationPresenter.call({
+      userId: userId,
+      organizationSlug,
+      projectSlug: projectParam,
+      clientSlug: clientParam,
+    }),
+    jobsPresenter.call({ userId, slug: projectParam }),
+  ]);
+
+  const integrationJobIds = integrationJobs.jobs.map((job) => job.id);
```
ericallam: This filtering should be happening in the call to prisma using where clauses. Maybe the `JobListPresenter.call` function could take some optional params to be able to do this in SQL instead? If it would be a ton of work to do this then don't worry, you can leave as is.

ericallam (changes requested):

apps/webapp/app/presenters/JobListPresenter.server.ts
SSHari: Good call. I went ahead and made this change to use `job.findMany` instead.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.integrations_.$clientParam._index/route.tsx
SSHari: Good idea. Makes sense. I've removed [IntegrationClientJobsPresenter.server.ts](https://github.com/triggerdotdev/trigger.dev/pull/185/files#diff-9c4eb5129b53961fdec83f0edc3a4a52727a089a14482a1548666ec5077e02e9) and updated the `JobListPresenter.call` to take optional arguments to filter the jobs list by organization and integration slugs.

apps/webapp/app/hooks/useJobs.tsx
SSHari: This one is a little tricky. I think you're correct that `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs._index` is the more accurate id for that page and job breadcrumb (this was my initial thought as well).

The problem is that we _also_ need the project jobs for the `useJob` hook. For example on the [job trigger page](https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/routes/_app.orgs.%24organizationSlug.projects.%24projectParam.jobs.%24jobParam.runs.%24runParam.trigger/route.tsx#L28). This means that we have to include the project jobs on the route `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam` because the trigger page _won't_ match `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam._index`.

If we make the suggested change for `useJobs` and make it `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam._index` then we would need to return the project jobs from both the `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam` and `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam._index` routes to account for both situations.

While that should work fine, I think the implication is that if you're on the `routes/_app.orgs.$organizationSlug.projects.$projectParam.jobs.$jobParam._index` page then we would make the same exact call from both loaders which seems wasteful. This is why I ended up using this id instead.

If you prefer we just make the call from both places, then I can update this id and then also update the associated loader to return the project jobs.

ericallam: Okay, that makes sense, and thanks for the clarification!

ericallam (approved):

PR #194, opened 2023-07-19
[TRI-699] using pathToFileURL on path used in detectPagesOrAppDir

### LINKED ISSUE #190: [TRI-699] The CLI init command fails on Windows because of file:// paths
## Steps to reproduce
Running the init command on Windows results in this:

```bash
"√ Successfully added dependencies to package.json: @trigger.dev/sdk@2.0.0-next.18, @trigger.dev/nextjs@1.0.0-next.8
✅ Added TRIGGER_API_KEY=*** to .env.local
✅ Added TRIGGER_API_URL=https://cloud.trigger.dev/ to .env.local
Aborting installation...
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'"
```

<sub>[TRI-699](https://linear.app/triggerdotdev/issue/TRI-699/the-cli-init-command-fails-on-windows-because-of-file-paths)</sub>

dylanrod21:
fixes #190 
/claim #190

### **Changes Made**

1. I followed the instructions to setup trigger.dev in my local NextJS app pointing to my local trigger.dev fork and reproduced the error.
2.  I traced the code and found the location of the path causing the issue.
3. I added url from npm to use pathToFileURL to convert the path to one that would be valid on windows.

### **Testing**
1. I spun up the trigger.dev app locally
2. I created a new NextJS app locally 
3. I ran the init command in the cli from my updated code for the cli in my local Next app and confirmed that the installation no longer aborted due to path issues

### DIFF
--- a/.changeset/mighty-gorillas-march.md
+++ b/.changeset/mighty-gorillas-march.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/cli": patch
+---
+
+This is a patch bump
--- a/apps/webapp/app/models/organization.server.ts
+++ b/apps/webapp/app/models/organization.server.ts
@@ -46,7 +46,7 @@ export function getOrganizations({ userId }: { userId: User["id"] }) {
         include: {
           _count: {
             select: {
-              jobs: true,
+              jobs: { where: { internal: false } },
             },
           },
         },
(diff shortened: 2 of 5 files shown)


ericallam:
This looks great! About to test to make sure it still works on macOS but in the meantime could you add a changeset to this PR by running this in the root:

```sh
pnpm run changeset:add
```

Then follow the prompts to select a "patch" bump, like so:

https://github.com/triggerdotdev/trigger.dev/assets/534/3a73aaf6-d6c3-4090-a0e9-e1ee7338a929

ericallam (approved):

PR #203, opened 2023-07-21
New integration: `@trigger.dev/supabase`

ericallam:
I also had to update the webapp OAuth2 functionality to work with Supabase's implementation of OAuth for this Management API. The `@trigger.dev/supabase` exposes two different integrations:

- `SupabaseManagementAPI` - this wraps the [supabase-management-js](https://github.com/supabase-community/supabase-management-js) package to give developers access to the [Management API](https://supabase.com/docs/reference/api/introduction). Developers can also use this to trigger jobs on database changes:

```ts
const supabase = new SupabaseManagement({
  id: "supabase",
});

new Job(client, {
  id: "supabase-on-user-insert",
  name: "Supabase On User Insert",
  version: "0.1.1",
  trigger: supabase.onInserted<UserRecord>({
    projectRef: process.env.SUPABASE_ID!,
    table: "users",
  }),
  run: async (payload, io, ctx) => {},
});
```

- `Supabase` - this wraps the [supabase-js](https://github.com/supabase/supabase-js) package and allows developers to access their supabase DB, auth, and functions using their Service Account access token:

```ts
const supabaseDB = new Supabase<Database>({
  id: "supabase-db",
  supabaseUrl: `https://${process.env.SUPABASE_ID}.supabase.co`,
  supabaseKey: process.env.SUPABASE_KEY!,
});

new Job(client, {
  id: "supabase-create-todo",
  name: "Supabase Create Todo",
  version: "0.1.1",
  trigger: eventTrigger({
    name: "supabase.create-todo",
    schema: z.object({
      contents: z.string(),
      user_id: z.number(),
    }),
  }),
  integrations: {
    supabaseDB,
  },
    const newTodo = await io.supabaseDB.runTask(
      "create-todo",
      async (db) => {
        const { data, error } = await db
          .from("todos")
          .insert({
            contents: payload.contents,
            user_id: payload.user_id,
            is_complete: false,
          })
          .select();

        if (error) throw error;

        return data;
      },
      {
        name: "Create Todo",
        properties: [{ label: "Contents", text: payload.contents }],
      }
    );
  },
});
```

I also added the new `runTask` function available on all integrations as you can see above.

Still todo:

- [x] Audit the Supabase integrations on the webapp side to make sure they all make sense
- [x] Test the GitHub OAuth flow
- [x] Test the Slack OAuth flow
- [x] Add documentation in the [trigger.dev-docs](https://github.com/triggerdotdev/trigger.dev-docs) repo


apps/webapp/app/routes/oauth2.callback.ts
matt-aitken, line 1: I'm assuming that this new import is actually not used? Not a big deal

apps/webapp/app/services/externalApis/integrationCatalog.server.ts
```diff
@@ -35,4 +36,6 @@ export const integrationCatalog = new IntegrationCatalog({
   resend,
   slack,
   typeform,
+  supabaseManagement,
+  supabase,
```
matt-aitken, line 40: It'd be good to keep these all alphabetical but not a big deal.

integrations/supabase/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "dependencies": {
+    "@supabase/supabase-js": "^2.26.0",
+    "@trigger.dev/integration-kit": "workspace:^2.0.0-next.5",
+    "@trigger.dev/sdk": "workspace:^2.0.0-next.16",
+    "supabase-management-js": "^0.1.2",
+    "zod": "3.21.4"
+  },
+  "engines": {
+    "node": ">=18.0.0"
```
matt-aitken, line 35: Do we need to reduce all the package node requirements so they're inline with the SDK? 16 I think. Not urgent.

matt-aitken (approved):

integrations/supabase/package.json
ericallam: The supabase-management-js package requires `>=18.0.0`, hence the reason this one does as well

PR #207, opened 2023-07-22
feat: Add Javascript support for init command

### LINKED ISSUE #205: [TRI-711] Add JavaScript support to the CLI init command
Currently the init command generates TypeScript files:

* trigger.ts
* The API route (location and content varies for pages or app router)
* Example job file

If JavaScript is detected then we should generate JS files instead of TS.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-711](https://linear.app/triggerdotdev/issue/TRI-711/add-javascript-support-to-the-cli-init-command)</sub>

jemiluv8:
Modify the init cli command to work with nextjs projects using Javascript

### Testing

- Start the local(self-hosted) version of trigger.dev
- Create a nextjs project using javascript
- Execute the cli init command from the repo and specify the project-path to point to your javascript project
- Enter relevant answers to the prompt
- The init command creates the necessary javascript files in the nextjs (Javascript) project

fixes #205
/claim #205


packages/cli/src/commands/init.ts
```diff
@@ -68,14 +68,6 @@ export const initCommand = async (options: InitCommandOptions) => {
 
   const isTypescriptProject = await detectTypescriptProject(resolvedPath);
 
-  if (!isTypescriptProject) {
```
jemiluv8, line 71: This normally throws an error if the project is not using typescript.
We don't need to do this check anymore.

packages/cli/src/commands/init.ts
```diff
@@ -114,7 +106,7 @@ export const initCommand = async (options: InitCommandOptions) => {
     logger.info("📁 Detected use of src directory");
   }
 
-  const nextJsDir = await detectPagesOrAppDir(resolvedPath, usesSrcDir);
+  const nextJsDir = await detectPagesOrAppDir(resolvedPath, usesSrcDir, isTypescriptProject);
```
jemiluv8, line 109: modify detectPagesOrAppDir to work with js projects as well.

packages/cli/src/commands/init.ts
```diff
@@ -123,13 +115,15 @@ export const initCommand = async (options: InitCommandOptions) => {
       resolvedPath,
       routeDir,
       resolvedOptions,
+      isTypescriptProject,
```
jemiluv8, line 118: Needed to add another argument to let the "generator" know the kind(ts/js) of project we are dealing with.

packages/cli/src/commands/init.ts
```diff
@@ -281,11 +276,13 @@ async function detectPagesOrAppDir(
     // If so then we return app
     // If not return pages
 
+    const extension = isTypescriptProject ? "tsx" : "js"
```
jemiluv8, line 279: determine extension of the file to lookout for

packages/cli/src/commands/init.ts
```diff
@@ -432,11 +430,18 @@ async function createTriggerAppRoute(
...
   options: ResolvedOptions,
+  isTypescriptProject: boolean,
   usesSrcDir = false
 ) {
-  const tsConfigPath = pathModule.join(projectPath, "tsconfig.json");
+  const configFileName = isTypescriptProject ? "tsconfig.json" : "jsconfig.json"
+  const tsConfigPath = pathModule.join(projectPath, configFileName);
   const tsConfig = await readFile(tsConfigPath);
 
+  const extension = isTypescriptProject ? ".ts" : ".js"
```
jemiluv8, line 440: Introduced to help us properly name files that will be generated.

packages/cli/src/commands/init.ts
```diff
@@ -432,11 +430,18 @@ async function createTriggerAppRoute(
   projectPath: string,
   path: string,
   options: ResolvedOptions,
+  isTypescriptProject: boolean,
   usesSrcDir = false
 ) {
-  const tsConfigPath = pathModule.join(projectPath, "tsconfig.json");
+  const configFileName = isTypescriptProject ? "tsconfig.json" : "jsconfig.json"
```
jemiluv8, line 436: Since it will not all be ts projects we needed to update the config file path.
I'm relying on the assumption here that all js projects will have a jsconfig.json file

packages/cli/src/commands/init.ts
```diff
@@ -464,7 +469,7 @@ export const client = new TriggerClient({
   const jobsPathPrefix = pathAlias ? pathAlias + "/" : "../";
 
   const jobsContent = `
-import { Job, eventTrigger } from "@trigger.dev/sdk";
```
jemiluv8, line 467: Remove unused import for Job

jemiluv8 (commented):
Self review completed.

ericallam:
This looks great! Going to review and test it out shortly, but could you go ahead and add a "patch" changeset for the `@trigger.dev/cli` package by running this in the root:

```sh
pnpm run changeset:add
```

Select just the `@trigger.dev/cli` package and then follow the prompts to make a `patch` update to the package, along with a message describing the change.

jemiluv8:
Change set for "minor" bump in cli added

packages/cli/src/commands/init.ts
matt-aitken: Shouldn't this be "jsx", not "js"?

jemiluv8: It can be jsx but I've seen .js used in the default page file when we generate a next project.
They are both valid. So depends on your tastes.
Would you prefer I use jsx?

ericallam:
Tested and everything seems to be working, merging now and will be out in the next package release.

PR #214, opened 2023-07-25
Feat: Detect if a Next.js project is running on the specified port before opening a tunnel

### LINKED ISSUE #210: [TRI-728] @trigger.dev/cli dev command should actually detect if a Next.js project is running on the specified port before opening a tunnel
Currently in the dev command we don't check if there is a server running locally on the specified port (by default port 3000) in the `dev` command, even though we say we are (see code here: https://github.com/triggerdotdev/trigger.dev/blob/09f906f7a193ade59fb65eaf8da936fe428f13b9/packages/cli/src/commands/dev.ts#L56)

If there is no server found running on the port, we should print an error message clearly explaining the issue and then exit the command.

The way to check if there is a server found running on the port is to perform a "PING" to the `endpointHandlerUrl`.  You can do that with the following code (which uses node-fetch)

```ts
const response = await fetch(endpointHandlerUrl, {
  method: "HEAD",
  headers: {
    "x-trigger-api-key": apiKey,
    "x-trigger-action": "PING",
    "x-trigger-endpoint-id": endpointId,
  },
});
```

<sub>[TRI-728](https://linear.app/triggerdotdev/issue/TRI-728/triggerdevcli-dev-command-should-actually-detect-if-a-nextjs-project)</sub>

Ashutosh-Bhadauriya:
Fixes #210
/claim #210

Detect if a Next.js project is running on the specified port before opening a tunnel

### DIFF
--- a/.changeset/blue-monkeys-speak.md
+++ b/.changeset/blue-monkeys-speak.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/cli": patch
+---
+
+Detect if a nextjs project is running on the specified port before opening tunnel
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -9,7 +9,7 @@ import { logger } from "../utils/logger.js";
 import { resolvePath } from "../utils/parseNameAndPath.js";
 import { TriggerApi } from "../utils/triggerApi.js";
 import dotenv from "dotenv";
-import fetch from 'node-fetch';
+import fetch from "node-fetch";
 
 export const DevCommandOptionsSchema = z.object({
   port: z.coerce.number(),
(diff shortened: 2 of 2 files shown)


Ashutosh-Bhadauriya:
@ericallam I have done the changes, but not sure how to test it.

What i'm doing right now?
Start trigger app and don't start the nextjs app, then it should show in terminal right? Or I've misunderstood the issue?

ericallam:
Also, it would be amazing to get the above instructions added to the CONTRIBUTING.md guide under a section towards the end called something like "Testing CLI changes". If you can do this afterwards in a separate PR I'll definitely tip you handsomely 👍

ericallam:
Oh, and don't forget to cleanup and delete that temporary Next.js project before committing anything

Ashutosh-Bhadauriya:
Okay great! Will also add to  CONTRIBUTING.md in the next pr 👍

Ashutosh-Bhadauriya:
Tested the changes and made some changes. Working fine 👍

Next, will raise a pr adding the above things you told in CONTRIBUTING.md

Ashutosh-Bhadauriya:
Okay, now its working, no idea what the issue was 😅

Ashutosh-Bhadauriya:
> I just realized, this might cause a race-condition when using something like the concurrently package to run both the Next.js dev server and the @trigger.dev/cli dev command at the same time, like so:

@ericallam I tried this, the dev command seems to be working fine. It didn't fail.

ericallam:
Awesome! Can you add a changeset to this PR for `@trigger.dev/cli`?

Ashutosh-Bhadauriya:
Just added it

PR #221, opened 2023-07-26
Get the status of runs and events, using the SDK or @trigger.dev/react hooks

matt-aitken:
# App changes

### Run details
`api.v1.runs.$runId.ts`
- Returns info on the run. 
- Can be called using public and private API key. 
- If called with a private key (i.e. from the SDK) it can return details on individual tasks too.

### Event details
`api.v1.events.$eventId.ts`
- Returns info on the event, including runs that were triggered by it.
- Can be called using public and private API key.

### Job runs
`api.v1.jobs.$jobSlug.runs.ts`
- Returns the latest runs for a given job slug (the `id` field in the job definition)
- Only works with the private API key

## API keys
Private API keys are now guaranteed to be unique (and non-null). Also lengthened new API keys of both varieties.
- [x] Add private API key to the environments page

## Subtasks query
In the RunPresenter there was some horrible logic for dealing with subtasks, where we retrieved them from the database in a hierarchy and then enriched them in the tree. This made the logic a lot less portable for getting subtasks elsewhere, like in the new endpoint above.

Now you get all the tasks (including subtasks) in a flat structure. Then use the shared `taskListToTree()` function to create a tree structure. Much more flexible (and probably better performance because of how Prisma does deep queries).

# @trigger.dev/sdk
- `client.getEvent()`
- `client.getRun()`
- `client.getRuns()`

# @trigger.dev/react
### `<TriggerProvider>`
This creates a context, that the hooks need to operate. Throws an error if you pass in a private API key. Also, throws an error if you try use a hook without having put this somewhere.

### `useRunDetails()` hook
Gets the details of the run, using the new run API endpoint. Refreshes automatically until the run is settled.

### `useEventDetails()` hook
Gets the details of an event, using the new event API endpoint. Refreshes automatically until the event's runs are completed.

### `useEventRunDetails()` hook
It combines the two hooks to make it really convenient to just pass in an event and get the first run details.

# nextjs-example project
Reworked this project so it only uses the app router, it was really confusing before. Also added a hook job and the code to test it inside some pages.


apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.environments/route.tsx
```diff
@@ -146,17 +147,38 @@ export default function Page() {
...
-                      fullWidth={false}
-                      secure
-                      value={environment.apiKey}
-                      variant={"primary/medium"}
-                      icon={<EnvironmentLabel environment={environment} />}
-                    />
-                  ))}
+                <div className="mb-8">
+                  <Paragraph variant="small" spacing>
+                    Secret API keys should be used on your server – they give
```
ericallam: What's the thinking behind calling these "Secret API keys" and not either "Private API keys" or "Server API keys"?

apps/webapp/app/routes/api.v1.events.$eventId.ts
```diff
@@ -0,0 +1,51 @@
...
+    allowPublicKey: true,
+  });
+  if (!authenticatedEnv) {
+    return apiCors(
+      request,
+      json({ error: "Invalid or Missing API key" }, { status: 401 })
+    );
+  }
+
+  const { eventId } = ParamsSchema.parse(params);
```
ericallam: This should probably safeParse and return an error message with a JSON response

apps/webapp/app/routes/api.v1.events.$eventId.ts
```diff
@@ -0,0 +1,51 @@
...
+          completedAt: true,
+        },
+      },
+    },
+    where: {
+      id: eventId,
+    },
+  });
+
+  return apiCors(request, json(event));
```
ericallam, line 69: This should return a 404 if the event is not found

apps/webapp/app/routes/api.v1.jobs.$jobSlug.runs.ts
```diff
@@ -0,0 +1,75 @@
...
+    return apiCors(
+      request,
+      json({ error: "Invalid or Missing API key" }, { status: 401 })
+    );
+  }
+
+  const { jobSlug } = ParamsSchema.parse(params);
+
+  const url = new URL(request.url);
+  const query = SearchQuerySchema.parse(Object.fromEntries(url.searchParams));
```
ericallam: Should use safeParse

apps/webapp/app/routes/api.v1.runs.$runId.ts
```diff
@@ -0,0 +1,124 @@
...
+      request,
+      json({ error: "Invalid or Missing API key" }, { status: 401 })
+    );
+  }
+
+  const { runId } = ParamsSchema.parse(params);
+
+  const url = new URL(request.url);
+  const query = SearchQuerySchema.parse(Object.fromEntries(url.searchParams));
+  const showTaskDetails = query.taskdetails && apiKeyResult.type === "PRIVATE";
```
ericallam: Would be better if `authenticateApiRequest` returned the type of key instead of having to do the extra call the `getApiKeyFromRequest`

examples/nextjs-example/src/app/api/send-test-event/route.ts
```diff
@@ -0,0 +1,9 @@
+import { client } from "@/trigger";
+import { NextResponse } from "next/server";
+
+export async function GET(request: Request) {
```
ericallam: Shouldn't this happen in a POST request?

examples/nextjs-example/src/components/ReactQuery.tsx
```diff
@@ -0,0 +1,12 @@
+"use client";
+
+import { useState } from "react";
+import { QueryClient, QueryClientProvider } from "react-query";
+
+export function ReactQuery({ children }: { children: React.ReactNode }) {
```
ericallam, line 6: Why doesn't this just happen in our Provider again?

packages/react/tsup.config.ts
```diff
@@ -0,0 +1,19 @@
...
+    platform: "node",
+    format: ["cjs"],
+    legacyOutput: true,
+    sourcemap: true,
+    clean: true,
+    bundle: true,
+    splitting: false,
+    dts: true,
+    external: ["http", "https", "util", "events", "tty", "os", "timers"],
+    esbuildPlugins: [],
```
ericallam, line 18: You need `noExternal: ["@trigger.dev/internal"],` to bundle the internal package here

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.environments/route.tsx
matt-aitken: I'm happy to do `Private` or `Server`. I took it from Stripe but everyone does this differently.

examples/nextjs-example/src/components/ReactQuery.tsx
matt-aitken: This has nothing to do with the package. This is just used in the example to create a React Query context so we can use it to do the mutation in the example to send the test event. 

I wanted to do an example that didn't use server actions because it's important to check that our react query plays nicely with another one.

PR #225, opened 2023-07-26
update cli init to create index file in examples folder

### LINKED ISSUE #224: [TRI-768] The CLI init command should generate a jobs/index file and jobs should be imported there.
When `npx @trigger.dev/cli@latest init` is run it generates the Trigger API route, and an example job.

In the Trigger API route (`app/api/trigger.route.ts` for app router and `pages/api/trigger.ts` for pages router) it imports an individual example job:

```
import { createAppRoute } from "@trigger.dev/nextjs";
import { client } from "@/trigger";

// Replace this with your own jobs
import "@/jobs/examples";

//this route is used to send and receive data with Trigger.dev
export const { POST, dynamic } = createAppRoute(client);
```

Let's change it so we have an `index` file inside the `jobs` folder and just that is imported here.

In that `index` file we would import the `examples` file.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-768](https://linear.app/triggerdotdev/issue/TRI-768/the-cli-init-command-should-generate-a-jobsindex-file-and-jobs-should)</sub>

jemiluv8:
Modified cli init command to create an index file in the jobs folder

modify other code to import from the index

fixes #224
/claim #224


packages/cli/src/commands/init.ts
```diff
@@ -463,7 +464,7 @@ import { createAppRoute } from "@trigger.dev/nextjs";
 import { client } from "${routePathPrefix}trigger";
 
 // Replace this with your own jobs
```
matt-aitken: It would be good to remove this comment, from this file and the one in the pages route.

Instead can we add a comment at the top of the jobs/index file: import all your job files here

matt-aitken (changes requested):

packages/cli/src/commands/init.ts
jemiluv8: Gotcha

packages/cli/src/commands/init.ts
```diff
@@ -569,6 +581,7 @@ async function createTriggerPageRoute(
   const extension = isTypescriptProject ? ".ts" : ".js"
   const triggerFileName = `trigger${extension}`
   const examplesFileName = `examples${extension}`
+  const examplesIndexFileName = `index${extension}`
 
   const routeContent = `
 import { createPagesRoute } from "@trigger.dev/nextjs";
```
matt-aitken, line 560: You need to add the import for the Jobs file to this pages route as well:

`import "@/jobs";`

matt-aitken (changes requested):

packages/cli/src/commands/init.ts
matt-aitken: It's a few lines below this but it wouldn't let me annotate the exact line

jemiluv8: Added jobs import to routeContent for createTriggerPageRoute function

matt-aitken (approved):

PR #235, opened 2023-07-28
Core package and move to @latest

matt-aitken:
## @trigger.dev/core
`@trigger.dev/internal` is now `@trigger.dev/core` and is a published package.

## @examples/package-tester
A new example that uses the compiled packages. No tsconfig.json project references. So we can test package bundling.

## Move from @next to @latest
Packages will now be published without @next, unless they are actually preview versions when testing is being done.

## @trigger.dev/react
The bundling creates separate files, so some of them can "use client" and some not. This means `<TriggerProvider>` works without being wrapped in a "use client" file by the user.

## Fix for Turborepo complaining about env vars
Turborepo doesn't like `process.env.WHATEVER` unless WHATEVER is in the turborepo file. You can use `process.env["WHATEVER"]` instead. I've changed the examples to use this.


apps/webapp/app/routes/api.v1.$endpointSlug.schedules.$id.registrations.ts
```diff
@@ -3,7 +3,7 @@ import { json } from "@remix-run/server-runtime";
 import {
```
ericallam, line 3: These should be type imports or else moved outside of a route and into a .server file

apps/webapp/app/routes/api.v1.$endpointSlug.sources.$id.ts
```diff
@@ -1,6 +1,6 @@
 import type { ActionArgs } from "@remix-run/server-runtime";
 import { json } from "@remix-run/server-runtime";
-import { UpdateTriggerSourceBodySchema } from "@trigger.dev/internal";
+import { UpdateTriggerSourceBodySchema } from "@trigger.dev/core";
```
ericallam, line 3: Type import

apps/webapp/app/routes/api.v1.$endpointSlug.triggers.$id.registrations.$key.ts
```diff
@@ -1,6 +1,6 @@
 import type { ActionArgs } from "@remix-run/server-runtime";
 import { json } from "@remix-run/server-runtime";
-import { RegisterTriggerBodySchema } from "@trigger.dev/internal";
+import { RegisterTriggerBodySchema } from "@trigger.dev/core";
```
ericallam, line 3: Type import

apps/webapp/app/routes/api.v1.$endpointSlug.triggers.$id.registrations.ts
```diff
@@ -1,6 +1,6 @@
 import type { ActionArgs } from "@remix-run/server-runtime";
 import { json } from "@remix-run/server-runtime";
-import { InitializeTriggerBodySchema } from "@trigger.dev/internal";
+import { InitializeTriggerBodySchema } from "@trigger.dev/core";
```
ericallam, line 3: Type import

apps/webapp/app/routes/api.v1.runs.$runId.logs.ts
```diff
@@ -1,8 +1,8 @@
 import type { Organization, RuntimeEnvironment } from "@trigger.dev/database";
 import type { ActionArgs } from "@remix-run/server-runtime";
 import { json } from "@remix-run/server-runtime";
-import type { LogMessage } from "@trigger.dev/internal";
-import { LogMessageSchema } from "@trigger.dev/internal";
+import type { LogMessage } from "@trigger.dev/core";
+import { LogMessageSchema } from "@trigger.dev/core";
```
ericallam, line 5: This should be moved into a server file to make sure core isn't being included in the browser bundle

ericallam (changes requested):

PR #240, opened 2023-07-31
feat: encrypt values in SecretStore using an encryption key

ericallam:
Uses the `aes-256-gcm` symmetric algorithm to encrypt values in the SecretStore. I added a "version" field to the SecretStore, with "1" meaning the old non-encrypted value, and a new version "2" that is the encrypted variant. Updating a secret that was previously version 1 will upgrade it to version 2, but otherwise version 1 SecretStore records will remain unencrypted (well, they are encrypted at rest in AWS). Stores the IV/nonce and the tag alongside the ciphertext (the encrypted data) in the value field instead of the raw value.

### DIFF
--- a/apps/webapp/app/routes/api.v1.integrations.$integrationSlug.connections.$connectionId.ts
+++ b/apps/webapp/app/routes/api.v1.integrations.$integrationSlug.connections.$connectionId.ts
@@ -13,19 +13,13 @@ const ParamsSchema = z.object({
 export async function loader({ request, params }: LoaderArgs) {
   const authenticationResult = await authenticateApiRequest(request);
   if (!authenticationResult) {
-    return apiCors(
-      request,
-      json({ error: "Invalid or Missing API key" }, { status: 401 })
-    );
+    return json({ error: "Invalid or Missing API key" }, { status: 401 });
   }
 
   if (authenticationResult.type !== "PRIVATE") {
-    return apiCors(
-      request,
-      json(
-        { error: "Only private API keys can access this endpoint" },
-        { status: 403 }
-      )
+    return json(
+      { error: "Only private API keys can access this endpoint" },
+      { status: 403 }
     );
   }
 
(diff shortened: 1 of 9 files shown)


ericallam:
Documentation we'll need to update once this is merged:

- [ ] The docker repo: https://github.com/triggerdotdev/docker
- [ ] The fly.io repo: https://github.com/triggerdotdev/fly.io
- [ ] Docs for fly.io: https://trigger.dev/docs/documentation/guides/deploy-to-flyio
- [ ] Docs for render.com: https://trigger.dev/docs/documentation/guides/deploy-to-render

apps/webapp/app/routes/api.v1.integrations.$integrationSlug.connections.$connectionId.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const ParamsSchema = z.object({
+  integrationSlug: z.string(),
+  connectionId: z.string(),
+});
+
+export async function loader({ request, params }: LoaderArgs) {
+  const authenticationResult = await authenticateApiRequest(request);
+  if (!authenticationResult) {
+    return apiCors(
```
matt-aitken: We should only use the `apiCors` function if we want to allow browsers to access the endpoint. It doesn't seem like we do in this situation?

apps/webapp/app/services/secrets/secretStore.server.ts
```diff
@@ -41,11 +44,20 @@ export class SecretStore {
   }
 }
 
+const EncryptedSecretValueSchema = z.object({
+  nonce: z.string(),
+  ciphertext: z.string(),
+  tag: z.string(),
+});
+
 /** This stores secrets in the Postgres Database, in plain text. NOT recommended outside of localhost. */
```
matt-aitken: We should remove this comment now that this is actually encrypted

apps/webapp/app/routes/api.v1.integrations.$integrationSlug.connections.$connectionId.ts
```diff
@@ -0,0 +1,95 @@
+import { LoaderArgs, json } from "@remix-run/server-runtime";
```
matt-aitken, line 1: What is this file used for? Just testing for now?

matt-aitken (changes requested):

apps/webapp/app/routes/api.v1.integrations.$integrationSlug.connections.$connectionId.ts
ericallam: Yup just for being able to look at the auth but something we'll probably need eventually

matt-aitken (approved):

PR #242, opened 2023-08-01
Added telemetry to the CLI

matt-aitken:
Added some telemetry to the CLI so we can tell when problems are happening
- init
- dev

Also, a couple of fixes
- It was asking for the "Project id" twice
- Fixed the "Module not found: Can't resolve 'encoding' in” node-fetch warning


packages/cli/src/cli/index.ts
```diff
@@ -27,7 +28,12 @@ program
   )
   .version(getVersion(), "-v, --version", "Display the version number")
   .action(async (options) => {
-    await initCommand(options);
+    try {
+      await initCommand(options);
+    } catch (e) {
+      telemetryClient.init.failed("unknown", options, e);
```
matt-aitken, line 34: @ericallam I needed to do it like this because we need a different fail event for init and dev

ericallam (approved):

PR #264, opened 2023-08-04
feature: initial work for sendgrid integration

Chigala:
## Description

Created an integration for sendgrid, Please review the code and give me your feedbacks so I could iterate on it. Thanks


integrations/sendgrid/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    const subjectProperty = Array.isArray(params) ? [] : (params.subject ? [{ label: "Subject", text: params.subject }] : []);
+    
+    return {
+      name: "Send Email",
+      params,
+      icon: "sendgrid", 
+      properties: [
+        {
+          label: "From",
+          text: Array.isArray(params) ? params[0].from : params.from,
```
matt-aitken: I got TypeScript errors in VSCode that were complaining about the `init` function.

It's because `.from` and `.to` are the EmailData type which is of this type:

```ts
type EmailData = string | { name?: string; email: string };
```
You could add a function like this to the end of this file which gets a string from from/to

```ts
type EmailData = string | { name?: string; email: string };

function getEmailFromEmailData(emailData: EmailData | undefined): string {
  if (emailData === undefined) {
    return "";
  }

  if (typeof emailData === "string") {
    return emailData;
  }
  return emailData.email;
}
```

You would use it like this:

```ts
text: Array.isArray(params)
            ? getEmailFromEmailData(params[0].from)
            : getEmailFromEmailData(params.from),
```

matt-aitken (changes requested):
This looks great!

There's a TypeScript error to fix which I've made a suggestion for. Should be an easy fix.

The other task is to add a test Job. The best way to do this now is to use the new examples/job-catalog. There's a Readme here: https://github.com/triggerdotdev/trigger.dev/blob/main/examples/job-catalog/README.md

In there you can add a sendgrid file, and a script to package.json. Add the appropriate env vars so you can test it works properly. There are also a couple of other integrations for inspiration.

integrations/sendgrid/src/index.ts
Chigala: Thanks, I've fixed this

Chigala:
Hello @matt-aitken I've hooked up everything. But I've been having issues getting my account to pass sendgrid's review🥲. In the past I usually just use my companies account. My emails aren't sending because of this. I'm kinda stuck on what to do next.

integrations/sendgrid/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+import type { AuthenticatedTask } from "@trigger.dev/sdk";
+
+type SendEmailData = Parameters<(typeof sgMail)["send"]>[0];
+type SendEmailResponse = [sgMail.ClientResponse, {}];
+
+export const sendEmail: AuthenticatedTask<typeof sgMail, SendEmailData, SendEmailResponse> = {
+  run: async (params, client) => {
+    const response = await client.send(params);
+    return response as SendEmailResponse;
+  },
```
matt-aitken: The SendGrid API is so weird… What actually gets returned from this call is this JSON:

```json
[
  {
    "body": "",
    "headers": {
      "date": "Wed, 09 Aug 2023 12:54:29 GMT",
      "server": "nginx",
      "connection": "close",
      "x-message-id": "iN4DrpljT6OOrZW_V1q-eA",
      "content-length": "0",
      "x-no-cors-reason": "https://sendgrid.com/docs/Classroom/Basics/API/cors.html",
      "access-control-max-age": "600",
      "strict-transport-security": "max-age=600; includeSubDomains",
      "access-control-allow-origin": "https://sendgrid.api-docs.io",
      "access-control-allow-headers": "Authorization, Content-Type, On-behalf-of, x-sg-elas-acl",
      "access-control-allow-methods": "POST"
    },
    "statusCode": 202
  },
  ""
]
```

I don't think any of this information is actually useful. If the call fails (like the forbidden error I was getting) it throws anyway. 

I suggest we make it so the function returns void. Which also means we don't need the SendEmailResponse type either.

```suggestion
export const sendEmail: AuthenticatedTask<typeof sgMail, SendEmailData, void> = {
  run: async (params, client) => {
    await client.send(params);
  },
```

matt-aitken (dismissed):
This works 😄 

One requested change (I've suggested code that works) because the SendGrid API is so weird…

CleanShot 2023-08-09 at 13 58 16

This is ready to merge apart from this.

PR #307, opened 2023-08-11
fix: avoid crashing the Node.js app on uncaught exceptions

### LINKED ISSUE #306: Bug: current async middleware doesn't catch exceptions
## Context

Express middleware usage is:

```js
const middleware = createMiddleware(client);
app.use(middleware);
```

## Expected

If for any reason the code in this middleware throws an exception then the error shouldn't crash the Node.js process, and instead be forwarded to the next middleware via `next(error)`

## Actual

If any exceptions are thrown within the middleware code provided by `@trigger.dev/express`, it won't be caught and will crash the Node.js process as an 'uncaught exception' scenario.

This happens to Express lacking native support for promises and so any async promises code within middleware needs to be wrapped with a try/catch.

lirantal:
Fixes #306 by properly creating an exception handler for the middleware code that would otherwise cause the Node.js process to crash.

The new error handler will forward the error thrown to the next middleware until an error handling takes place in the Express app and is the responsibility of the user to define and handle it as they choose to.

### DIFF
--- a/.changeset/smooth-dodos-obey.md
+++ b/.changeset/smooth-dodos-obey.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/express": patch
+---
+
+fix: avoid crashing the Node.js app on uncaught exceptions
--- a/packages/express/src/index.ts
+++ b/packages/express/src/index.ts
@@ -66,18 +66,21 @@ export function createMiddleware(client: TriggerClient, path: string = "/api/tri
       return;
     }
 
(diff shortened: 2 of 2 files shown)


ericallam:
When I saw this I legit knew you were talking about us 😂
CleanShot 2023-08-11 at 14 17 12@2x

lirantal:
lol yeah it's an easy one to get bitten by :-)

PR #309, opened 2023-08-11
fix: support express jsonParser middleware if used in the Express app

### LINKED ISSUE #308: Bug: Express applications using own body parser middleware will crash
## Context

Express middleware usage per the example provided in the package/code is as follows:

```js
 * @example
 * ```ts
 * import express from "express";
 * import { TriggerClient } from "@trigger.dev/sdk";
 * import { createMiddleware } from "@trigger.dev/express";
 *
 * const client = new TriggerClient({
 *  id: "my-client",
 *  apiKey: process.env["TRIGGER_API_KEY"]!,
 * });
 *
 * const app = express();
 *
 * const middleware = createMiddleware(client);
 *
 * app.use(middleware);
 *
 * app.listen(8080, () => {
 *  console.log("Listening on port 8080");
 * });
 * ```
 */
```

## Expected

It's a probable assumption that many Express applications make use of Express's built-in `bodyParser` in order to access HTTP requests JSON payload via `req.body`, such as:

```js
const express = require('express')
const app = express()
app.use(express.json())

// respond with "hello world" when a GET request is made to the homepage
app.post('/api/hello', (req, res) => {
  res.send('hello ', req.body.name)
})
```

Using Trigger.dev's Express integration should work fine.

## Actual

Due to the Express integration code here https://github.com/triggerdotdev/trigger.dev/blob/main/packages/express/src/index.ts#L93C1-L97 

The `body` object key isn't receiving a stream in the format it expects, and as such the middleware function throws an exception (which leads to a Node.js process crash, documented reasons via #306):

```sh
$ node server.js

Listening on port 8080
undefined:1

SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Request.json (/Users/lirantal/projects/repos/trigger.dev/node_modules/.pnpm/@remix-run+web-fetch@4.3.5/node_modules/@remix-run/web-fetch/dist/lib.node.cjs:465:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async TriggerClient.handleRequest (/Users/lirantal/projects/repos/trigger.dev/packages/trigger-sdk/dist/index.js:1262:22)
    at async /Users/lirantal/projects/repos/trigger.dev/packages/express/dist/index.js:61:22
```

lirantal:
Fixes #308 to support both scenarios in which:
1. Express makes use of a body parser middleware such as `express.json()` or `bodyParser.json()` (see references [here](https://expressjs.com/en/resources/middleware/body-parser.html) in official Express docs)
2. Express app doesn't explicitly use a body parser and the request parsing falls back to a readable stream passed to the underlying fetch implementation

### DIFF
--- a/.changeset/sixty-windows-run.md
+++ b/.changeset/sixty-windows-run.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/express": patch
+---
+
+fix: support express jsonParser middleware if used in the Express app
--- a/packages/express/src/index.ts
+++ b/packages/express/src/index.ts
@@ -94,6 +94,6 @@ function convertToStandardRequest(req: express.Request): StandardRequest {
     headers,
     method,
     // @ts-ignore
-    body: req,
+    body: req.body ? JSON.stringify(req.body) : req,
   });
 }


ericallam:
Thanks for the patch!

PR #313, opened 2023-08-11
added SMTP support and removed resend dependency

### LINKED ISSUE #282: [TRI-951] Remove resend dependency from the webapp and replace with nodemailer and smtp
Now that Resend.com [supports sending email via SMTP](https://resend.com/blog/why-smtp-still-matters-today), we should convert the webapp to use a native SMTP mailer instead of the Resend SDK. We will need to replace the RESEND_API_KEY with SMTP_* environment variables (which should still be optional). When SMTP_* env vars are missing, we should continue to print the Magic Link login emails to the console how we are currently doing.

This will allow self-hosters to send email* without having to sign up to Resend.com and possibly use an existing email provider.

* Currently only used for Magic Link login

<sub>[TRI-951](https://linear.app/triggerdotdev/issue/TRI-951/remove-resend-dependency-from-the-webapp-and-replace-with-nodemailer)</sub>

Chigala:
## Description 

Added SMTP support

## Changes
- Removed resend
- Added  nodemailer
- added support for SMTP

fixes #282


apps/webapp/app/env.server.ts
```diff
@@ -28,6 +28,10 @@ const EnvironmentSchema = z.object({
   FROM_EMAIL: z.string().optional(),
   REPLY_TO_EMAIL: z.string().optional(),
   RESEND_API_KEY: z.string().optional(),
+  SMTP_HOST: z.string(),
```
ericallam: These env vars should be optional

packages/emails/src/index.tsx
```diff
@@ -42,15 +42,29 @@ export const DeliverEmailSchema = z
...
-  constructor(config: { apikey?: string; imagesBaseUrl: string; from: string; replyTo: string }) {
-    this.#client =
-      config.apikey && config.apikey.startsWith("re_") ? new Resend(config.apikey) : undefined;
+  constructor(config: {
+    smtpConfig: NodeMailerTransportOptions;
+    imagesBaseUrl: string;
+    from: string;
+    replyTo: string;
+  }) {
+    this.#client = nodemailer.createTransport(config.smtpConfig);
```
ericallam: If the smtpConfig doesn't have the required values, then `this.#client` should be left undefined, just like how it was with the Resend client previously.

packages/emails/src/index.tsx
```diff
@@ -42,15 +42,29 @@ export const DeliverEmailSchema = z
 
 export type DeliverEmail = z.infer<typeof DeliverEmailSchema>;
 
+export type NodeMailerTransportOptions = {
+  host: string;
```
ericallam: These should be optional

ericallam (changes requested):
As well as these changes, we will need to update the following docs

- [ ] Contributing guide
- [ ] The docker repo: https://github.com/triggerdotdev/docker
- [ ] The .env.example
- [ ] The fly.io repo: https://github.com/triggerdotdev/fly.io
- [ ] The fly.io self hosting guide: https://trigger.dev/docs/documentation/guides/self-hosting/flyio
- [ ] The Render self hosting guide: https://trigger.dev/docs/documentation/guides/self-hosting/render

apps/webapp/app/env.server.ts
Chigala: I'll fix this

Chigala:
Lol, I'll definitely look into this. Thanks

apps/webapp/app/env.server.ts
```diff
@@ -27,7 +27,10 @@ const EnvironmentSchema = z.object({
   AUTH_GITHUB_CLIENT_SECRET: z.string().optional(),
   FROM_EMAIL: z.string().optional(),
   REPLY_TO_EMAIL: z.string().optional(),
-  RESEND_API_KEY: z.string().optional(),
+  SMTP_HOST: z.string().optional(), 
+  SMTP_PORT: z.number().optional(),
```
matt-aitken: @Chigala Environment variables always come through as strings, so you need to do this:

```suggestion
SMTP_PORT: z.coerce.number().optional(), 
```

matt-aitken: I'm going to just make this change now so I can merge

Chigala: Thanks

matt-aitken (approved):
This works! I had to make a small change to coerce the env var string to a number

matt-aitken:
@Chigala to merge this that checklist above from @ericallam needs to be completed

Chigala:
@matt-aitken I've already sent pull requests to the docker and fiy.io repositories.

ericallam (approved):

packages/emails/src/index.tsx
```diff
@@ -42,15 +42,29 @@ export const DeliverEmailSchema = z
...
-  constructor(config: { apikey?: string; imagesBaseUrl: string; from: string; replyTo: string }) {
-    this.#client =
-      config.apikey && config.apikey.startsWith("re_") ? new Resend(config.apikey) : undefined;
+  constructor(config: {
+    smtpConfig?: NodeMailerTransportOptions;
+    imagesBaseUrl: string;
+    from: string;
+    replyTo: string;
+  }) {
+    this.#client = config.smtpConfig ? nodemailer.createTransport(config.smtpConfig) : undefined;
```
ericallam: this `smtpConfig` option is always going to be "truthy"  so the `nodemailer.createTransport` thing is always going to be called:

CleanShot 2023-08-21 at 16 14 58

ericallam:
@Chigala when you get a chance could you check my comment above?

packages/emails/src/index.tsx
Chigala: I'm sorry but I don't quite understand this

Chigala: @ericallam  Should I put back the resend api key and dependency as a fallback to when the smtp config doesn't exist?

nicktrn: I think Eric just meant the config values should be checked instead of the entire object. Otherwise, this will never return `undefined`. Checking for truthy `smtpConfig.host` should probably be enough.

Chigala: Ohh, makes sense. Thanks

gautamsi:
is this planned? I can recreate the PR and update the code if @Chigala is not available

ghostdevv:
also happy to help work on a revival of this :pray:

matt-aitken:
Thanks for this PR @Chigala, sorry it's taken so long. There's a new PR inspired by this #1580.

Issue #305, opened 2023-08-11
[TRI-1006] New `@trigger.dev/cli` whoami command. 


ericallam:
Run it in a configured project and it will print out information about who you currently are. It will use the env vars (similar to how the dev command does) and the `trigger.dev` config data in package.json to print out useful information about the current setup. Some prior art:

https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-auth-whoami
https://vercel.com/docs/cli/whoami

<sub>[TRI-1006](https://linear.app/triggerdotdev/issue/TRI-1006/new-triggerdevcli-whoami-command)</sub>

neo773:
@ericallam can we close this?

PR #316, opened 2023-08-11
[TRI-1006] New @trigger.dev/cli whoami command.

neo773:
/claim #305
This PR adds `whoami` command to trigger-cli
Couldn't find a way to get the user email since the API route doesn't return it

https://github.com/triggerdotdev/trigger.dev/assets/62795688/076c2c1a-66a5-46b5-8140-8578924a23a9

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -18,10 +18,10 @@ import { TriggerApi } from "../utils/triggerApi.js";
 const asyncExecFile = util.promisify(childProcess.execFile);
 
 export const DevCommandOptionsSchema = z.object({
-  port: z.coerce.number().optional(),
+  port: z.coerce.number(),
   envFile: z.string(),
-  handlerPath: z.string().optional(),
-  clientId: z.string().optional(),
+  handlerPath: z.string(),
+  clientId: z.string(),
 });
 
 export type DevCommandOptions = z.infer<typeof DevCommandOptionsSchema>;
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/dev.ts
```diff
@@ -21,7 +21,7 @@ export const DevCommandOptionsSchema = z.object({
   port: z.coerce.number(),
   envFile: z.string(),
   handlerPath: z.string(),
-  clientId: z.string().optional(),
+  clientId: z.string(),
```
matt-aitken: This change will break the dev command. 

You can do `type WhoAmIOptions = Required<DevCommandOptions>` or just create a new type for WhoAmI.

packages/cli/src/commands/whoami.ts
```diff
@@ -0,0 +1,68 @@
...
+  envFile: z.string()
+});
+
+export type WhoAmICommandOptions = z.infer<typeof WhoAmICommandOptionsSchema>;
+
+export async function whoamiCommand(path: string, anyOptions: any) {
+  const loadingSpinner = ora(`Hold while we fetch your data`);
+  loadingSpinner.start();
+
+  telemetryClient.dev.started(path, anyOptions);
```
matt-aitken: Remove all the telemetry calls from this file. We use them to track that the commands are running but only for critical commands, and this one is less important to track.

matt-aitken (changes requested):
This is good. A couple of changes required

packages/cli/src/commands/dev.ts
neo773: @matt-aitken  Yeah, I realized that later, check the commit below that one named "revert zod schema for dev.ts"

packages/cli/src/commands/whoami.ts
neo773: @matt-aitken Done, just pushed the commit.

packages/cli/src/commands/dev.ts
matt-aitken: The latest code has still made this optional

matt-aitken: I fixed it

matt-aitken (approved):

PR #317, opened 2023-08-11
fix: Add an update sub-command the @trigger.dev/cli that updates all @trigger.dev/* packages 

### LINKED ISSUE #166: [TRI-654] Add an `update` sub-command the @trigger.dev/cli that updates all @trigger.dev/* packages

rishi-raj-jain:
Resolves #166
/claim #166


packages/cli/package.json
```diff
@@ -53,6 +53,7 @@
   },
   "dependencies": {
     "@types/degit": "^2.8.3",
+    "axios": "^1.4.0",
```
wesleymatosdev: The project is on node v18 already; can you use native `fetch` instead?

rishi-raj-jain: Even though the project is using fetch, this'd be packaged in a CLI. I've used node-fetch as per https://github.com/triggerdotdev/trigger.dev/pull/174#discussion_r1271930801

rishi-raj-jain:
I've tested this with both npm & pnpm as well.

@ologbonowiwi would you like to confirm?

ologbonowiwi:
Can you give some guidance on how to test it, @rishi-raj-jain? I'm pretty new here, but I can try if you give me the step-by-step

ologbonowiwi (approved):
Tested here and it works locally

rishi-raj-jain:
Thanks for the confirmation! @ericallam Can you take a look at this now that we've confirmed working of this?

rishi-raj-jain:
Thanks for the confirmation! @ericallam Can you take a look at this now that we've confirmed working of this?

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,78 @@
...
+        return { packageName, latestVersion };
+      } catch (error) {
+        // @ts-ignore
+        console.error(`Error fetching version for ${packageName}: ${error.message}`);
+        return null;
+      }
+    })
+  );
+  const packagesToUpdate = newVersions.filter(
+    (entry) => entry !== null && entry.latestVersion !== dependencies[entry.packageName]
```
matt-aitken: It's most common to not have exact versions specified in the package.json file.

For example, after run the CLI init command you end up with this:

```json
"@trigger.dev/sdk": "^2.0.8",
"@trigger.dev/nextjs": "^2.0.8"
```

It's no good comparing the strings of the versions, even if you use something like semver comparison because you can't tell what's actually installed.

You need to do something like npm list to do this. That tells you which packages are actually installed on that machine.

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,78 @@
...
+    console.log("No @trigger.dev/* packages found in package.json.");
+    return;
+  }
+
+  const newVersions = await Promise.all(
+    triggerPackages.map(async (packageName) => {
+      try {
+        const response = await fetch(`https://registry.npmjs.org/${packageName}`);
+        const data = await response.json();
+        const latestVersion = data["dist-tags"].latest;
```
matt-aitken: There's a Typescript error here because data is unknown.


This fixes it: 
```suggestion
const schema = z.object({
          "dist-tags": z.object({
            latest: z.string(),
          }),
        });

        const parsed = schema.parse(data);
        return { packageName, latestVersion: parsed["dist-tags"].latest };
```

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  if (packagesToUpdate.length === 0) {
+    console.log("All @trigger.dev/* packages are up to date.");
+    return;
+  }
+
+  console.log("Newer versions found for the following packages:");
+  packagesToUpdate.forEach((entry) => {
+    console.log(
+      `- ${entry.packageName}: current ${dependencies[entry.packageName]} -> latest ${
```
matt-aitken: There are TypeScript errors here because entry can be null. The easiest fix would be to do a null check before line 53.

matt-aitken (changes requested):
This is mostly there. But there's an issue with how you're evaluating whether packages are the same or not.

You can test CLI changes like this by following from line 123 in the Contributions guide.

rishi-raj-jain:
Thanks for the detailed review, @matt-aitken, I've updated.

rishi-raj-jain:
@matt-aitken @ericallam Did you get a chance to look at this one?

rishi-raj-jain:
@matt-aitken @ericallam Did you get a chance to look at this one?

rishi-raj-jain:
@matt-aitken @ericallam Did you get a chance to look at this one?

matt-aitken:
@rishi-raj-jain the code looks good, but I think we should use this package (recommended by another developer): https://www.npmjs.com/package/npm-check-updates

This would massively simplify the code here and will be more reliable as that package has a lot of contributors and high usage.

You can see how it's used as a package at the bottom of the readme. You can specify wildcards to get all the `@trigger.dev/` packages.

I'll tip you generously if you make this change because we've messed you around a bit here. Sorry about that.

rishi-raj-jain:
Sure, I'll be on it. Appreciate your kindness ☺️

rishi-raj-jain:
Hey @matt-aitken,

For this won't the user need to install the npm-check-updates package on their device?

Also in the environment that don't provide access to installing global packages would be affected?

rishi-raj-jain:
Happy to implement the package if you want me to. I'm fine with that as well. Just lmk.

rishi-raj-jain:
Happy to implement with the package suggested if you want me to. I'm fine with that as well. Just lmk.

ericallam:
@rishi-raj-jain You can use `npm-check-updates` programatically:

CleanShot 2023-08-21 at 16 16 56

So you just need to add `npm-check-updates` as a dep of the cli package and then use it in the code just like any other package.

rishi-raj-jain:
@ericallam

Thanks for pointing me in that direction!

This is ready for review.

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,122 @@
+import fs from "fs";
+import path from "path";
+import inquirer from "inquirer";
+import { run, RunOptions } from "npm-check-updates";
+import { installDependencies } from "../utils/installDependencies.js";
+import { Index } from "npm-check-updates/build/src/types/IndexType.js";
```
ericallam: You don't need to import this `Index` type

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,122 @@
...
+
+function getPackageJSON(projectPath: string) {
```
ericallam: Please use the utility functions here for reading and writing JSON files:

https://github.com/triggerdotdev/trigger.dev/blob/main/packages/cli/src/utils/fileSystem.ts

packages/cli/src/commands/update.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  // Use npm-check-updates to get updated dependency versions
+  const ncuOptions: RunOptions = {
+    packageData,
+    upgrade: true,
+    jsonUpgraded: true,
+  };
+
+  // Can either give a json like package.json or just with deps and their new versions
+  const updatedDependencies: Index | void = await new Promise((resolve, reject) =>
```
ericallam: There is no need to explicitly type this or create a promise like you are doing. You can just do:

```ts
const updatedDependencies = await run(ncuOptions)
```

ericallam (changes requested):

packages/cli/src/commands/update.ts
rishi-raj-jain: What you're suggesting won't be doable as run is not exported as an async function from the library itself.

packages/cli/src/commands/update.ts
rishi-raj-jain: The Index type comes in handy to represent any key value pair index, shall I create an interface by myself in the file instead?

packages/cli/src/commands/update.ts
ericallam: Are you sure? According to their docs it is

rishi-raj-jain: Yes, I tried that. Wait let me share the function declaration.

rishi-raj-jain: Screenshot 2023-08-22 at 2 10 11 AM

rishi-raj-jain: It's a function that returns a promise, but isn't async. Hence, need to await the promise resolution.

ericallam: Any function that returns a promise can be awaited. You only use the `async` keyword when defining a function if you want to **use** await inside the function:

```ts
function foo(): Promise<string> {
  return new Promise((resolve, reject) => resolve("foo"));
}

async function bar() {
  const baz = await foo();

  console.log(baz); // logs "foo"
}
```

rishi-raj-jain: Oh, I see. Thanks for teaching me this. I just relied on the function declaration always.

rishi-raj-jain: I've pushed the code as taught.

packages/cli/src/commands/update.ts
rishi-raj-jain: Removed.

ericallam:
This is looking good! Can you please leave a comment on this PR the steps you've taken to test this change?

rishi-raj-jain:
1. So in the video, I've copy pasted the whole code into a test folder which has latest axios and node-fetch.
2. Then I downgrade axios & node-fetch to ^0 versions.
3. As utilites in your project don't exist in my test environment, as you can see in the video, I've quickly added some code to read and write json file.
4. Then I update the code to call the update command in the file (which in your case is mapped through the cli option).
5. Then I run the function, it comes back with the message of no trigger.dev packages to update.
6. Then I run the function after changing the selector to axios, and you can see it printing a table of old and new.
7.  Then I run the function after changing the selector to node-fetch, and you can see it printing a table of old and new.
8. I press yes, and the package.json is updated.
9. I expect the installDependencies to work with the utility you already have, hence haven't rewrote that in my test env.

ericallam:
Can you use our guide for testing CLI changes here:

https://github.com/triggerdotdev/trigger.dev/blob/main/CONTRIBUTING.md#testing-cli-changes

rishi-raj-jain:
@ericallam 

Thanks for merging the PR.

https://github.com/triggerdotdev/trigger.dev/pull/317#issuecomment-1683711591

Just saying that in that comment, I was told that I'd get a generous tip. Would be good to hear from you regarding the same.

rishi-raj-jain:
Thank you so much for considering that, @ericallam! 🙏🏻

Issue #310, opened 2023-08-11
[TRI-1007] Better handle retrying rate-limited resend requests in `@trigger.dev/resend`


ericallam:
Resend has a rate limit of 10 requests per second. They also send [Rate limit headers](https://resend.com/docs/api-reference/introduction#rate-limit) with every request, meaning we should be able to use the same method as we do in our GitHub integration to only retry a sendEmail task when it will actually succeed, instead of doing a backoff strategy.

The `sendEmail` task is defined here: 

https://github.com/triggerdotdev/trigger.dev/blob/main/integrations/resend/src/index.ts#L10

You can see how we schedule retries using rate limiting headers in our github integration here:

https://github.com/triggerdotdev/trigger.dev/blob/c246578a8123cdd53bed819bee33cb16c85df561/integrations/github/src/tasks.ts#L18

<sub>[TRI-1007](https://linear.app/triggerdotdev/issue/TRI-1007/better-handle-retrying-rate-limited-resend-requests-in)</sub>

RamK777-stack:
/attempt https://github.com/triggerdotdev/trigger.dev/issues/305

rishi-raj-jain:
@samejr 

I think this issue is resolved?

ericallam:
@rishi-raj-jain 

This is being help up due to the following issue: https://github.com/resendlabs/resend-node/issues/189

matt-aitken:
v2 is deprecated as of Jan 31 2025

PR #320, opened 2023-08-11
Better handle retrying rate-limited resend requests

Akshay-Patel-dev:
/claim #310 
ref: #310 
instead of using a simple backoff strategy,  uses sendEmail code to handle retries based on rate limits.


integrations/resend/src/index.ts
```diff
@@ -1,17 +1,43 @@
 import type { IntegrationClient, TriggerIntegration } from "@trigger.dev/sdk";
 import { Resend as ResendClient } from "resend";
+import { RequestError } from "@octokit/request-error";
```
matt-aitken: We shouldn't have a dependency on the GitHub octokit package here. 

Instead you should import the `ErrorResponse` from the resend package. Pretty sure it's exported.

matt-aitken (changes requested):

integrations/resend/src/index.ts
Akshay-Patel-dev: Hi @matt-aitken ,

I attempted to import ErrorResponse from the Resend package; however, I couldn't find it available within the Resend package.

Akshay-Patel-dev:
I explored an alternative approach. Could you please review the alternative method I've employed?

integrations/resend/src/index.ts
matt-aitken: If you update the resend package to the latest, they now export `ErrorResponse`

integrations/resend/src/index.ts
```diff
@@ -1,17 +1,44 @@
...
 
 type SendEmailData = Parameters<InstanceType<typeof ResendClient>["sendEmail"]>[0];
 
 type SendEmailResponse = { id: string };
 
+function isRequestError(error: unknown): boolean {
+  return typeof error === "object" && error !== null && "status" in error;
+}
+
+function onError(error: any) {
```
ericallam: The types for this are only working because you are typing the error as any here, it should be changed to `unknown` and the type guard above (`isRequestError`) should correctly guard to the error type in the Resend SDK.

integrations/resend/src/index.ts
```diff
@@ -33,13 +60,6 @@ export const sendEmail: AuthenticatedTask<
         },
         ...subjectProperty,
       ],
-      retry: {
```
ericallam: Don't remove this, these retry options will still be used in the `onError` callback returns undefined.

ericallam (changes requested):

integrations/resend/src/index.ts
Akshay-Patel-dev: Okay, sure

matt-aitken:
@ericallam I don't think we can use the Resend ErrorResponse type, it doesn't expose the headers of a value for the rate limit remaining. Only `ErrorResponse` is exported from here: https://github.com/resendlabs/resend-node/blob/main/src/interfaces.ts

@Akshay-Patel-dev it's hard for you to test without a Resend API key. We'll test the latest and get back to you

ericallam:
Resend have a generous free-tier so it should be easy enough to get an API key, right?

ericallam:
You don't need to get at the headers:

CleanShot 2023-08-24 at 09 46 41

Akshay-Patel-dev:
@ericallam I believe we need to determine the remaining time until the rate limit is reset before attempting another request.


If we create custom interface  using existing Resend **ErrorResponse interface** we can use it.

  if (error.status === 429) {
    const rateLimitReset = error.response.headers["x-ratelimit-reset"];

    if (rateLimitReset) {
      const resetDate = new Date(Number(rateLimitReset) * 1000);

      return {
        retryAt: resetDate,
        error,
      };
    } 
  }

Akshay-Patel-dev:
Hi @ericallam  please review and accept  this PR.

Akshay-Patel-dev:
> Hi @ericallam  please review  this PR.

matt-aitken (approved):
This is great 👍

PR #322, opened 2023-08-11
handles retry ratelimited resend request

Chigala:
## Description 
Made it so requests only retries if the rate limiting header indicates that it's permissible to do so. you should be able to send the sendEmail requests freely, and the retry mechanism will engage automatically when conditions allow.

fixes #320 
/claim #320


Akshay-Patel-dev:
Thank you @Chigala,

matt-aitken:
@Chigala closing this because @Akshay-Patel-dev was first on this one.

PR #323, opened 2023-08-11
fix: drop fetch fork dependency for native Request in Node.js

### LINKED ISSUE #303: Removing the fetch fork from the Express package
## Expected results

The Express package at `@trigger.dev/express` uses the remix fork of a fetch implementation which IMHO adds an unnecessary dependency that is yet one more thing to maintain and worry about, and unnecessarily adds more code handling and import that don't help with readability.

Is there a specific reason to rely on `@remix-run/web-fetch` ? Node.js natively supports fetch.

## Actual results

I'm in favor of keeping things simple and straight-forward and would want to propose a code update that refactors the dependency out and implements the headers copy for the trigger.dev handshake via vanilla Node.js / JS code.

lirantal:
Fixes #303 for when the decision is made to drop the fetch fork dependency

### DIFF
--- a/packages/express/package.json
+++ b/packages/express/package.json
@@ -35,7 +35,6 @@
     "@trigger.dev/sdk": "workspace:^2.0.8"
   },
   "dependencies": {
-    "@remix-run/web-fetch": "^4.3.5",
     "debug": "^4.3.4",
     "express": "^4.18.2"
   },
--- a/packages/express/src/index.ts
+++ b/packages/express/src/index.ts
@@ -1,6 +1,5 @@
 import express, { Express } from "express";
 import { TriggerClient } from "@trigger.dev/sdk";
-import { Request as StandardRequest, Headers as StandardHeaders } from "@remix-run/web-fetch";
 
 /**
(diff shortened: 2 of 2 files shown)


matt-aitken:
@lirantal thanks for this, when we moved to Node v18+ we'll merge this 👍

matt-aitken (changes requested):
I'm getting an error when running this:

```
ReferenceError: Request is not defined
    at /Users/matt/Work/Git/APIHero/examples/express-vanilla/server.js:90:21
```

Which seems to point to that the Request class isn't imported. 

```
    const request = new Request("https://express.js/api/trigger", {
```

lirantal:
@matt-aitken which version of Node.js are you running?

matt-aitken:
@lirantal v18.12.1

lirantal:
the global [Request](https://nodejs.org/api/globals.html#request) is supported in >= 18.0.0

image

lirantal:
@matt-aitken is it possible that's some typescript related error?

ericallam:
I'll test this again when we do the migration to Node v18 👍

Issue #206, opened 2023-07-22
[TRI-712] Create an ESLint plugin to make sure task keys are unique inside a Job run function


ericallam:
Accidentally using duplicate task keys inside an individual job can lead to unexpected results (see our docs for more info: https://trigger.dev/docs/documentation/concepts/resumability). It would be great to create an ESLint plugin that can detect this.

<sub>[TRI-712](https://linear.app/triggerdotdev/issue/TRI-712/create-an-eslint-plugin-to-make-sure-task-keys-are-unique-inside-a-job)</sub>

Ashutosh-Bhadauriya:
@ericallam just to confirm. I should be doing this inside `config-packages/eslint-config-custom`, right?

b4s36t4:
@ericallam could you please verify below test-cases should works for the plugin-rule or maybe I'm missing any rule.

<details>
<summary>Test Cases</summary>

```javascript
/**
 * @fileoverview No Unique key
 * @author no-unique-key
 */
"use strict";

//------------------------------------------------------------------------------
// Requirements
//------------------------------------------------------------------------------

const rule = require("../../../lib/rules/no-unique-key"),
  RuleTester = require("eslint").RuleTester;

//------------------------------------------------------------------------------
// Tests
//------------------------------------------------------------------------------

const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018 } });
ruleTester.run("no-unique-key", rule, {
  valid: [
    // give me some code that won't trigger a warning
    {
      code: `
      new Job(client, {
        id: "event-1",
        name: "Run when the foo.bar event happens",
        version: "0.0.1",
        trigger: eventTrigger({
          name: "foo.bar",
        }),
        run: async (payload, io, ctx) => {
          await io.slack.postMessage("⭐️ New Star", {
            channel: "C04GWUTDC3W",
            text: 'hello'
          });
        },
      });
      `,
    },
    {
      code: `
      const runJob  = async (payload, io, ctx) => {
        await io.slack.postMessage("⭐️ New Star", {
          channel: "C04GWUTDC3W",
          text: 'hello'
        });
      }
      new Job(client, {
        id: "event-1",
        name: "Run when the foo.bar event happens",
        version: "0.0.1",
        trigger: eventTrigger({
          name: "foo.bar",
        }),
        run: runJob
      });
      `,
    },
  ],

  invalid: [
    {
      code: `
      new Job(client, {
        id: "event-1",
[...]

PR #324, opened 2023-08-12
feature/TRI 712/create eslint plugin to ensure uniqueness on task keys

ologbonowiwi:
/claim #206 

- test: add test to no-duplicated-task-keys rule
- feat: implement no-duplicated-task-keys
- chore: create testing scripts
- refactor: fill in rule meta
- feat: export no-duplicated-task-keys rule
- chore: bump package version to 0.0.1


ologbonowiwi:
@ericallam got the rule working (you can enter `config-packages/eslint-config-custom` and run `pnpm test` or `pnpm test:watch` to take a look at it working).

ologbonowiwi:
Although the rule was successfully implemented, I hadn't success in getting it to validate an actual code. It would be nice to get some help on doing so, if possible 😄

matt-aitken:
@ologbonowiwi this is great. I think the idea was that this would be a published package, so it would live in the `packages` folder and be public. That way we could get the CLI to install it so people can use it. (We'd have manual installation as well).

ologbonowiwi:
@matt-aitken fixed it to match what we currently need.

To test it, you can enter any `examples/nextjs-example` with access to jobs, create a job with a duplicated key and run `pnpm lint`.

The expectation is that you'll see the error on your console.

PR #325, opened 2023-08-12
Resolved TRI-1007-Better handle retrying rate-limited resend requests

### LINKED ISSUE #310: [TRI-1007] Better handle retrying rate-limited resend requests in `@trigger.dev/resend`

MAYANK-TRIPATH:
Fixes #310 
/claim #310

### DIFF
--- a/integrations/resend/src/index.ts
+++ b/integrations/resend/src/index.ts
@@ -3,6 +3,8 @@ import { Resend as ResendClient } from "resend";
 
 import type { AuthenticatedTask } from "@trigger.dev/sdk";
 
+import { RequestError } from "@octokit/request-error";
+
 type SendEmailData = Parameters<InstanceType<typeof ResendClient>["sendEmail"]>[0];
 
 type SendEmailResponse = { id: string };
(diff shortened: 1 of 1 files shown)


MAYANK-TRIPATH:
@ericallam Try to solved in more optimised way also added comments on my changes. Let me know if anything need improvement. Thanks

matt-aitken:
@MAYANK-TRIPATH closed this because another PR was first

MAYANK-TRIPATH:
@matt-aitken Okay no worries :)

Issue #263, opened 2023-08-04
[TRI-887] Tests for the CLI


matt-aitken:
### Tests that

* The CLI init command succeeds or fails as appropriate
  * The project is correctly detected
  * The correct files have been created and configured correctly
  * The project builds, i.e. no syntax errors
* The CLI dev command succeeds or fails as appropriate

### Test situations

A separate public GitHub repo which has lots of directories in the root, each one is a project (just Next.js for now but soon Express etc). The first step in the e2e test would be to download the repo and then iterate through the folders.

* Different package managers
* Next.js App and pages directories
* Next.js 12 and 13

### Inspiration

[https://github.com/gmrchk/cli-testing-library](https://github.com/gmrchk/cli-testing-library)

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-887](https://linear.app/triggerdotdev/issue/TRI-887/tests-for-the-cli)</sub>

ologbonowiwi:
I would like to take a shot on this one.

ologbonowiwi:
The first step to enable end-to-end tests would be setting up a test environment on the `cli` package.

I'll open a PR running a simple test (`expect(true).toBe(true)`), but ensuring that we're good in terms of configurations with Jest;
I'll do this to be sure that in upcoming test implementations, we'll need to care only about actual test assertions rather than getting everything working.

ologbonowiwi:
#326 opened. I'm still familiarizing myself with `@gmrchk/cli-testing-library`, but from what I found so far, we'll probably need to run commands like on their examples `node ./my-cli something`, but we'll need to pass the next js project path as parameter to the CLI.

I left one example file doing this on the PR.

I hope it helps 😄

ologbonowiwi:
> A separate public GitHub repo which has lots of directories in the root, each one is a project (just Next.js for now but soon Express etc). The first step in the e2e test would be to download the repo and then iterate through the folders.

One nice way we could do so would be by creating these projects where the CLI needs to fit, and we can iterate over them on the tests making the assertions. If you look [here](https://github.com/triggerdotdev/trigger.dev/pull/326/files#diff-0160ffc15207546cc8e2aca5c3a6f75b9b64f3a43959c94de95f90c6e6bdc2e1R18), you'll see that we have a NextJS path to pass to the commands. Instead of having one, we would iterate over all paths in `examples` files and run. The test suite multiple times, ensuring the same behavior over all projects.

[...]

PR #326, opened 2023-08-13
test/TRI 887/setup jest on cli package

ologbonowiwi:
solves part of #263

- chore: run `pnpm install --prefer-frozen-lockfile`
- chore: install `jest` and `ts-jest`
- test: create `jest.config.js`
- chore: install `@types/jest`
- chore: add `jest` on `tsconfig.json` `types` property
- chore: create test script
- test: create example test 🎉🎊🕺🏼
- chore: install `@gmrchk/cli-testing-library`
- chore: override `tsconfig` with `@trigger.dev/tsconfig`


packages/cli/test/example.test.ts
```diff
@@ -0,0 +1,39 @@
...
+afterAll(async () => {
+  await environment.cleanup();
+});
+
+describe('cli', () => {
+  // can be any path with a nextjs project
+  const NEXT_PROJECT_PATH = join(__dirname, '..', '..', '..', 'examples', 'nextjs-example');
+
+  it('should be able to execute cli', async () => {
+    const { waitForText, getStdout, wait, pressKey } = await environment.spawn('node', `${join(__dirname, '..', 'dist', 'index.js')} init -p ${NEXT_PROJECT_PATH}`)
```
wesleymatosdev, line 21: In this case, I'm unsure if we should use `node` and execute from `dist/index.js` or run directly from the typescript file.

I noticed that we use `tsup`, so if we use `node` to execute, probably we'll need to `pnpm dev` and then `pnpm test`

packages/cli/test/example.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    await waitForText('Are you using the Trigger.dev cloud or self-hosted?');
+
+    console.log('getStdout() :>> ', getStdout());
+
+    await pressKey('enter');
+
+    console.log('getStdout() :>> ', getStdout());
+
+    // wait next prompt and keep going
```
wesleymatosdev: Here is where we would start to assert behaviors and make validations that everything is working fine

packages/cli/test/example.test.ts
matt-aitken: This is the correct approach. 

You just need to add a dependency in the [turbo.json](https://github.com/triggerdotdev/trigger.dev/blob/12de1b0d34911e568505f417d0dfb8831ccbc6d4/turbo.json#L49) file. This should look like this:

```json
"test": {
    "dependsOn": ["build"],
    "outputs": []
  },
```

matt-aitken: That will ensure that it is built first. It will also rebuild automatically if a file has changed, and cache appropriately.

matt-aitken (changes requested):
This is great! If you add the line to the turbo.json file I think we should merge this so separate issues can be created for different useful tests

ologbonowiwi:
Changes addressed; can you re-review @matt-aitken?

matt-aitken (approved):

PR #327, opened 2023-08-13
docs: Update runs.mdx

lirantal:
From what I can tell, it is expected for run functions to return with some sort of payload. Providing an example here as a reference and guideline.


matt-aitken (approved):
Good point, I've made it clear this this optional and added some links to how you can retrieve the data using the SDK/react hooks.

lirantal:
Nice! thanks.

PR #328, opened 2023-08-13
docs: Update integrations.mdx

lirantal:
The first sentence in this paragraph for `### API Keys and Tokens` was missing a word or two and it wasn't entirely clear what did the original author of this docs page intended.

I've entered what seems to be the missing context but happy to re-phrase too if another wording makes better sense here.


matt-aitken (approved):
Yep this wasn't clear, I've tweaked what you did a bit to be more explicit.

PR #332, opened 2023-08-14
feat: added the ability to cancel scheduled Events based on the event ID

### LINKED ISSUE #230: Cancel scheduled events
Currently, you're able to schedule events like so:
```
const eventId = await client.sendEvent({
  name: "user.created",
  payload: { name: "John Doe", email: "john@doe.com", paidPlan: true },
  deliverAt: new Date('2024'),
});
```

I want the ability to cancel scheduled events.

Either an identifier should be provided as an argument to the function or returned by it so that it can later be referenced when the scheduled event should be canceled.

Chigala:
## Description
Added a function to cancel scheduled Events based on their ID. 

## Changes
- added an API endpoint `api.v1.events.$eventId.cancel.ts` to update the cancelledAt column in the EventRecord table to the time the event was cancelled. 
- add a dequeue public function in the Zodworker class that handles removal of already scheduled jobs by calling the `graphile_worker.remove_job` SQL to remove the job.  
- Added a sendEvent function in the API Client class. 

fixes #230 
/claim #230

### DIFF
--- a/apps/webapp/app/platform/zodWorker.server.ts
+++ b/apps/webapp/app/platform/zodWorker.server.ts
@@ -175,10 +175,9 @@ export class ZodWorker<TMessageCatalog extends MessageCatalogSchema> {
     if (!this.#runner) {
       throw new Error("Worker not initialized");
     }
+
     const job = await this.#removeJob(jobKey, option?.tx ?? this.#prisma);
-    logger.debug("dequeued worker task", {
-      job,
-    });
+    logger.debug("dequeued worker task", { job });
 
     return job;
   }
@@ -206,7 +205,7 @@ export class ZodWorker<TMessageCatalog extends MessageCatalogSchema> {
       spec.queueName || null,
       spec.runAt || null,
       spec.maxAttempts || null,
-      identifier,
+      spec.jobKey,
       spec.priority || null,
       spec.jobKeyMode || null,
       spec.flags || null
@@ -226,22 +225,25 @@ export class ZodWorker<TMessageCatalog extends MessageCatalogSchema> {
   }
 
   async #removeJob(jobKey: string, tx: PrismaClientOrTransaction) {
-    const result = await tx.$queryRawUnsafe(
-      `SELECT * FROM graphile_worker.remove_job(
+    try {
+      const result = await tx.$queryRawUnsafe(
+        `SELECT * FROM graphile_worker.remove_job(
           job_key => $1::text
         )`,
-      jobKey
-    );
-
-    const job = GraphileJobSchema.safeParse(result);
-
-    if (!job.success) {
-      throw new Error(
-        `Failed to remove job from queue, zod parsing error: ${JSON.stringify(job.error)}`
+        jobKey
       );
-    }
+      const job = AddJobResultsSchema.safeParse(result);
 
-    return job.data as GraphileJob;
+      if (!job.success) {
+        throw new Error(
+          `Failed to remove job from queue, zod parsing error: ${JSON.stringify(job.error)}`
+        );
+      }
+
+      return job.data[0] as GraphileJob;
+    } catch (e) {
+      throw new Error(`Failed to remove job from queue, ${e}}`);
+    }
   }
 
   #createTaskListFromTasks() {
--- a/apps/webapp/app/platform/zodWorker.server.ts
+++ b/apps/webapp/app/platform/zodWorker.server.ts
@@ -205,7 +205,7 @@ export class ZodWorker<TMessageCatalog extends MessageCatalogSchema> {
       spec.queueName || null,
       spec.runAt || null,
       spec.maxAttempts || null,
-      spec.jobKey,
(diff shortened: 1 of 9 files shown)


ericallam:
Thanks for this @Chigala, will be reviewing tomorrow 👍

RamK777-stack:
@Chigala I am curious to know, how you have tested sendEvent with some future date in the deliverAt key. because the job immediately runs for me. added my doubt in https://github.com/triggerdotdev/trigger.dev/issues/230#issuecomment-1677631632. please guide me on this.

Chigala:
Hello @RamK777-stack  I believe @ericallam  has already answered you. You weren't calling the sendEvent function correctly. That's why the job was running immediately.

apps/webapp/app/platform/zodWorker.server.ts
```diff
@@ -190,7 +205,7 @@ export class ZodWorker<TMessageCatalog extends MessageCatalogSchema> {
       spec.queueName || null,
       spec.runAt || null,
       spec.maxAttempts || null,
-      spec.jobKey || null,
+      spec.jobKey,
```
ericallam: Let's keep `jobKey` as optional for `addJob`

apps/webapp/app/routes/api.v1.events.$eventId.cancel.ts
```diff
@@ -0,0 +1,71 @@
...
+    where: {
+      id: eventId,
+      environmentId: authenticatedEnv.id,
+    },
+  });
+  if (!event) {
+    return json({ error: "Event not found" }, { status: 404 });
+  }
+
+//update the cancelledAt column in the eventRecord table
```
ericallam: Can you update this to happen inside a transaction? See this file as an example: https://github.com/triggerdotdev/trigger.dev/blob/improvements/job-run-performance/apps/webapp/app/services/events/deliverEvent.server.ts

ericallam (changes requested):
Looks pretty close! Can you also add a `cancelEvent` method to the `IO` class, similar to `sendEvent`:

https://github.com/triggerdotdev/trigger.dev/blob/45c1578f2abb421f37df9eba2c3fa4d9083764a7/packages/trigger-sdk/src/io.ts#L203

Chigala:
> Looks pretty close! Can you also add a `cancelEvent` method to the `IO` class, similar to `sendEvent`:
> 
> https://github.com/triggerdotdev/trigger.dev/blob/45c1578f2abb421f37df9eba2c3fa4d9083764a7/packages/trigger-sdk/src/io.ts#L203

Yeah, I realised I had to add that too when I saw your code snippet on Discord. I'll implement it.

apps/webapp/app/services/events/cancelEvent.server.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(prismaClient: PrismaClientOrTransaction = prisma) {
+    this.#prismaClient = prismaClient;
+  }
+
+  public async call(environment: AuthenticatedEnvironment, eventId: string): Promise<EventRecord | undefined> {
+    return await $transaction(
+      this.#prismaClient,
+      async (tx) => {
+        const event = await tx.eventRecord.findFirst({
```
ericallam: This should be a call to `findUnique`, using the `eventId_environmentId` where:

```
tx.eventRecord.findUnique({
  where: {
    eventId_environmentId: {
      eventId: id,
      environmentId: env.id,
    }
  }
})
```

apps/webapp/app/services/events/cancelEvent.server.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        //update the cancelledAt column in the eventRecord table
+        const updatedEvent = await prisma.eventRecord.update({
+          where: { id: event.id },
+          data: { cancelledAt: new Date() },
+        });
+
+        // Dequeue the event after the db has been updated
+        await workerQueue.dequeue(event.id, { tx: prisma });
```
ericallam: I think the job_key should be `event:${event.id}` instead of just the event.id. This will need be matched by the call to enqueue.

apps/webapp/app/services/events/ingestSendEvent.server.ts
```diff
@@ -91,7 +88,7 @@ export class IngestSendEvent {
             {
               id: eventLog.id,
             },
-            { runAt: eventLog.deliverAt, tx }
+            { runAt: eventLog.deliverAt, tx, jobKey: eventLog.id }
```
ericallam: `jobKey` should be `event:${eventLog.id}`

ericallam (changes requested):
A few more requested changes. One other thing is this should be updated to check to make sure the event `cancelledAt` is not set. If it is set, it should just return:

https://github.com/triggerdotdev/trigger.dev/blob/improvements/job-run-performance/apps/webapp/app/services/events/deliverEvent.server.ts

PR #340, opened 2023-08-15
updated the docs and codebase to reflect changes in the io.runTask function

### LINKED ISSUE #338: [TRI-1068] Improve `io.runTask`: move callback to second arg and move "task options" to be an optional 3rd argument
Currently `io.runTask` looks like this:

```ts
await io.runTask("key", { name: "Task 1" }, async (task) => {});
```

Which is kind of ugly. We should make the second "task options" parameter by an optional 3rd argument so the above can be written like:

```ts
await io.runTask("key", async (task) => {});
```

And then if you want to specify task options:

```ts
await io.runTask("key", async (task) => {}, { name: "Task 1" });
```

<sub>[TRI-1068](https://linear.app/triggerdotdev/issue/TRI-1068/improve-ioruntask-move-callback-to-second-arg-and-move-task-options-to)</sub>

Chigala:
## Description
Moved the callback to the second argument and made the options argument optional

` await io.runTask("key", async (task) => {});`

`await io.runTask("key", async (task) => {}, { name: "Task 1" });`

## Changes
- updated the docs to reflect the new changes made 
- updated example projects in the repo to reflect the changes too. 

## Testing
- successfully built the application and removed all the errors caused by the changes. 
- made sure the new io.runTask() was working as expected in the packageTester repo. 

fixes #338

### DIFF
--- a/.changeset/pretty-files-rule.md
+++ b/.changeset/pretty-files-rule.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/sdk": minor
+---
+
+moved callback argument in the io.runTask function to be the 2nd argument and then moved "task options" to be the optional 3rd argument
--- a/apps/webapp/app/components/integrations/CustomHelp.tsx
+++ b/apps/webapp/app/components/integrations/CustomHelp.tsx
@@ -25,14 +25,15 @@ client.defineJob({
     //wrap an SDK call in io.runTask so it's resumable and displays in logs
     const repo = await io.runTask(
       "Get repo",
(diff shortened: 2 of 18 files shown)


matt-aitken:
@Chigala there's a related PR to this that's in progress with a bigger change to runTask. I'll leave this PR here and probably I'll cherry pick some changes from it, like docs and example updates.

matt-aitken:
This change was made as part of the Airtable integration PR. It was a useful reference to see places where updates were needed

Issue #280, opened 2023-08-08
[TRI-950] `io.runTask` return type of the callback has TypeScript error when it includes an `unknown` type


ericallam:
To recreate this issue:

```ts
await io.runTask("foo", { name: "Foo" }, async () => {
  return {
    foo: "bar",
  } as unknown;
});
```

That is because the return type of the callback is typed as required to extend `SerializedJson`:

https://github.com/triggerdotdev/trigger.dev/blob/fa942fc9f5aa1c2255efdb6304600be7f89cf20b/packages/trigger-sdk/src/io.ts#L434

We do this so the types of the return of `runTask` will match what is actually returned, since we run the results of the `runTask` callback through a JSON stringify/parsing as task output is stored in the database as JSON. Because of Resumability, this is a requirement, unless another solution can be identified.

I'm also now realizing that this SerializedJson might be incorrectly typed as well, as it can include `Date` and `Symbol`:

https://github.com/triggerdotdev/trigger.dev/blob/fa942fc9f5aa1c2255efdb6304600be7f89cf20b/packages/core/src/schemas/json.ts#L12

So this could cause issues when resuming a run and using the cached task output, where Date's would be symbols and the types would be wrong. It would be great to test this and possible change the required type to `DeserializedJson`. 

<sub>[TRI-950](https://linear.app/triggerdotdev/issue/TRI-950/ioruntask-return-type-of-the-callback-has-typescript-error-when-it)</sub>

MeenuyD:
Hello,@ericallam Can you explain a bit more

ologbonowiwi:
I was able to reproduce the issue but it is not clear what needs to be done so far. It would be nice to get some details

/attempt #280

yashug:
how about we having a wrapper something like this 
```
async function runTaskWrapper<TResult>(
  key: string | any[],
  options: RunTaskOptions,
  callback: (task: IOTask, io: IO) => Promise<TResult>
): Promise<SerializableJson> {
  const result = await io.runTask(key, options, async (task, io) => {
    const userResult = await callback(task, io);
    // Ensure that userResult is of SerializableJson type
    return userResult as SerializableJson;
  });

  return result;
}
```

ologbonowiwi:
@yashug even thought this would work, it's not safety right? We would be getting a type safety at transpile/development, but on runtime the data type could be different and bring unexpected behaviors

ologbonowiwi:
@ericallam, can you take a look at #342?

[...]

PR #342, opened 2023-08-15
feat: parse callback result with zod before return

ologbonowiwi:
/claim #280


matt-aitken (approved):
This works 👍  nice one

PR #349, opened 2023-08-17
test/add test on get user pkg manager

ologbonowiwi:
- chore: make possible to write tests aside with implementations
- test: write getUserPkgManager test TO-DOs
- test: add test to building the path
- test: add tests to manual file checking
- test: add tests to npm_config_user_agent

### DIFF
--- a/packages/cli/src/utils/getUserPkgManager.spec.ts
+++ b/packages/cli/src/utils/getUserPkgManager.spec.ts
@@ -1,7 +1,7 @@
 import { randomUUID } from "crypto";
 import { pathExists } from "./fileSystem";
 import { getUserPackageManager } from "./getUserPkgManager";
-import { join } from "path";
+import * as pathModule from "path";
 
 jest.mock('path', () => ({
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/getUserPkgManager.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it('should return "npm" if npm-shrinkwrap.json exists', async () => {
+      (pathExists as jest.Mock).mockImplementation((path: string) => path.endsWith('npm-shrinkwrap.json'));
+
+      expect(await getUserPackageManager(path)).toBe('npm');
+    });
+  });
+  
+  describe(`if doesn't found a artifacts, should use process.env.npm_config_user_agent to detect package manager`, () => {
```
wesleymatosdev:
```suggestion
  describe(`if doesn't found artifacts, should use process.env.npm_config_user_agent to detect package manager`, () => {
```

PR #350, opened 2023-08-17
refactor: remove `.js` extension from cli package

ologbonowiwi:


packages/cli/src/commands/createIntegration.ts
```diff
@@ -94,12 +94,12 @@ export async function createIntegrationCommand(path: string, cliOptions: any) {
     name: resolvedOptions.packageName,
     version: "0.0.1",
     description: `Trigger.dev integration for ${resolvedOptions.sdkPackage}`,
-    main: "./dist/index.js",
```
ericallam: these shouldn't be changed, only the imports

packages/cli/src/commands/init.ts
```diff
@@ -285,7 +285,7 @@ async function detectPagesOrAppDir(
   usesSrcDir = false,
   isTypescriptProject = false
 ): Promise<"pages" | "app"> {
-  const nextConfigPath = pathModule.join(path, "next.config.js");
```
ericallam: This shouldn't change either

packages/cli/src/commands/init.ts
```diff
@@ -445,7 +445,7 @@ async function createTriggerAppRoute(
   const tsConfigPath = pathModule.join(projectPath, configFileName);
   const { tsconfig } = await parse(tsConfigPath);
 
-  const extension = isTypescriptProject ? ".ts" : ".js";
```
ericallam: again don't change this

packages/cli/src/commands/init.ts
```diff
@@ -569,7 +569,7 @@ async function createTriggerPageRoute(
   const pathAlias = getPathAlias(tsconfig, usesSrcDir);
   const routePathPrefix = pathAlias ? pathAlias + "/" : "../../";
 
-  const extension = isTypescriptProject ? ".ts" : ".js";
```
ericallam: or this

packages/cli/src/utils/detectNextJsProject.ts
```diff
@@ -14,7 +14,7 @@ export async function detectNextJsProject(path: string): Promise<boolean> {
 
 async function detectNextConfigFile(path: string): Promise<boolean> {
   return fs
-    .access(pathModule.join(path, "next.config.js"))
+    .access(pathModule.join(path, "next.config"))
```
ericallam: don't change this

ericallam (changes requested):
I don't see any changes here that change the tsconfig or whatever was causing these .js extensions to be added. Let's make sure these don't get added back in whenever a file is saved

ericallam (approved):

Issue #345, opened 2023-08-16
Feat: do we fancy a native Astro integration?


lirantal:
Hey folks!

Similar to the Express and Next.js packages, do we want to have a native Astro integration package as part of the officially provided packages from Trigger.dev?

See the following resources for more context:
- The Astro integration package: https://github.com/lirantal/triggerdev-astro-integration
- An example Astro project built with the above integration package: https://github.com/lirantal/trigger.dev-astro-example

Of course, that would be amazing! Here's everything I can think of at the moment that would be nice to do to get astro support in (this isn't to say you should do these, just that they should be done at some point):

```[tasklist]
### Tasks
- [x] #354 Add astro adapter package to packages in the monorepo as @trigger.dev/astro
- [ ] https://github.com/triggerdotdev/examples/pull/5 Add example Astro project to the examples repo: https://github.com/triggerdotdev/examples
- [x] Add a Quick start guide to the Trigger.dev/docs
- [x] Add support for astro in @trigger.dev/cli init
```

ericallam:
Of course, that would be amazing! Here's everything I can think of at the moment that would be nice to do to get astro support in (this isn't to say you should do these, just that they should be done at some point):

- [ ] Add astro adapter package to `packages` in the monorepo as `@trigger.dev/astro` 
- [ ] Add example Astro project to the examples repo: https://github.com/triggerdotdev/examples
- [ ] Add a Quick start guide to the Trigger.dev/docs
- [ ] Add support for astro in `@trigger.dev/cli init`

lirantal:
Sounds like a good list. I might just tackle them as one-by-one PRs to make it scoped and easier to review and iterate while working to land these in officially.

lirantal:
@ericallam I think you can remove the bullet-list you created as I copied it over to my opening comment body as a tasklist so it is easier to track.

On that note, see first PR introduced for the astro package.

ericallam:
@lirantal I detailed in a github discussion [here](https://github.com/triggerdotdev/trigger.dev/discussions/411) all the steps it would take to "officially" support another framework for Trigger.dev, thought you might find it interesting for the Astro work.

lirantal:
Looks like this is now already shipped thanks to:
- [x] https://github.com/triggerdotdev/trigger.dev/pull/354
- [x] https://github.com/triggerdotdev/trigger.dev/issues/452

So I'd imagine we can close?

matt-aitken:
[...]

PR #354, opened 2023-08-17
feat: introduce support for native astro integration package

lirantal:
Implement #345 - astro integration

Signed-off-by: Liran Tal <liran.tal@gmail.com>


packages/astro/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License Copyright (c) 2023 Liran Tal
```
ericallam: This made me realize that we don't have individual licenses currently for our public packages, which has been fixed in this commit:

https://github.com/triggerdotdev/trigger.dev/commit/b5db9f5e5cf121e37b6350bf28fb64dc81ec949b

Do you think you could update this LICENSE file to match the others with our copyright? I think our policy will be that if it's in the monorepo then the LICENSE should be under Trigger.dev. If that doesn't work for you, then it's also okay for it to live as a standalone package (although not under the `@trigger.dev` npm org.

packages/astro/main.js
```diff
@@ -0,0 +1,48 @@
+export function createAstroRoute(client) {
```
ericallam, line 1: Ideally this package would be written in TypeScript to most of our other code

packages/astro/README.md
```diff
@@ -0,0 +1,34 @@
+# Trigger.dev native integration add-on for Astro
+
+This Astro integration project provides a `createAstroRoute` function that can be used to create an API endpoint route in your Astro project for integration with the hosted Trigger.dev platform.
+
+## Usage
+
+1. Install the package:
+
+```bash
+npm install --save https://github.com/lirantal/triggerdev-astro-integration
```
ericallam: These instructions should be updated to reference the final package `@trigger.dev/astro`

packages/astro/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/lirantal/triggerdev-astro-integration.git"
```
ericallam: These urls should be updated as well

ericallam (changes requested):
Thanks for this, it's a great start! Just added a few comments, let me know what you think.

packages/astro/README.md
lirantal: Fixed

packages/astro/LICENSE
lirantal: Yes, fixed!

packages/astro/main.js
lirantal: @ericallam I'm going to leave that part for you folks as TS isn't my thing and you know best how and what you want to do  ;-)

packages/astro/package.json
lirantal: Fixed

lirantal:
@ericallam addressed your comments, excluding the TS part.

ericallam (approved):

PR #356, opened 2023-08-17
fix: avoid code duplication by repostioning the default switch case

lirantal:
Small code clean up (on the borderline of nitpick so we can dismiss if not up to your taste).

This change proposes handling the same case for 'default' and 'completed'. I figured it'd be easier on the eye this way rather than to create a new variable and re-use it.


matt-aitken (approved):
Nice, I've never considered putting default anywhere but last. This is a neat trick.

PR #358, opened 2023-08-18
added @trigger.dev/react on CLI init and the necessary documentation

### LINKED ISSUE #330: [TRI-1018] Add the `@trigger.dev/react` package when doing `@trigger.dev/cli init`
We should also add the a bit about setting up React hooks from [this guide](https://trigger.dev/docs/documentation/guides/react-hooks) to the Next.js quickstart guide

<sub>[TRI-1018](https://linear.app/triggerdotdev/issue/TRI-1018/add-the-triggerdevreact-package-when-doing-triggerdevcli-init)</sub>

### LINKED ISSUE #331: [TRI-1029] `@trigger.dev/react` doesn't work out of the box with Next.js Page Router: Cannot use import statement outside a module Error
When using `@trigger.dev/react` in Next.js Page Router, importing will cause an error because Page Router modules aren't automatically transpiled to CJS (whereas they are in App Router).

From Next 13.1 onwards, the fix is to the add `@trigger.dev/react` to the list of transpiled packages in the next.config.js file:

```js
const nextConfig = {
  transpilePackages: ["@trigger.dev/react"],
};

module.exports = nextConfig;
```

In Next versions below 13.1, the [next-transpile-modules](https://www.npmjs.com/package/next-transpile-modules) package should be used in `next.config.js`:

```js next.config.js
const withTM = require('next-transpile-modules')(['@trigger.dev/react']); // pass the modules you would like to see transpiled

module.exports = withTM({});
```

We should do the following:

- [ ] Look into why Next.js is using the ESM module exported by `@trigger.dev/react` instead of the CJS
- [ ] If nothing can be done about the above issue, then we should make sure `@trigger.dev/cli init` correctly handles this (whenever the `@trigger.dev/react` stuff is added to the `init` command (see https://github.com/triggerdotdev/trigger.dev/issues/330)
- [ ] Add to the React Hooks docs (and the Next.js quickstart) info about this

<sub>[TRI-1029](https://linear.app/triggerdotdev/issue/TRI-1029/triggerdevreact-doesnt-work-out-of-the-box-with-nextjs-page-router)</sub>

Chigala:
## Changes
- Added trigger.dev/react package to the Next.js project when `@trigger.dev/cli init` is used. 
-  updated the next.js guide to include the use of react hooks

fixes #330 
temporary fix #331

### DIFF
--- a/docs/documentation/quickstarts/nextjs.mdx
+++ b/docs/documentation/quickstarts/nextjs.mdx
@@ -67,6 +67,182 @@ It will ask you a few questions
 2. Enter your development API key. Enter the key you copied earlier.
 3. Enter a unique ID for your endpoint (you can just use the default by hitting enter)
 
+
+### Add the `TriggerProvider` component
+
+The [TriggerProvider](/sdk/react/triggerprovider) component is a React Context Provider that will make the Trigger API client available to all child components.
+
+Generally you'll want to add this to the root of your app, so that it's available everywhere. However, you can add it lower in the hierarchy but it must be above any of the hooks.
+
+```tsx app/layout.tsx
+export default function RootLayout({ children }: { children: React.ReactNode }) {
+  return (
+    <html lang="en">
+      <body className={inter.className}>
+        <TriggerProvider publicApiKey={process.env.NEXT_PUBLIC_TRIGGER_API_KEY!}>
+          {children}
+        </TriggerProvider>
+      </body>
+    </html>
+  );
+}
+```
+
+### Add hooks to your components
+
+There are three hooks that you can use to show statuses:
+
+- [useEventDetails](/sdk/react/useeventdetails): get the details of a specific event
+- [useRunDetails](/sdk/react/userundetails): get the details of a specific Run
+- [useEventRunDetails](/sdk/react/useeventrundetails): get the details of a Run triggered from a specific event
+
+All of these hooks will automatically refresh your components as the state of your Runs or events change.
+
+#### a. useEventDetails
+
+The `useEventDetails` hook will get the details of a specific event. You can use this to show the status of a specific event.
+
+<Snippet file="how-to-get-event-id.mdx" />
+
+This component will show the details of an event and the overall status of Runs that were triggered by the event:
+
+```tsx
+import { useEventDetails } from "@trigger.dev/react";
+
+export default function EventDetails({ eventId }: { eventId: string }) {
+  const { data, error } = useEventDetails(eventId);
+
+  if (error) {
+    return <div>Error: {error.message}</div>;
+  }
+
+  if (!data) {
+    return <div>Loading...</div>;
+  }
+
+  return (
+    <div>
+      <h1>{data.name}</h1>
+      <p>Runs: {data.runs?.length}</p>
+      <div>
+        {data.runs?.map((run) => (
+          <div key={run.id}>
+            <p>
+              Run {run.id}: {run.status}
+            </p>
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+}
+```
+
+#### b. useRunDetails
+
+The `useRunDetails` hook will get the details of a specific Run. You can use this to show the status of a specific Run.
+
+<Accordion title="How do I get a Run id?">
+  You can call [client.getRuns()](/sdk/triggerclient/instancemethods/getruns) with a Job id to get a
+  list of the most recent Runs for that Job. You can then pass that run id to your frontend to use
+  in the hook.
+</Accordion>
+
+This component will show the details of a Run and the status of each task in the Run:
+
+```tsx
+import { useRunDetails } from "@trigger.dev/react";
+
+export default function RunDetails({ runId }: { runId: string }) {
+  const { data, error } = useRunDetails(runId);
+
+  if (error) {
+    return <div>Error: {error.message}</div>;
+  }
+
+  if (!data) {
+    return <div>Loading...</div>;
+  }
+
+  return (
+    <div>
+      <h1>Run {data.id}</h1>
+      <p>Status: {data.status}</p>
+      <div>
+        {data.tasks?.map((task) => (
+          <div key={task.id}>
+            <p>
+              Task {task.id}: {task.status}
+            </p>
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+}
+```
+
+#### c. useEventRunDetails
+
+The `useEventRunDetails` hook will get the details of a specific Run that was triggered by a specific event. You can use this to show the status of a specific Run.
+
+<Snippet file="how-to-get-event-id.mdx" />
+
+This component will show the details of a Run and the status of each task in the Run:
+
+```tsx
+import { useEventRunDetails } from "@trigger.dev/react";
+
+export default function EventRunDetails({ eventId }: { eventId: string }) {
+  const { data, error } = useEventRunDetails(eventId);
+
+  if (error) {
+    return <div>Error: {error.message}</div>;
+  }
+
+  if (!data) {
+    return <div>Loading...</div>;
+  }
+
+  return (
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/init.ts
```diff
@@ -726,3 +740,16 @@ async function setupEnvironmentVariable(
     logger.success(`✅ Added ${variableName}=${renderer(value)} to ${fileName}`);
   }
 }
+
+async function detectNextJsVersion(projectPath: string): Promise<number> {
+  const packageFile = pathModule.join(projectPath, "package.json");
+  const packageData = await fs.readFile(packageFile, "utf8");
+  const packageJson = JSON.parse(packageData);
```
matt-aitken: You should use the `readPackageJson()` function. It returns a typed `PackageJson` object, or undefined if the files doesn't exist.

packages/cli/src/commands/init.ts
```diff
@@ -47,6 +47,9 @@ export const initCommand = async (options: InitCommandOptions) => {
   // Detect if are are in a Next.js project
   const isNextJsProject = await detectNextJsProject(resolvedPath);
 
+  //Detect Next.Js version
+  const nextJsVersion = await detectNextJsVersion(resolvedPath)
+
```
matt-aitken: I think it'd be better if this code was after the check for if it's a Next.js project, so moved to line 60.

packages/cli/src/commands/init.ts
```diff
@@ -145,15 +149,26 @@ export const initCommand = async (options: InitCommandOptions) => {
...
   const projectUrl = `${options.triggerUrl}/orgs/${authorizedKey.organization.slug}/projects/${authorizedKey.project.slug}`;
 
   logger.success(`✅ Successfully initialized Trigger.dev!`);
 
+
+if (nextJsDir === "pages") {
+  const message = `Detected you are using Next.js Page Router version ${nextJsVersion}, check ${
+    nextJsVersion > 13.1
+      ? "out this article on our docs on how to complete the setup. https://trigger.dev/docs/documentation/quickstarts/nextjs"
+      : "out this article on our docs on how to complete the setup. https://trigger.dev/docs/documentation/quickstarts/nextjs"
```
matt-aitken: Should these two lines say different things?

matt-aitken (changes requested):

packages/cli/src/commands/init.ts
Chigala: it shouldn't actually, I thought there was a way to link the  particular part of the docs with the guide for the changes to be made for the particular Next.js version with mintlify.

packages/cli/src/commands/init.ts
Chigala: I'll do this instead

packages/cli/src/commands/init.ts
matt-aitken: In which case would it better to just do this:

```
const message = `Detected you are using Next.js Page Router version ${nextJsVersion}, check out this article on our docs on how to complete the setup. https://trigger.dev/docs/documentation/quickstarts/nextjs`
```

Chigala: I'll go with this instead, I couldn't find a way to link to a particular part of a doc on mintlify.

docs/documentation/quickstarts/nextjs.mdx
```diff
@@ -67,6 +67,182 @@ It will ask you a few questions
...
+The [TriggerProvider](/sdk/react/triggerprovider) component is a React Context Provider that will make the Trigger API client available to all child components.
+
+Generally you'll want to add this to the root of your app, so that it's available everywhere. However, you can add it lower in the hierarchy but it must be above any of the hooks.
+
+```tsx app/layout.tsx
+export default function RootLayout({ children }: { children: React.ReactNode }) {
+  return (
+    <html lang="en">
+      <body className={inter.className}>
+        <TriggerProvider publicApiKey={process.env.NEXT_PUBLIC_TRIGGER_API_KEY!}>
```
ericallam, line 82: We need to add this environment variable during `init` as well. You'll need to add `pkApiKey` to the `WhoamiResponseSchema` here: https://github.com/triggerdotdev/trigger.dev/blob/302bd02ff16bdba5dce811e93c605d0f832c7920/packages/cli/src/utils/triggerApi.ts#L30

The `/api/v1/whoami` endpoint already returns the `pkApiKey` so no changes will need to be made on the API side.

ericallam (changes requested):

docs/documentation/quickstarts/nextjs.mdx
Chigala: I'll fix this ASAP.

packages/cli/src/commands/init.ts
```diff
@@ -671,6 +684,20 @@ export * from "./examples"
 }
 
 async function setupEnvironmentVariables(path: string, options: ResolvedOptions) {
+  if (options.apiKey && options.apiUrl) {
+    const apiClient = new TriggerApi(options.apiKey, options.apiUrl);
```
ericallam: `whoami` is already being called during init, is it possible to do this without calling it twice?

Chigala: Sorry, I didn't see it was already being called.

packages/cli/src/commands/init.ts
```diff
@@ -145,15 +148,26 @@ export const initCommand = async (options: InitCommandOptions) => {
...
+  authorizedKey: WhoamiResponse,
+  nextJsVersion: number,
+  nextJsDir: "pages" | "app"
+) {
   const projectUrl = `${options.triggerUrl}/orgs/${authorizedKey.organization.slug}/projects/${authorizedKey.project.slug}`;
 
   logger.success(`✅ Successfully initialized Trigger.dev!`);
 
+  if (nextJsDir === "pages") {
+    const message = `Detected you are using Next.js Page Router version ${nextJsVersion}, check out this article on our docs on how to complete the setup. https://trigger.dev/docs/documentation/quickstarts/nextjs`;
```
ericallam, line 166: I'm not sure what this is meant to be for? Is this for doing the transpile module stuff? It doesn't look like we actually have anything in our docs referencing how to finish setting up the Page Router version

packages/cli/src/commands/init.ts
```diff
@@ -670,7 +683,18 @@ export * from "./examples"
   }
 }
 
-async function setupEnvironmentVariables(path: string, options: ResolvedOptions) {
+async function setupEnvironmentVariables(path: string, options: ResolvedOptions, authorizedKey: WhoamiResponse) {
+  if (options.apiKey && options.apiUrl) {
```
ericallam, line 687: Do we need this conditional here?

packages/cli/src/commands/init.ts
```diff
@@ -726,3 +750,21 @@ async function setupEnvironmentVariable(
     logger.success(`✅ Added ${variableName}=${renderer(value)} to ${fileName}`);
   }
 }
+
+async function detectNextJsVersion(projectPath: string): Promise<number> {
```
ericallam, line 754: Why is this function being added here?

ericallam (changes requested):

packages/cli/src/commands/init.ts
Chigala: I updated the docs in the first commit to show how to do that

packages/cli/src/commands/init.ts
Chigala: For the log message that'll have the link to the blog where there's a guide on how to complete @trigger.dev/react set up.

ericallam:
@matt-aitken is currently doing a big refactor on the CLI to account for multi-framework support. Matt could you incorporate some of these changes into the work that you are doing? I've going to close this for now.

Issue #367, opened 2023-08-18
[TRI-1117] Docs: Create a self-hosting guide for AWS ECS using the Stacktape


ericallam:
It should include all the instructions required to deploy Trigger.dev, and any required code. Similar to what was done for [Fly.io](http://Fly.io): [https://trigger.dev/docs/documentation/guides/self-hosting/flyio](https://trigger.dev/docs/documentation/guides/self-hosting/flyio)

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1117](https://linear.app/triggerdotdev/issue/TRI-1117/docs-create-a-self-hosting-guide-for-aws-ecs-using-the-stacktape)</sub>

matt-aitken:
Not relevant anymore, we will have a k8s guide for v4

PR #369, opened 2023-08-19
added docs for aws ecs using stacktape

Anshul1811:
I have tried to add the basic documentation of AWS ECS using Stacktape.Let me know for any changes to make.
#367


matt-aitken:
This is a bit too thin I think. It needs to include instructions for how to create the secret values, like the Fly.io one, as well as the Stacktape instructions actually giving you a deployable app you could run.

PR #381, opened 2023-08-22
Features/delete and disable jobs

samejr:
Draft PR for the new delete/disable feature

PR #382, opened 2023-08-22
Implement the disable job feature backend when indexing endpoints

ericallam:
This PR implements proper job disabling when the `enabled` option is set the `false`. Previously it just did not include the job when indexing.

When a job is set to enabled = false, the following things happen when the endpoint is indexed:

- The JobVersion.status field is set to `"DISABLED"`
- The EventDispatcher.enabled field is set to `false`. This will cause new runs from being created.
- If the job has a ScheduleSource, it will be deactivated and the next scheduled `events.deliverScheduled` task will be dequeued.
- In progress runs will continue to run, and new tasks from those runs will work.
- Resuming runs will not be allowed
- Any in-progress task operations will complete

When a job is then set backed to enabled = true, the following things happen when the endpoint is indexed:

- The JobVersion.status field is set to `"ACTIVE"`
- The EventDispatcher.enabled field is set to `true`. This will re-enable new runs.
- If the job has a ScheduleSource, it will be re-activated and the next scheduled task will be queued
- Failed runs from when the job was disabled will NOT be retried.

One other thing this PR does is it will disable "missing" jobs when an endpoint is indexed. It will basically go through the same "Disable Job" process above when indexing an endpoint if the job exists in the database but NOT in the indexed jobs. It will find the latest version for the environment being indexed and disable that version.

Once this PR is merged into main, main will be merged into the UI PR #381

### DIFF
--- a/apps/webapp/app/services/jobs/disableJob.server.ts
+++ b/apps/webapp/app/services/jobs/disableJob.server.ts
@@ -1,0 +1,103 @@
+import type { JobVersion } from "@trigger.dev/database";
+import type { PrismaClient } from "~/db.server";
+import { prisma } from "~/db.server";
+import { ExtendedEndpoint, findEndpoint } from "~/models/endpoint.server";
+import type { AuthenticatedEnvironment } from "../apiAuth.server";
+import { DisableScheduleSourceService } from "../schedules/disableScheduleSource.server";
+
+export type DisableJobServiceOptions = {
+  slug: string;
+  version: string;
+};
+
+export class DisableJobService {
+  #prismaClient: PrismaClient;
+
+  constructor(prismaClient: PrismaClient = prisma) {
+    this.#prismaClient = prismaClient;
+  }
+
+  public async call(
+    endpointIdOrEndpoint: string | ExtendedEndpoint,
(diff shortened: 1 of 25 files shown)


matt-aitken:
Note for testing that we need to check re-enabling Jobs works properly

ericallam:
@matt-aitken added some extra info in the PR about what happens when jobs are re-enabled

ericallam:
Also added some more details about missing jobs.

apps/webapp/app/services/jobs/disableJob.server.ts
```diff
@@ -0,0 +1,103 @@
...
+      where: {
+        projectId_slug: {
+          projectId: environment.projectId,
+          slug: options.slug,
+        },
+      },
+    });
+
+    if (!job) {
+      return;
```
matt-aitken, line 47: Is it worth adding a log here? Feels like this shouldn't happen

ericallam: It's just defensive programming against thrown errors, but yea I'll add a log

matt-aitken (approved):

Issue #186, opened 2023-07-18
[TRI-689] Login through magic link not working in Safari when running on localhost:3030


ericallam:
Safari version: Version 16.4.1 (17615.1.26.101.10, 17615)

To recreate, make sure you DON'T have the `RESEND_API_KEY` env var set, so the magic link is printed to the terminal instead of actually sent as an email (I can't recreate the problem when the magic link is sent through the email).

After navigating to the magic link (in Safari) you will get redirected back to the magic link page, instead of into the dashboard.

I'm thinking this might have something to do with the the cookie session storage (which you can find here: https://github.com/triggerdotdev/trigger.dev/blob/bca0df2692a9255d79a3858e8bb091cdfa883632/apps/webapp/app/services/sessionStorage.server.ts) but I'm not exactly sure what is causing it

<sub>[TRI-689](https://linear.app/triggerdotdev/issue/TRI-689/login-through-magic-link-not-working-in-safari-when-running-on)</sub>

imrishabh18:
Hey @ericallam,  I have tried couple of times to recreate the issue, but the magic link in my terminal is redirecting it to the Dashboard. Can you share some more steps please or am I missing something?

P.S. - I am a beginner to open source contributions, would love to help and contribute.

SSHari:
@ericallam It looks like this might be the result of the cookie being sent with the [secure flag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie). What's odd is that MDN mentions that this shouldn't be a problem for `localhost`, but it might be a bug with Safari.

I changed the `NODE_ENV` in the `start` script as follows so it runs in development mode:
<img width="1197" alt="image" src="https://github.com/triggerdotdev/trigger.dev/assets/11434879/26941109-e469-4385-900a-de4615ec94ae">

Changing the environment toggles the secure flag here:
<img width="967" alt="image" src="https://github.com/triggerdotdev/trigger.dev/assets/11434879/4e943d82-4a11-410f-8f5f-7f6c5b5141b4">

Running the app this way removes the `Secure` flag from the cookies that are returned from the server. I was able to get things working after making this change.

Here's an example of the before/after of a returned cookie:

### Before
<img width="758" alt="image" src="https://github.com/triggerdotdev/trigger.dev/assets/11434879/afe183e0-4356-49bc-bfd5-fc37cbfad812">

### After
[...]

PR #390, opened 2023-08-24
Fix: Magic link login not working properly when booted from docker

arika0093:
## Problem
When using docker-compose to deploy to any remote environment, the same problem as #186 occurs.
This is caused by the following factors happening at the same time:

1. the `Secure` attribute exception does not work because the domain is not `localhost` (or browsers that respect the `Secure` attribute even on `localhost`, such as safari).
1. `NODE_ENV` is fixed to `production` and `Secure: NODE_ENV === "production"` is always `true`.

It can be reproduced in the local environment by replacing `http://localhost` with `http://127.0.0.1.nip.io`.

This pull request solves factor 2. of the above factors.
Therefore, it is assumed that #186 will be resolved in many cases.

## Detailed cause of occurrence
When booted from a docker image, `entrypoint` calls `/docker/scripts/entrypoint.sh`,
Inside it, `exec dumb-init pnpm run start` is executed.
where `cross-env NODE_ENV=production node --max-old-space-size=8192 . /build/server.js` will be executed.
At this time, since `NODE_ENV` is fixed to `production`, even if environment variables are specified in `env_file` etc., they are overwritten and the above problem occurs.

## Solution
Add `start-from-docker` command to `package.json` and call it when starting from docker container.
The command is exactly the same as `start` except that `NODE_ENV` is not overwritten.
This fix will ensure that the `NODE_ENV` specified in the `.env` file is respected and the above bug is repaired.

/claim #186

### DIFF
--- a/apps/webapp/package.json
+++ b/apps/webapp/package.json
@@ -12,7 +12,7 @@
     "format": "prettier --write .",
     "lint": "eslint --cache --cache-location ./node_modules/.cache/eslint .",
     "start": "cross-env NODE_ENV=production node --max-old-space-size=8192 ./build/server.js",
-    "start-from-docker": "cross-env node --max-old-space-size=8192 ./build/server.js",
+    "start:local": "cross-env node --max-old-space-size=8192 ./build/server.js",
     "typecheck": "tsc --noEmit",
     "db:seed": "node prisma/seed.js",
     "db:seed:local": "ts-node prisma/seed.ts",
--- a/apps/webapp/package.json
+++ b/apps/webapp/package.json
@@ -12,6 +12,7 @@
     "format": "prettier --write .",
     "lint": "eslint --cache --cache-location ./node_modules/.cache/eslint .",
     "start": "cross-env NODE_ENV=production node --max-old-space-size=8192 ./build/server.js",
+    "start_docker": "cross-env node --max-old-space-size=8192 ./build/server.js",
     "typecheck": "tsc --noEmit",
     "db:seed": "node prisma/seed.js",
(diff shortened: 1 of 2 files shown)


apps/webapp/package.json
```diff
@@ -12,6 +12,7 @@
     "format": "prettier --write .",
     "lint": "eslint --cache --cache-location ./node_modules/.cache/eslint .",
     "start": "cross-env NODE_ENV=production node --max-old-space-size=8192 ./build/server.js",
+    "start-from-docker": "cross-env node --max-old-space-size=8192 ./build/server.js",
```
wesleymatosdev: could it be `start:local`, maybe?
```suggestion
    "start:local": "cross-env node --max-old-space-size=8192 ./build/server.js",
```

arika0093: The reason for avoiding `start:local` is to avoid the misunderstanding that they are started together by the `run-s start:*` command, like `dev` and `dev:*`.
However, the name really could be anything, so I will modify it to `start:local`.

wesleymatosdev: If you want, you can use something like `docker:start` or something like that. I commented because except for `build-storybook`, we don't have other script commands that use `-`. Feel free to use another name like `docker:start` or something like that.

Thanks for accepting my review 😄

arika0093:
The command name was changed to `start:local`. 
Please let me know if you have additional suggestions or comments.

ologbonowiwi (approved):

ericallam:
Thank you so much for the contribution! I will review this tomorrow 👍

xAmast:
It has been two weeks. still planning to merge this PR?

Issue #396, opened 2023-08-25
[TRI-1152] chore: upgrade webapp to use graphile-worker `0.14.0-rc.0`


ericallam:
We're currently on 0.13.0 but `0.14.0` comes with automatic queue management, so it will be easier to do #392 along with other things:

* New "batch jobs" feature for merging payloads with a job_key (see README)
* Significantly improved 'large jobs table' performance (e.g. when a large queue is locked, or there's a lot of jobs queued for task identifiers your worker instance doesn't support, or a lot of failed jobs). Around 20x improvement in this 'worst case' performance for real user workloads.
* Added new (experimental) much faster add_jobs batch API.
* Fix error handling of cron issues in 'run' method.
* CronItem.match can now accept either a pattern string or a matcher function
* Jobs that were locked more than 4 hours will be reattempted as before, however they are slightly de-prioritised by virtue of having their run_at updated, giving interim jobs a chance to be executed (and lessening the impact of queue stalling through hanging tasks).

The full release notes are here:

[https://github.com/graphile/worker/blob/main/RELEASE_NOTES.md#v0140](https://github.com/graphile/worker/blob/main/RELEASE_NOTES.md#v0140)

As mentioned in the release notes, this is a breaking change, meaning workers running `0.13` cannot run against the `0.14` schema. So we'll need to make this clear in any release notes for this feature.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1152](https://linear.app/triggerdotdev/issue/TRI-1152/chore-upgrade-webapp-to-use-graphile-worker-0140-rc0)</sub>

nicktrn:
Just ticking some boxes to know what needs actioning.

### Breaking changes

- [x] Bump minimum Node version to 14
- [x] Bump minimum PG version to 12
- [x] `priority`, `attempts` and `max_attempts` are now `smallint`, i.e. values have to be between `-32768` and `32768`
  * Using current max of `100`
- [ ] `CronItem.pattern` has been renamed to `CronItem.match`
  * TODO: Rename this in recurring tasks and `ZodWorker` 
- [x] Database error codes have been removed because we've moved to CHECK constraints
  * We don't rely on specific error codes

### Internal changes

- [x] The 'jobs' table no longer has `queue_name` and `task_identifier` columns
  * We don't rely on specific column names (see below for raw query returns)
- [x] Internal SQL functions (`get_job`, `fail_job`, `complete_job`) have been moved to JS
[...]

Issue #348, opened 2023-08-17
[TRI-1081] Feature idea: Batch Event Trigger


ericallam:
It would be useful to have a "batch" event trigger, where you could set the frequency that the batched events would be flushed:

```ts
batchEventTrigger({ name: "batched.event", schema: z.object({ id: z.string() }), flushEvery: 60 })
```

Then whenever you sent an event with an ID, instead of immediately triggering runs with the payload it will get added to a batch. If more events with the same ID are sent within the 60 seconds of the first one, those payloads will get appended.

Then, when the run is triggered the payload in the run function becomes an array of all the event payloads.

<sub>[TRI-1081](https://linear.app/triggerdotdev/issue/TRI-1081/feature-idea-batch-event-trigger)</sub>

ericallam:
Could also maybe do it with a cron expression

ericallam:
The more I think about it the more I feel like maybe this can be done without a new trigger type and just using the `eventTrigger` and `sendEvent`. 

Ideally we could just add options to `sendEvent` like so:

```ts
await client.sendEvent({ id: "event123", name: "new.message", payload: { message: "Hey there!" }}, { batchKey: "user123", deliverAfter: 60 })
await client.sendEvent({ id: "event456", name: "new.message", payload: { message: "It's me!" }}, { batchKey: "user123", deliverAfter: 60 })
await client.sendEvent({ id: "event789", name: "new.message", payload: { message: "Where are you?" }}, { batchKey: "user123", deliverAfter: 60 })
```

If all three of those events were sent within 60 seconds of the first one, then they would all trigger a single run of an eventTrigger with the batch option set to true.

```ts
client.defineJob({
  id: "batched-event-example",
  name: "Batched Event Example",
  version: "1.0.0",
  trigger: eventTrigger({
    name: "new.message",
    schema: z.object({ message: z.string() }),
    batch: true,
  }),
  run: async (payload, io, ctx) => {
    // payload is typed as Array<{ message: string }>
  },
});
```

One other thing we'd need to consider is the `TriggerContext["event"]` type might need to change because a single run can now be caused by more than 1 event:

```ts
export interface TriggerContext {
  /** Job metadata */
  job: { id: string; version: string };
  /** Environment metadata */
  environment: { slug: string; id: string; type: RuntimeEnvironmentType };
  /** Organization metadata */
  organization: { slug: string; id: string; title: string };
  /** Run metadata */
[...]

PR #399, opened 2023-08-25
Airtable integration with webhook and integration changes

matt-aitken:
## What does this include?
- Airtable integration with Tasks support
- Creating authenticated integrations (ioWithIntegrations) now clones them with the auth details. This allows us to have much more flexible SDKs where Tasks don't need to just live in the root (e.g.`io.airtable.base(payload.baseId).table<LaunchGoalsAndOkRs>.getRecords`)
- TriggerSources aren't limited to a single TriggerSourceEvent, instead they TriggerSourceOptions which allows greater flexibility
- Integrations can now do API calls when they receive webhook payloads. This is required for Airtable

## To-do
- [ ] Don't export the Airtable webhooks for now
- [ ] Fix all the other integrations because of the changes with cloneForRun

## Other PRs required to have Airtable webhooks
1. Upgrade Graphile worker to 0.14.0 (https://github.com/triggerdotdev/trigger.dev/issues/396)
2. Add batch event support (https://github.com/triggerdotdev/trigger.dev/issues/348)
3. Add Airtable webhooks (using batching)


ericallam:
## Dynamic Triggers

Adding some documentation here to help when you are making changes to the Dynamic Trigger code for this PR. 

There are two flows to registering new "Trigger Sources" for a Dynamic Trigger: 

- API initiated 
- TriggerClient initiated

### API Initiated

Dynamic trigger registrations can be initiated via an API call, but this requires an extra round trip to the Client App to get registration information, which you can see in the diagram below:

[mermaid-diagram-2023-08-28-165623](https://mermaid.live/edit#pako:eNp1ksGO1DAMhl_FivY4S8U1h5VGTFlFgmXV9rSqhLyN6QbapDjpoGU0744zbUFooIcq-v3Zv534pLpgSWkV6ftMvqODw55xbD3INyEn17kJfYIDYITDq8fRddCw63via2qfqXeDo3yepmugycCa_sbS8ZowmTDeJYeD-0krWxMfXUfXeJXxinoXE_EGh5k7-m9KuTj0FFNN3pZH6Xah9rd3d42Gx091AwVOrji-LW4EmYLzqR7mvkiLQyxunC34YsuYXPBxqdDcSgmjocNhWBQjwl6DeTCN2X8wT-XnpjL392W1WUr8oIHXEciuQ7wP_IjyFPFPa-ZfZSQzTuJPm52A1d8NiFJqyLNGoDwtpAA8-9-mEC83BjIlsccBvobnJXn7q50aiUd0VpbllNVWpRcaqVVajhb5W6t2i77tUg6d1gIqvoQf9Rp4mMdnucMcTzxfGj-3_iweOKdQv_pO6RzYqXmymLalVPoLDlFUsi4F_rhs7mWBd0pe9imEjTn_An728nA)

1. Make a POST request with a JSON body. The `$id` param maps to the `DynamicTrigger.id`. Here's an exmaple for a GitHub dynamic trigger. `accountId` is optional, but this maps to an external account identifier for Trigger.dev Connect.

```json
{
  "id": "abc1235",
  "params": {
    "owner": "triggerdotdev",
    "repo": "empty-eric"
  },
  "accountId": "eric1234",
  "metadata": {
    "full_repo": "triggerdotdev/empty-eric"
  }
}
```

2. The `InitializeTriggerService` is called with the value of the request body
3. A request to the TriggerClient is made with the action INITIALIZE_TRIGGER
4. The TriggerClient looks up the `DynamicTrigger` instance and calls `registeredTriggerForParams`.
5. The result is passed back as the response to the `INITIALIZE_TRIGGER` request
6. The `InitializeTriggerService` uses that response to call `RegisterTriggerSourceService`
7. Uses the result of `RegisterTriggerSourceService` to Ingest a new event that will trigger a run of the DynamicTrigger internal registration job (see https://github.com/triggerdotdev/trigger.dev/blob/4685b630d90f6b63db691759c15f2fd6aa433e80/packages/trigger-sdk/src/triggerClient.ts/#L429)

## TriggerClient Initiated

The second way a DynamicTrigger source can be registered is via the TriggerClient (see https://github.com/triggerdotdev/trigger.dev/blob/4685b630d90f6b63db691759c15f2fd6aa433e80/packages/trigger-sdk/src/io.ts)

I actually think we maybe should just update registerTrigger to work exactly like the API initiated version, except we can skip the roundtrip `INITIALIZE_TRIGGER` request to get all the necessary data for the call to `RegisterTriggerSourceService` since that will be sent in the request. I propose we update the TriggerClient initiated flow to this:

[mermaid-diagram-2023-08-28-165843](https://mermaid.live/edit#pako:eNplkj1vgzAQhv_KycpIiroyZCJDh0ZVoEvF4thX4gZsej5TRVH-e20-qkrxgKz3fe6L800op1EUwuN3QKuwNLIl2TcW4hkksVFmkJahBOmhvFrZGwU1mbZFeqTqRC3uk8bxkTgm4oit8Yy0kJULpLBCGo3Cx5B9CnmxLXqu0Or9iJZnqtzudnUBb-815HIw-ficbyIxOGO56kKb81zA5xujc5qqkmTjbFQueJ2zHBwjuBEJyixmW4Zcp6ClV5jhehtrHgtQsuv-KfsCUm8eMHUH7ICChb9YP00IsS0kKzv4cqc5eP2KTPRIvTQ6LuOW1EbwGXtsRBGvWtKlEdmsr7tK1m1JIPzZ_VSLcQj9KQ6dfKYw_dF7Y--xhgzsqqtVokhGJsKgJa9LF8Wn7HxUURt29Dq_jOmBZCJu4sO5lbn_AkONwL0)

matt-aitken:
Duplicate runs are happening with GitHub

CleanShot 2023-09-01 at 11 07 21@2x

apps/webapp/app/routes/api.v2.$endpointSlug.triggers.$id.registrations.$key.ts
```diff
@@ -0,0 +1,106 @@
...
+      authenticatedEnv,
+      {
+        id: registration.id,
+        name: REGISTER_SOURCE_EVENT_V2,
+        source: "trigger.dev",
+        payload,
+      }
+      //todo accountId?
+      // {accountId: body.data.accountId}
+    );
```
matt-aitken, line 91: @ericallam I couldn't figure out where the accountId comes from for this

packages/trigger-sdk/src/triggers/dynamic.ts
```diff
@@ -81,14 +81,18 @@ export class DynamicTrigger<
...
       source: {
+        version: "2",
         key,
         channel: this.source.channel,
         params,
-        events: typeof this.event.name === "string" ? [this.event.name] : this.event.name,
+        //todo add other options here
+        options: {
+          event: typeof this.event.name === "string" ? [this.event.name] : this.event.name,
+        },
```
matt-aitken, line 91: @ericallam currently this just has the event name, I couldn't figure out how we would support other options here

matt-aitken (commented):
@ericallam added a couple of questions in here

apps/webapp/app/routes/api.v2.$endpointSlug.triggers.$id.registrations.$key.ts
ericallam: I think you'd need to either pass it as an option on `IO.registerTrigger` or use the `IO._context` to get at it? I wouldn't worry about it for now though.

packages/trigger-sdk/src/triggers/dynamic.ts
ericallam: I think that's fine for now, we can sort out the other options later if needed.

PR #402, opened 2023-08-26
feat: implemented io.waitUntil and io.waitForNextEvent

### LINKED ISSUE #315: [TRI-1008] Add `io.waitUntil` function to allow a job run to wait for a dynamic condition
This kind of thing would be really useful in jobs:

```ts
await io.waitUntil("has used invite", {
  condition: async () => {
    const user = await prisma.user.findUniqueOrThrow({
      where: {
        id: payload.userId,
      },
    });

    return user.invitationCodeId !== null;
  },
});
```

This would basically retry the condition function in an interval, and only continue once the condition function returned `true`.

The strategy for how frequent to call the condition function could be configurable:

- Set an interval frequency (minimum would be 60 seconds, max would be 1 day). Would also need some kind of timeout thing like "exit if the condition doesn't return true within 30 days"
- Set an exponential backoff strategy (similar to our task retries) with a maximum of 25 retries

This can be achieved using only subtasks and `io.wait`, something like:

```ts
type WaitUntilOptions = {
  condition: () => Promise<boolean>;
  checkInterval?: number;
};

function waitUntil(
  key: string,
  io: IO,
  options: WaitUntilOptions
): Promise<void> {
  return io.runTask(
    key,
    {
      name: "Wait until",
    },
    async (task) => {
      const result = await options.condition();
      if (result) return;

      //if more than 14 days has passed then throw
      const startedAt = task.startedAt;
      if (!startedAt) return;
      const now = new Date();
      const diff = now.getTime() - startedAt.getTime();
      const days = diff / (1000 * 60 * 60 * 24);
      if (days > 14) {
        throw new Error(
          `Waited for 14 days for condition to be true but it never was.`
        );
      }

      await io.wait(
        `${key}:${new Date().toISOString()}`,
        options.checkInterval ? Math.max(options.checkInterval, 30) : 60 * 60
      );
    }
  );
}
```

But i'd prefer this not to produce a bunch of tasks that are only there for the interval. 

<sub>[TRI-1008](https://linear.app/triggerdotdev/issue/TRI-1008/add-iowaituntil-function-to-allow-a-job-run-to-wait-for-a-dynamic)</sub>

Chigala:
## Feature Description
`io.waitUntil`
#### usage: 
  ```js
    await io.waitUntil("form-submitted", {
    condition: async () => {
      // Simulate waiting for a form submission status
      const isFormSubmitted = await checkFormSubmissionStatus();
      return isFormSubmitted;
    },
    checkInterval: 5000, // Check every 5 seconds
  });

```
This would wait till the condition becomes truthy. 

`io.waitUntilNextEvent`
#### usage: 
```js
await io.waitForNextEvent("account-verified", {
    name: "user-registration",
    interval: 60, // Check every 60 seconds
    filter: {
      isVerified: async () => {
        // Simulate checking the user's verification status
        const isVerified = await checkUserVerificationStatus();
        return isVerified;
      },
      hasValidEmail: async () => {
        // Simulate checking if the user has a valid email
        const hasValidEmail = await checkUserEmailValidity();
        return hasValidEmail;
      },
      meetsAgeRequirement: async () => {
        // Simulate checking if the user meets the age requirement
        const userAge = await getUserAge(); // Assume this retrieves the user's age
        return userAge >= 18; // Assuming the age requirement is 18 or older
      },
      gender: payload.gender === "male"
    },
```
This would wait till all the filters are truthy. It'll keep checking every 60secs

## Related Issues

Closes #315 

## Screenshots (if applicable)

here's a screenshot of my dashboard showing this works

Screenshot 2023-08-26 at 9 53 05 AM

## Additional Notes

@ericallam  Please review this and tell me what you think so I can iterate further.

### DIFF
--- a/packages/trigger-sdk/src/io.ts
+++ b/packages/trigger-sdk/src/io.ts
@@ -47,6 +47,11 @@ export type IOOptions = {
   cachedTasks?: Array<CachedTask>;
 };
 
+type WaitUntilOptions = {
+  condition: () => Promise<boolean>;
+  checkInterval?: number;
+};
+
 export class IO {
   private _id: string;
   private _apiClient: ApiClient;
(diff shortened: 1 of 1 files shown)


ericallam:
Thanks for taking an initial crack at this, but I think we will need to rethink how we approach the `waitUntil` function, because this could cause a massive amount of additional requests to the client, the vast majority of which would be wasted. I'll write this up properly in a Github discussion but one idea would be to do a "poll" to the client at most every 60 seconds, all the polling "work" that had been built up in the last interval would be done at once.  Maybe we could call it something like "Batch Polling". This is going to be needed by things like Polling triggers. So I'm going to close this for now, thanks again for the attempt!

PR #404, opened 2023-08-26
feat: add proxy support to the CLI

arika0093:
This feature is expected to allow the CLI and SDK to work in environments where proxy is enabled.
The standalone `node-fetch` library does not support proxies, so I have added an external library `proxy-agent`.

Without proxy support, errors may occur at the following locations during initiation. 
The addition of the feature fixes this bug.
```ts
// /packages/cli/src/utils/addDependencies.ts
const response = await fetch(`https://registry.npmjs.org/${packageName}`);
```


arika0093:
I removed the SDK changes and made the changes to the CLI only.
Please check back when you have time.

PR #410, opened 2023-08-27
Enhancement/lower max attempts for enqueued task for development environment

### LINKED ISSUE #393: Lower maxAttempts for specific tasks when those tasks call to development endpoints
For the following worker tasks: 

- performRunExecutionV2
- deliverHttpSourceRequest
- indexEndpoint

We currently are setting the maxAttempts statically to:

- performRunExecutionV2 to 12
- deliverHttpSourceRequest to 14
- indexEndpoint to 7

While those values make sense when they are hitting production endpoints, as we want to retry just in case production endpoints are temporary down for whatever reason, they really don't make sense for development endpoints, since development endpoints are usually either running or not, so maxAttempts should be set to 1.

This would greatly reduce the graphile worker load for these tasks.

So we should set the maxAttempts for these at `enqueue` time. 

### performRunExecutionV2

This is enqueued from a single location:

https://github.com/triggerdotdev/trigger.dev/blob/1fd1d267800647af8139851bc22876cbf4643d5f/apps/webapp/app/models/jobRunExecution.server.ts#L32

We can add an option to 

```ts
export type EnqueueRunExecutionV2Options = {
  runAt?: Date;
  resumeTaskId?: string;
  isRetry?: boolean;
};
```

TODO the rest of this issue

arizworld:
closes #393 
/claim #393 

In this PR I have done the below changes :

1. added a new optional boolean flag skipRetrying for the options of enqueueRunExecutionV2 to restrict maxAttempts to 1 if the flag is true.
2.  updated the calls to enqueueRunExecutionV2 so that it skips retrying for development enviroment.
3.  updated the workerQueue.enqueue call for deliverHttpSourceRequest so that it just attempts once for development enviroment.
4.  updated the workerQueue.enqueue call for indexEndpoint so that it just attempts once for development enviroment.

Note: I have
n't updated the call from this function as it only fetches the endpoints which are already having a enviorment type of production 

Screenshot 08-27-2023 19 05 31

Please let me know if any further changes is required, Thank You. It was fun contributing this awesome codebase

### DIFF
--- a/apps/webapp/app/models/jobRunExecution.server.ts
+++ b/apps/webapp/app/models/jobRunExecution.server.ts
@@ -27,6 +27,7 @@ export type EnqueueRunExecutionV2Options = {
   runAt?: Date;
   resumeTaskId?: string;
   isRetry?: boolean;
+  skipRetrying?: boolean
 };
 
 export async function enqueueRunExecutionV2(
@@ -47,6 +48,7 @@ export async function enqueueRunExecutionV2(
       tx,
       runAt: options.runAt,
       jobKey: `job_run:${run.id}`,
+      maxAttempts: options.skipRetrying ? 1 : undefined
     }
   );
 }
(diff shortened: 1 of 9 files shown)


arizworld:
There is some wrong code

PR #432, opened 2023-09-04
Add `@trigger.dev/testing` package

### LINKED ISSUE #379: [TRI-1127] Improvement: Add a `@trigger.dev/testing` package that allows for unit testing job runs
Currently, we don't have a clean way of unit testing jobs, it would be nice to have a `@trigger.dev/testing` package that could be used to do test job runs with mocked `io` and make assertions:

```ts
const jobToTest = client.defineJob({
  id: "stripe-example-1",
  name: "Stripe Example 1",
  version: "0.1.0",
  trigger: eventTrigger({
    name: "stripe.example",
    schema: z.object({
      customerId: z.string(),
      source: z.string(),
    }),
  }),
  integrations: {
    stripe,
  },
    await io.stripe.createCharge("create-charge", {
      amount: 100,
      currency: "usd",
      source: payload.source,
      customer: payload.customerId,
    });
  },
});

import { testJob } from "@trigger.dev/testing";

// Inside a test file
const testRun = testJob(jobToTest, {
  payload: {
    customerId: "cus_123",
    source: "src_123",
  },
});

expect(testRun).toHaveSucceeded();
// Test the stripe charge was called with the correct params
expect(testRun.io.stripe.createCharge).toHaveBeenCalledWith("create-charge", {
  amount: 100,
  currency: "usd",
  source: "src_123",
});
```

<sub>[TRI-1127](https://linear.app/triggerdotdev/issue/TRI-1127/improvement-add-a-triggerdevtesting-package-that-allows-for-unit)</sub>

nicktrn:
As ordered.

```env
# build
pnpm build --filter @trigger.dev/testing

# install
pnpm install

# test
pnpm test --filter @examples/unit-testing
```

Leaving this in draft as I won't be available for a few days. Should be ready though, so feel free to test and review!

Fixes #379
/claim #379

### DIFF
--- a/packages/testing/src/index.ts
+++ b/packages/testing/src/index.ts
@@ -1,0 +1,197 @@
+import { RunJobBody, SendEvent, SendEventBodySchema, SendEventOptions } from "@trigger.dev/core";
+import type {
+  EventSpecification,
+  IOWithIntegrations,
+  IntegrationClient,
+  Job,
+  Trigger,
+  TriggerClient,
+  TriggerEventType,
+  TriggerIntegration,
+} from "@trigger.dev/sdk";
+import { vi as vitestVi } from "vitest";
+import { entries } from "./utils";
+
+declare module "vitest" {
+  interface Assertion<T = any> extends CustomMatchers<T> {}
+  interface AsymmetricMatchersContaining extends CustomMatchers {}
+}
+
+interface CustomMatchers<R = unknown> {
+  toHaveSucceeded(): R;
+}
+
+// @ts-ignore
+declare module "expect" {
+  interface AsymmetricMatchers {
+    toHaveSucceeded(): void;
+  }
+  interface Matchers<R> {
+    toHaveSucceeded(): R;
+  }
+}
+
+let promiseCounter = 0;
+
+export const toHaveSucceeded = (received: any) => {
+  if (received?.status === "SUCCESS") {
+    return {
+      message: () => "run succeeded",
+      pass: true,
+    };
+  } else {
+    return {
+      message: () =>
+        received instanceof Promise
+          ? promiseCounter++ < 1
+            ? "you passed a promise, please don't do that again"
+            : "you did it again.."
+          : "run failed",
+      pass: false,
+    };
+  }
+};
+
+type TriggerAction =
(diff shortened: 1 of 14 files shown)


packages/testing/src/index.ts
```diff
@@ -0,0 +1,197 @@
...
+    let ioSpy: IOWithIntegrations<TIntegrations> | undefined;
+
+    const taskMockImplementation = async (key: string) => {
+      if (!opts.tasks || typeof key !== "string") return;
+      return opts.tasks[key];
+    };
+
+    job.options.run = (payload, io, ctx) => {
+      if (integrations) {
+        for (const [integrationName, integration] of entries(integrations)) {
```
matt-aitken: Because we slightly changed how integrations work a few days ago, this unfortunately doesn't work anymore. I took a quick look to see if I could fix it easily but I couldn't. 

Tasks are now just methods inside integrations, and can even be nested inside other objects.

Given that they're all powered by io.runTask, we could mock that function. This doesn't work but is the rough idea:

```ts
job.options.run = (payload, io, ctx) => {
      vi.spyOn(io, "runTask").mockImplementation((key, callback) => {
        //do some mocked thing instead
      });

      if (integrations) {
        for (const [integrationName, integration] of entries(integrations)) {
          const readyIntegration = integration.cloneForRun(io, integrationName as string);

          io[integrationName] = readyIntegration;
        }
      }
      return run(payload, io, ctx);
    };
```

matt-aitken (changes requested):
Unfortunately we made some changes to how integrations work – they now don't have explicit Tasks. Tasks can be structured however an author wants, including nested.

I've made a suggestion about mocking io.runTask

nicktrn:
Well, I guess that was bad timing. @matt-aitken thanks for updating the dummy integration and pointers re the new format. Wasn't a trivial fix indeed.

Minor API change: Function spies are now exposed and indexed via their key on `jobTestResult.tasks`.

I've also added tests to the package itself, so it doesn't have to be constantly rebuilt for testing during dev.

matt-aitken (approved):
This is a great PR @nicktrn. Thanks for adapting it so quickly after we changed how integration tasks work.

nicktrn:
Thank you kindly! :tophat:

Issue #423, opened 2023-08-28
[TRI-1171] Add to docs: note about https://github.com/nextauthjs/next-auth/issues/7660


ericallam:
Users on Next.js 13.4.4 will run into the error linked in the title, we should add this to the docs in the Next.js platform guide

<sub>[TRI-1171](https://linear.app/triggerdotdev/issue/TRI-1171/add-to-docs-note-about-httpsgithubcomnextauthjsnext-authissues7660)</sub>

hrutik7:
hey @ericallam is this looks good to you ?
![image](https://github.com/triggerdotdev/trigger.dev/assets/62372737/94d60f81-2fc3-49a9-95ed-93ec61685c63)
Or I need to change a bit ?

PR #439, opened 2023-09-06
next 13.4.4 note added

hrutik7:
#423


matt-aitken:
Hi @hrutik7, the page @ericallam was referring to was this one: https://github.com/triggerdotdev/trigger.dev/blob/main/docs/documentation/guides/platforms/nextjs.mdx. Not the quickstart.

I think it would be good to create a new section called "Known issues" with a bulleted list. This would be the only bullet for now and would provide a bit more information than you have on the page.

Ideally the bullet would be a description of the issue (so people can figure out if the'yre experiencing it) with a link to the issue.

Issue #449, opened 2023-09-08
[TRI-1252] Framework: support for Nest.js


samejr:
# Build support for Nest.js into Trigger.dev

**_Please make sure to use Typescript for all code_**

### Task 1 - Nest.js Adaptor
- [ ] Create a new folder in `packages` called `nestjs`
- [ ] Create the adaptor for Nest.js. Use [this Next.js adaptor](https://github.com/triggerdotdev/trigger.dev/blob/main/packages/nextjs/src/index.ts) code as reference when creating yours. Include all setup configurations to fully support the framework, e.g. Next.js requires adaptor support for both App and Pages routers.
- [ ] Create a basic Nest.js project that uses the adaptor in `examples/nestjs` folder. Include a basic Job in Trigger.dev, e.g. a simple Job that just does a delay inside it. Screen grab the full Run dashboard page showing it successfully worked. Here are examples of how to do cross-project references in [package.json](https://github.com/triggerdotdev/trigger.dev/blob/aae143d96e1cd3c2b2b2ccb5120f75651a4261d0/examples/nextjs-example/package.json#L14) and [tsconfig.json](https://github.com/triggerdotdev/trigger.dev/blob/aae143d96e1cd3c2b2b2ccb5120f75651a4261d0/examples/nextjs-example/tsconfig.json#L31).
- [ ] Provide a screen grab of your terminal showing the Adaptor is working

### Task 2 – Add a README.md file
- [ ] Create a new README.md file inside the `packages/nestjs` folder and copy in the contents from `packages/nextjs/README.md`. Adjust the wording to reference Nest.js where needed, including updating the the URLs.

### Task 3 – Write the documentation
- [ ] Locate the `manual-setup-nestjs.mdx` folder in `/docs/_snippets/`. Use this file to write your documentation, overwriting any contents of that file. Follow the structure of [this Next.js example](https://trigger.dev/docs/documentation/guides/manual/nextjs) for reference when writing yours.
- [ ] Update the table at the bottom of the `client-adaptors.mdx` page with the framework you’ve worked on.

### Task 4 – Update the webapp onboarding
[...]

PR #461, opened 2023-09-09
feat: added nestjs package adapter

H4ad:
/claim #449

This PR adds a basic integration with NestJS by exporting `TriggerClient` through `@InjectTriggerDevClient()` decorator.

## Images

image

image

image

image

## Suggestions for the next iterations

For now, I focus only on adding the initial structure and documentation for this integration.

NestJS can be very extensive, and one of the APIs we can add to integrate Trigger.dev can be something like this:

```ts
export interface TriggerDevJobConsumer<TTrigger extends Trigger<EventSpecification<any>>, TIntegrations extends Record<string, TriggerIntegration> {
  getTrigger(): TTrigger;
  run: (payload: TriggerEventType<TTrigger>, io: IOWithIntegrations<TIntegrations>, context: TriggerContext) => Promise<any>;
}

@Job({
  id: "test-job",
  name: "Test Job One",
  version: "0.0.1",
})
export class MyCustomJobConsumer {
  public getTrigger() {
    return eventTrigger({
      name: "test.event",
    });
  }

  public async run(payload, io, ctx) {
    await io.logger.info("Hello world!", { payload });

    return {
      message: "Hello world!",
    };
  }
}
```

Then, you just need to import `MyCustomJobConsumer` in any module and it will automatically register the Job.

But I suggest first creating an issue describing an `Interface` for this consumer and then trying to create this feature.


matt-aitken:
This is a great PR.

We made some major changes on the platform side, so I had to modify your branch a bit to get it merged. But that's now done and merged: #574. I'll tip you using Algora now on the issue. 

Sorry it took so long, the platform changes made all of the framework PRs a nightmare… 🤦 

Thanks a lot 🙏

PR #463, opened 2023-09-10
Created Pull Request, Bug Report, labeler Template

visheshrwl:
Create bug_report.yml, feature_request.yml, config.yml, labeler.yml, and Create pull_request_template.md


ericallam:
This is a great idea! These look to be pulled from create-t3-app so will need to be customised before merging

Issue #159, opened 2023-07-13
[TRI-321] Better handle errors bubbling up in subtasks


ericallam:
Currently when viewing a failed task because a subtasks threw an error is not a great experience. We need to improve that

When a subtask fails (and all retries are done) the error doesn't bubble up to the parent task. Instead you get a very generic: "Error: Task errored at IO.runTask".

Instead, a failed subtask should bubble that error up to parent tasks (so the subtask error is duplicated onto the parent).

You can easily reproduce this by creating a task with subtask inside that throws a specific error.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-321](https://linear.app/triggerdotdev/issue/TRI-321/better-handle-errors-bubbling-up-in-subtasks)</sub>

SSHari:
@ericallam would it be possible to get more information about what the bad experience is and what the expected experience should be?

rishi-raj-jain:
@ericallam Can I work on this one? Can you elaborate the issue?

rishi-raj-jain:
@ericallam would it be possible to get more information about what the bad experience is and what the expected experience should be?

rishi-raj-jain:
@matt-aitken @ericallam

Would it be possible to get more information about what the bad experience is and what the expected experience should be?

matt-aitken:
I've added more information to the original issue

matt-aitken:
@RamK777-stack would you like to claim the Algora bounty?

RamK777-stack:
Yes @matt-aitken

RamK777-stack:
claim #159

matt-aitken:
This was fixed

PR #468, opened 2023-09-11
fix: Added retry config and showed error from subtask in web

### LINKED ISSUE #422: Ensure `@trigger.dev/openai` background tasks have retry configuration
These backgroundFetch tasks need to have their retry configuration set:

- [`backgroundCreateCompletion`](https://github.com/triggerdotdev/trigger.dev/blob/4685b630d90f6b63db691759c15f2fd6aa433e80/integrations/openai/src/tasks.ts)
- [`backgroundCreateChatCompletion`](https://github.com/triggerdotdev/trigger.dev/blob/4685b630d90f6b63db691759c15f2fd6aa433e80/integrations/openai/src/tasks.ts)

View the docs on `backgroundFetch` retry here: https://trigger.dev/docs/sdk/io/backgroundfetch

And the OpenAI error codes here: https://platform.openai.com/docs/guides/error-codes

RamK777-stack:
fixes #422 and #159

Currently we are not showing the exact error on trigger.dev web If openAPI request failed. So I am trying to show the exact error and reason for the task failed.

My request failed because of insufficient quota, status code is 429. Unfortunately, both too many requests and insufficient quota are having same status code 429. This is with the background functions.

Previously it was 

image

Now I have fixed this like this

image (1)

Note: I didn't test positive flow. because currently, I don't have an active openAI key. I consumed all of my free quota


packages/trigger-sdk/src/io.ts
```diff
@@ -527,7 +527,7 @@ export class IO {
         task,
       });
 
-      throw new Error(task.error ?? "Task errored");
+      throw new Error(task.error ?? JSON.stringify(task?.output));
```
matt-aitken: I think we still need the fallback "Task errored", if there's no task.output.

```suggestion
      throw new Error(task.error ?? task?.output ? JSON.stringify(task.output) : "Task errored");
```

matt-aitken (approved):
This is great! I made it so it falls back to "Task errored" if the task has no output

Issue #469, opened 2023-09-12
[TRI-1273] Create a function call example for OpenAI


ericallam:
It would be great to create a function call example for OpenAI here: https://github.com/triggerdotdev/trigger.dev/blob/main/examples/job-catalog/src/openai.ts

Use these docs for how to do function calling: https://gpt.pomb.us/

<sub>[TRI-1273](https://linear.app/triggerdotdev/issue/TRI-1273/create-a-function-call-example-for-openai)</sub>

RamK777-stack:
@Chigala Can I work on this !! I am very interested to work on this one.

Chigala:
That's ohk @RamK777-stack  you can have it.

RamK777-stack:
Thanks @Chigala

RamK777-stack:
attempt #469

prabaljainn:
/attempt #469, Can I work on this? @ericallam

prabaljainn:
> Update: I already completed the development, about to raise the PR

it's been 6 hours. Sir

prabaljainn:
@ericallam, Let me know if you need any changes or something. I will do it ASAP.

asr-alurisanthoshreddy:
Is this issue need any further development, as i can see it is still open.

matt-aitken:
v2 is deprecated as of Jan 31 2025

PR #474, opened 2023-09-13
Add a function call example for OpenAI

prabaljainn:
/claim #469
I've inserted a section of code that simulates making a function call via the OpenAI API. The system message establishes that the assistant can make function calls, and the user message specifies the function call ("Function: Tell me a joke about programming").

If OpenAI were to support function calls natively in the future, then the function_call parameter (which is hypothetical and not currently supported) could be used to specify the function name and any necessary arguments.

### DIFF
--- a/examples/job-catalog/src/openai.ts
+++ b/examples/job-catalog/src/openai.ts
@@ -34,15 +34,18 @@ client.defineJob({
       });
     }
 
-    await io.openai.backgroundCreateChatCompletion("background-chat-completion", {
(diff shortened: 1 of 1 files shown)


prabaljainn:
@ericallam let me know if you need any changes or something.

ericallam:
We appreciate the contribution but this issue was assigned to @RamK777-stack and he's submitted a competing PR, so I'm just going to close this for now 👍

PR #475, opened 2023-09-14
feat: Added example for openAI function call

### LINKED ISSUE #469: [TRI-1273] Create a function call example for OpenAI

RamK777-stack:
Added function call example in openAI.

/claim #469 
/closes #469

Note: 

I can't able to test the changes because I don't have an active openAI key. I consumed all of my free quota 🥺🥺

image

### DIFF
--- a/examples/job-catalog/src/openai.ts
+++ b/examples/job-catalog/src/openai.ts
@@ -10,6 +10,59 @@ export const client = new TriggerClient({
   ioLogLocalEnabled: true,
 });
 
+const functions = [
+  {
+    "name": "get_current_affairs",
+    "description": "Get the current affairs from google",
+    "parameters": {
+      "type": "object",
+      "properties": {
+        "query": {
+          "type": "string",
+          "description": "query to search for google result",
+        }
+      },
+      "required": ["query"],
+    },
+  }
+]
+
+type ChatMessage = {
+  role: 'system' | 'user' | 'assistant' | 'function';
+  content: string;
+  name?: string;
+  function_call?: () => void
+};
+
+const chatMessages: ChatMessage[] = [
+  {
+    role: "user",
+    content: "who won fifa world cup 2023 with some related links to read more about it",
+  }
+]
+
+const get_current_affairs = async (query: string) => {
+  const params = {
+    key: process.env["GOOGLE_CUSTOM_SEARCH_API_KEY"] || '',
+    cx: process.env["googleCSEId"] || '',
+    q: query
+  }
+  const queryString = new URLSearchParams(params).toString();
+  const apiUrl = `https://www.googleapis.com/customsearch/v1?${queryString}`;
+  const response = await fetch(apiUrl);
+  const data = await response.json();
+  const results = data?.items?.slice(0, 5)?.map((item: any) => ({
+    title: item.title,
+    link: item.link,
+    snippet: item.snippet,
+  })) || []
+  return JSON.stringify(results)
+}
+
+const available_functions: { [key: string]: (query: string) => Promise<string> } = {
+  get_current_affairs: get_current_affairs,
+};
+
 const openai = new OpenAI({
   id: "openai",
   apiKey: process.env["OPENAI_API_KEY"]!,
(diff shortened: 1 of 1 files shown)


examples/job-catalog/src/openai.ts
```diff
@@ -10,6 +10,59 @@ export const client = new TriggerClient({
...
+};
+
+const chatMessages: ChatMessage[] = [
+  {
+    role: "user",
+    content: "who won fifa world cup 2023 with some related links to read more about it",
+  }
+]
+
+const get_current_affairs = async (query: string) => {
```
ericallam, line 44: I don't think we need to really make a request to the Google APIs here, as that complicates the example. Let's just return some mock JSON

examples/job-catalog/src/openai.ts
```diff
@@ -74,6 +127,38 @@ client.defineJob({
       model: "text-embedding-ada-002",
       input: "The food was delicious and the waiter...",
     });
+
+    //Function call example for OpenAI
+    const createCompletion = async () => {
```
ericallam, line 121: This won't work, because the `key` param is the same on each call. Not sure why it's necessary to do two calls to `createChatCompetion`?

examples/job-catalog/src/openai.ts
```diff
@@ -74,6 +127,38 @@ client.defineJob({
...
+      const chatCompletion = await io.openai.createChatCompletion("chat-completion", {
```
ericallam, line 122: We should use the new way of doing this `io.openai.chats.completions.create`

examples/job-catalog/src/openai.ts
```diff
@@ -74,6 +127,38 @@ client.defineJob({
...
+        functions,
+        function_call: "auto"
+      });
+      return chatCompletion
+    }
+
+    const chatCompletion = await createCompletion();
+
+    // @ts-ignore
+    const response = chatCompletion?.data?.choices[0].message;
```
ericallam: Why is this ts-ignore necessary?

examples/job-catalog/src/openai.ts
```diff
@@ -74,6 +127,38 @@ client.defineJob({
...
+
+    if ('function_call' in response) {
+      const function_name: string = "get_current_affairs";
+      const function_call = available_functions[function_name];
+      const args = JSON.parse(response['function_call']['arguments'])
+      // @ts-ignore
+      const fn_response = await function_call(...Object.values(args));
```
ericallam: Let's try and do this without any ts-ignores if possible.

examples/job-catalog/src/openai.ts
```diff
@@ -74,6 +127,38 @@ client.defineJob({
...
+      chatMessages.push({
+        role: 'function',
+        name: function_name,
+        content: fn_response
+      })
+      const finalCompletion = await createCompletion();
+      // @ts-ignore
+      await io.logger.info(finalCompletion?.data?.choices[0].message);
```
ericallam: In general, could you add more comments in the code to clearly explain what is going on?

examples/job-catalog/src/openai.ts
RamK777-stack: In the first call, openAI returns JSON which contains the function name and arguments

Example:
{'index': 0,
 'message': {'role': 'assistant',
  'content': None,
  'function_call': {'name': 'get_n_day_weather_forecast',
   'arguments': '{\n  "location": "Glasgow, Scotland",\n  "format": "celsius",\n  "num_days": 5\n}'}},
 'finish_reason': 'function_call'}
 
 Using this we need to call our functions. and pass the result to GPT again for summarization. so we need 2 calls. 

Reference: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_call_functions_with_chat_models.ipynb

examples/job-catalog/src/openai.ts
RamK777-stack: I didn't understand this part @ericallam, 

Do I need to use await io.openai.chats.completions.create instead of await io.openai.createChatCompletion ?

examples/job-catalog/src/openai.ts
RamK777-stack: I have removed this now and sent mock response !!

examples/job-catalog/src/openai.ts
RamK777-stack: Added more comments now !!

RamK777-stack:
Thanks @ericallam 
1. I have added more comments.
2. Removed Google API request and mocked responses.
3. Fixed ts errors.

Please check and share your feedback.

RamK777-stack:
HI @ericallam Did you get a chance to look into this? I have made the changes.

PR #478, opened 2023-09-15
Going exponential with `Linear`

nicktrn:
## Description

This adds a Linear integration with triggers, tasks, and helpers.

## Testing

Added job examples to the `job-catalog`.

## Changelog

- Added a brand new and shiny Linear integration with docs
- Added pagination helper `getAll()` and docs
- `io.runTask()` error handlers can now cause tasks to `skipRetrying`

## Todo
- [x] Webhook tasks
- [x] Webhook triggers
- [x] Webhook triggers (details)
- [x] Base tasks
- [x] Additional tasks
- [x] Payload examples
- [x] Final integration catalog entry
- [x] Final job catalog examples
- [x] Linear docs


integrations/linear/src/events.ts
```diff
@@ -0,0 +1,376 @@
...
+  icon: "linear",
+  filter: {
+    action: ["remove"],
+  },
+  parsePayload: (payload) => payload as ProjectEvent,
+  runProperties: (payload) => [{ label: "Change action", text: payload.action }],
+};
+
+// TODO: think of a better naming scheme (clashes with ProjectUpdate entity)
+export const onProject_Update: EventSpecification<ProjectEvent> = {
```
nicktrn: ~~Can anyone think of a sensible naming scheme?~~

`Project` and `ProjectUpdate` entities exist.

onProjectUpdate - any ProjectUpdate changes
onProjectUpdate - a Project update
onProjectUpdateUpdate - a ProjectUpdate update

:exploding_head: 

_edit:_ Will just go with past tense, i.e.

onProjectUpdate - any ProjectUpdate changes
onProjectUpdated - a Project update
onProjectUpdateUpdated - a ProjectUpdate update

integrations/linear/src/index.ts
```diff
@@ -0,0 +1,330 @@
...
+      errorCallback
+    );
+  }
+
+  // TODO: create separate sources to remove resourceTypes
+
+  onAttachment(params: OnChangeParams = {}) {
+    return createTrigger(this.source, events.onAttachment, {
+      ...params,
+      resourceTypes: ["Attachment"],
```
nicktrn: This TODO would create 8 sources in total. Would simplify things re dynamic triggers, but we'd also end up with more endpoints. Sound reasonable?

_edit:_ Didn't need params for this in the first place.

integrations/linear/src/schemas.ts
```diff
@@ -0,0 +1,263 @@
...
+      id: z.string(),
+      color: z.string(),
+      name: z.string(),
+    })
+  ),
+  description: z.string(),
+});
+
+// FIXME: can't confirm schema as does not seem to trigger
+const AttachmentDataSchema = z.object({});
```
nicktrn: There was a mention of this being in alpha, will have another look.

_edit:_ Triggers on URL attachments, but should be used with caution. Will add a comment re alpha.

integrations/linear/src/webhooks.ts
```diff
@@ -0,0 +1,341 @@
...
+
+// TODO: types
+const withoutFunctions = <T>(obj: T): T => {
+  return JSON.parse(JSON.stringify(obj), (key, value) => {
+    if (typeof value === "function") {
+      return undefined;
+    }
+    return value;
+  });
+};
```
nicktrn: This will likely have to be used in many places to get serializable output. Plan is to create a useful generic to go along with this.

Reason: No easy way to get clean entity outputs from the Linear SDK. This goes for all endpoints. I would like to avoid using raw GraphQL queries.

integrations/linear/src/webhooks.ts
```diff
@@ -0,0 +1,341 @@
...
+    // TODO: tidy up return type
+  ): Promise<Omit<WebhookPayload, "webhook"> & { webhook: Webhook | undefined }> {
+    return this.runTask(
+      key,
+      async (client, task, io) => {
+        const payload = await client.createWebhook(params);
+        return withoutFunctions({
+          ...payload,
+          webhook: await payload.webhook,
+        });
```
nicktrn: Problem: The Linear SDK likes to use getters. Also, `JSON.stringify()` will expose private properties (that could be hidden via the return type), so it's not possible to remove functions generally. I could add a deep function remover without stringification instead.

integrations/linear/src/webhooks.ts
```diff
@@ -0,0 +1,341 @@
...
+        event: allEvents,
+      };
+
+      // easily identify webhooks on linear
+      const label = `trigger.${params.teamId ? params.teamId : "all"}`;
+
+      // TODO: remove tunnel
+      const url = process.env["DEV_TUNNEL"]
+        ? httpSource.url.replace("http://localhost:3030", process.env["DEV_TUNNEL"])
+        : httpSource.url;
```
nicktrn: This has been useful for dev via ngrok et al. Are there plans to integrate tunnels for webhooks more generally? Maybe they already exist and I missed something!

~~_edit:_ My bad, just need to set the api URL to anything but localhost~~

Will leave this as is during dev unless I have time to find a better solution.

integrations/linear/src/webhooks.ts
```diff
@@ -0,0 +1,341 @@
...
+        allPublicTeams: !params.teamId,
+        label,
+        resourceTypes: allEvents,
+        secret: httpSource.secret,
+        teamId: params.teamId,
+        url,
+      });
+
+      // TODO
+      // if(!createPayload.success)
```
nicktrn: Add specific error handling if any webhook operation fails or just throw via schema parsing as currently?

integrations/linear/src/webhooks.ts
```diff
@@ -0,0 +1,341 @@
...
+    register: async (event, io, ctx) => {
+      const { params, source: httpSource, options } = event;
+
+      // (key-specific) stored data, undefined if not registered yet
+      const webhookData = WebhookRegistrationDataSchema.safeParse(httpSource.data);
+
+      console.log(webhookData);
+      if (!webhookData.success) {
+        console.log(webhookData.error);
+      }
```
nicktrn: Will remove this.

integrations/linear/src/webhooks.ts
```diff
@@ -0,0 +1,345 @@
...
+  const body = JSON.parse(rawBody);
+  const webhook = new LinearWebhooks(source.secret);
+
+  // TODO: might want to do two passes, with and without timestamp (delay parsing)
+  if (!webhook.verify(Buffer.from(rawBody), signature, body[LINEAR_WEBHOOK_TS_FIELD])) {
+    logger.error("[@trigger.dev/linear] Error validating webhook signature, they don't match");
+    throw Error("[@trigger.dev/linear] Invalid signature");
+  }
+
+  const webhookPayload = WebhookPayloadSchema.parse(body);
```
nicktrn, line 302: This results in the following error caused by a circular dependency:
```
TypeError: Cannot read properties of undefined (reading '_parse')
```

apps/webapp/app/services/externalApis/integrations/linear.ts
```diff
@@ -0,0 +1,335 @@
+import type { HelpSample, Integration, ScopeAnnotation } from "../types";
+
+const repoAnnotation: ScopeAnnotation = {
```
ericallam: this file looks to be just a duplicate of the Github one

nicktrn: Yes, just needed a quick placeholder for testing, already on my todos

integrations/linear/src/webhooks.ts
ericallam: The `httpSource.url` is provided by the webapp, and uses the `APP_ORIGIN` env var that needs to be set when running the webapp locally.  For example: `APP_ORIGIN='http://abc1234.ngrok.io'`

integrations/linear/src/webhooks.ts
ericallam: Throwing via schema parsing is fine for now 👍

integrations/linear/src/webhooks.ts
nicktrn: Excellent. Any unintended side-effects? Would rather not force local traffic via the tunnel, just incoming webhooks.

apps/webapp/app/services/externalApis/integrations/linear.ts
```diff
@@ -0,0 +1,128 @@
+import type { HelpSample, Integration } from "../types";
+
+const usageSample: HelpSample = {
+  title: "Using the client",
+  code: `
+import { Linear, events } from "@trigger.dev/linear";
+
+const linear = new Linear({
+  id: "__SLUG__",
+  token: process.env.LINEAR_TOKEN!,
```
matt-aitken: For the OAuth sample, we don't want this `token` line. 

I created a function in the Airtable one to deal with this: https://github.com/triggerdotdev/trigger.dev/blob/cd70a299707ccaa3657e1c271e2a593a72908fe8/apps/webapp/app/services/externalApis/integrations/airtable.ts#L3

integrations/linear/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@trigger.dev/linear",
+  "version": "0.0.1",
```
matt-aitken: Before we merge the PR, we'll need to look at the latest version of the other packages in main and set this to the same version.

We decided to have lockstep versioning for all our packages, i.e. they're always the same version as each other so compatibility and upgrading is easy to figure out. Remix do the same thing

integrations/linear/src/index.ts
```diff
@@ -0,0 +1,423 @@
...
+  if (rateLimitRemaining === "0" && rateLimitReset) {
+    const resetDate = new Date(Number(rateLimitReset) * 1000);
+
+    return {
+      retryAt: resetDate,
+      error,
+    };
+  }
+
+  const queryComplexity = error.raw?.response?.headers?.get("X-Complexity");
```
matt-aitken, line 2001: This is really cool 😸 I didn't know they did this. I've never created a GraphQL API, I'm guessing this is more standard in them?

integrations/linear/src/webhooks.ts
matt-aitken: Totally agree on avoiding GraphQL queries wherever possible.

We could add a function like this (I think you've updated it) to packages/integration-kit? There are some other useful things in there

integrations/linear/src/webhooks.ts
matt-aitken: It looks like you might have solved this problem now?

matt-aitken (changes requested):
@nicktrn this is really good

integrations/linear/package.json
nicktrn: Good catch. Could probably grab the latest version during `create-integration`:
https://github.com/triggerdotdev/trigger.dev/blob/c2ce707f3d1d453a050dff613b4cd9fc198dd8c5/packages/cli/src/commands/createIntegration.ts#L97

..but would still have to update before merge as you mentioned.

integrations/linear/src/index.ts
nicktrn: Indeed. Nesting, batching, circular references (hopefully not), etc. Simple rate limits just don't do it here.

integrations/linear/src/webhooks.ts
nicktrn: Yes, I think it's going to be useful for many integrations to come.

Right now it's still still Linear-specific, as I used a simple template literal to filter out their underscored private props:
https://github.com/triggerdotdev/trigger.dev/blob/b3fec827deb6159f00b897788ef4fcba5507ba8f/integrations/linear/src/types.ts#L16

integrations/linear/src/webhooks.ts
nicktrn: Yes, thought I'd marked this as resolved already! Extended the JSON approach and added a type to keep it simple.

integrations/linear/src/schemas.ts
```diff
@@ -0,0 +1,423 @@
...
+    action: REMOVE,
+  }),
+  ReactionEventBaseSchema.extend({
+    action: UPDATE,
+    updatedFrom: ReactionDataSchema.partial(),
+  }),
+]);
+export type ReactionEvent = z.infer<typeof ReactionEventSchema>;
+
+export const WebhookPayloadSchema = z.union([
```
nicktrn, line 432: This was once a discriminated union (as it should be), but there were issues with nesting. See:
- https://github.com/colinhacks/zod/issues/1618
- https://github.com/colinhacks/zod/issues/2523

Even with the proposed workarounds it always ended up breaking somewhere.

nicktrn:
This is almost finished. Mainly docs now.

I've added the ability to `skipRetrying` on runTask errors in https://github.com/triggerdotdev/trigger.dev/pull/478/commits/cb6e260430fd448a4ebda902c936955f5e09b539.

This overrides any existing retry options. Felt this was needed - some errors are final, e.g. user input errors. Retrying will just waste resources and could lead to being rate limited unnecessarily.

**Specific example**: Hitting single-query complexity limits produces an invalid input error that won't change by retrying.

matt-aitken (approved):
This is brilliant @nicktrn

PR #493, opened 2023-09-21
React status hooks

matt-aitken:
## Description

You can now report the status from a Job and then easily display that in your UI. This is a much more powerful way of building UIs to show the state of your background jobs.

## Testing

- Updated the `references/nextjs-reference` project to use this
- Published the packages with a snapshot and tested in the openai-text-summarizer example project

## Changelog

- Created `io.createStatus()` and `status.updateStatus()` so you can create and update a status from inside a Job
- Created `useEventRunStatuses()` and `useRunStatuses()` that allow you to display those statuses in your frontend
- Created `client.getRunStatuses()` so you can get the statuses from your backend.
- Added the associated API endpoints
- Updated the React hooks guide in the docs to include these changes


packages/core/src/schemas/runs.ts
```diff
@@ -78,6 +79,8 @@ export const GetRunSchema = RunSchema.extend({
   output: z.any().optional(),
   /** The tasks from the run */
   tasks: z.array(RunTaskWithSubtasksSchema),
+  /** Any status updates that were published from the run */
+  statuses: z.array(JobRunStatusRecordSchema),
```
ericallam: Let's double check that this isn't going to break any backwards compatibility stuff. Maybe this should be optional or `.default([])`

matt-aitken: Great point, I've added a default

PR #495, opened 2023-09-21
feat: zod version checker

### LINKED ISSUE #436: [TRI-1230] TypeScript "Type instantiation is excessively deep and possibly infinite.ts(2589)" error when using eventTrigger w/zod schema (different Zod versions)
When using the `eventTrigger` with zod to define the event schema TypeScript gives this error:

This only seems to be happening in Next.js projects created with `create-t3-app`. I've reproduced it here:

[https://github.com/ericallam/t3-trigger-app/blob/bbade197052ae5d790c7b46509628a5c8d57ba94/src/jobs/examples.ts#L19](https://github.com/ericallam/t3-trigger-app/blob/bbade197052ae5d790c7b46509628a5c8d57ba94/src/jobs/examples.ts#L19)

We need to track down the issue and figure out how to fix it (or provide a workaround for t3-app)

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1230](https://linear.app/triggerdotdev/issue/TRI-1230/typescript-possibly-infinite-error-when-using-eventtrigger-wzod-schema)</sub>

neo773:
## ✅ Checklist

https://github.com/triggerdotdev/trigger.dev/assets/62795688/5c799008-22ca-4fe4-9f63-34abf112d541

/closes #436
/claim #436


packages/cli/src/commands/dev.ts
```diff
@@ -14,6 +14,12 @@ import { resolvePath } from "../utils/parseNameAndPath";
 import { TriggerApi } from "../utils/triggerApi";
 import { run as ncuRun } from 'npm-check-updates'
 import chalk from "chalk";
+import path from 'path';
+import { createRequire } from 'node:module';
+import { PackageFile } from "npm-check-updates/build/src/types/PackageFile";
```
matt-aitken, line 19: Is this not exported from the main index file? When you build the package I think this will fail

neo773: @matt-aitken 
Built fine, since it's a type only import

image

ericallam:
Thanks for this 🙏 but it's no longer necessary to check for the Zod version because of #500, so closing this 👍

PR #507, opened 2023-09-26
`Replicate` integration and remote callbacks

nicktrn:
## Description

This adds a Replicate integration and support for tasks with [remote callbacks](https://github.com/triggerdotdev/trigger.dev/discussions/502).

## Testing

See `references/job-catalog`

## Changelog

- Support tasks with remote callbacks
- Inherit common tsconfig for integrations
- Add Replicate integration
- Add pagination helpers
- Use inferred types for all `io.runTask()` calls

## Todo
- [x] Remote callbacks
- [x] Replicate tasks
- [x] Replicate helpers
- [x] Fail with error on timeout
- [x] Replicate docs
- [x] Update `io.runTask()` docs
- [x] More job-catalog entries
- [x] Sensible `io.runTask()` return type
- [x] Mirror final changes to `performRunExecutionV1`
- [x] Add changeset


apps/webapp/app/routes/api.v1.runs.$runId.tasks.$id.callback.$secret.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  const { runId, id } = ParamsSchema.parse(params);
+
+  // Now parse the request body
+  const anyBody = await request.json();
+
+  // Allows any valid object
+  // TODO: maybe add proper schema parsing during io.runTask(), or even skip this step
+  const body = CallbackTaskBodyInputSchema.safeParse(anyBody);
```
nicktrn: Currently `z.object({}).passthrough()`. Leave as is? Change to `z.any()` or even remove schema parsing? There are no user-provided schemas for remote callbacks (yet), and we couldn't use them here.

packages/trigger-sdk/src/io.ts
```diff
@@ -621,6 +605,11 @@ export class IO {
           task,
         });
 
+        // TODO: empty return? maybe don't even parse first
+        if (task.status === "WAITING" && task.callbackUrl) {
+          return output;
+        }
```
nicktrn: ~~Ideas? Gets discarded anyway.~~

_edit:_ Will return prior to parsing. Throwing immediately after.

nicktrn:
@ericallam Take it apart :scissors: 

Didn't find a clean way to narrow the callbackUrl type, so took out the workaround in the end.

Same goes for the `io.runTask()` return, but will just add some docs re type casting the async return for now.

apps/webapp/app/routes/api.v1.runs.$runId.tasks.$id.callback.$secret.ts
ericallam: I think we can probably skip the schema parsing and just save the body to the task output without it

apps/webapp/app/routes/api.v1.runs.$runId.tasks.ts
```diff
@@ -217,6 +228,15 @@ export class RunTaskService {
...
           { tx, runAt: task.delayUntil ?? undefined }
         );
+      } else if (task.status === "WAITING" && callbackUrl && taskBody.callback) {
+        // We need to schedule the callback timeout
+        await workerQueue.enqueue(
+          "processCallbackTimeout",
+          {
+            id: task.id,
+          },
+          { tx, runAt: new Date(Date.now() + taskBody.callback.timeoutInSeconds * 1000) }
```
ericallam: What happens if the timeout isn't set? Is it required? Might want to only schedule this if the timeout is set and larger than 0

nicktrn: The timeout currently defaults to an hour

apps/webapp/app/routes/api.v1.runs.$runId.tasks.ts
```diff
@@ -233,16 +238,21 @@ export class RunTaskService {
...
+          delayUntilInFuture || callbackEnabled || taskBody.trigger
             ? "WAITING"
             : taskBody.noop
             ? "COMPLETED"
             : "RUNNING";
       }
 
+      const taskId = ulid();
+      const callbackUrl = callbackEnabled
+        ? `${env.APP_ORIGIN}/api/v1/runs/${runId}/tasks/${taskId}/callback/${generateSecret()}`
```
ericallam: Could you update the `generateSecret` util function to receive a param with number of bytes (defaulted to 32) and pass in 12 here? The URLs are a little long using a 32 byte hex secret.

integrations/replicate/src/predictions.ts
```diff
@@ -0,0 +1,92 @@
...
+          ...params,
+          webhook: task.callbackUrl ?? "",
+          webhook_events_filter: ["completed"],
+        });
+      },
+      {
+        name: "Create And Await Prediction",
+        params,
+        properties: [...createPredictionProperties(params), ...callbackProperties(params)],
+        callback: { enabled: true },
```
ericallam: These have default timeouts right?

nicktrn: Yes

docs/sdk/io/runtask.mdx
```diff
@@ -201,4 +219,43 @@ client.defineJob({
...
+```typescript Remote callbacks
+client.defineJob({
+  id: "remote-callback-example",
+  name: "Remote Callback example",
+  version: "0.1.1",
+  trigger: eventTrigger({ name: "predict" }),
+  integrations: { replicate },
+  run: async (payload, io, ctx) => {
+    //runTask
+    const prediction = await io.runTask(
```
ericallam: I think this is meant to be `io.replicate.runTask` right?

nicktrn: Yes, good catch!

nicktrn: Had to update the GitHub example too, as underlying clients are not exposed at `io.integration.client` anymore. Also added a section to mention the wrappers at `io.integration.runTask()`.

docs/sdk/io/runtask.mdx
```diff
@@ -6,6 +6,8 @@ description: "`io.runTask()` allows you to run a [Task](/documentation/concepts/
 
 A Task is a resumable unit of a Run that can be retried, resumed and is logged. [Integrations](/integrations) use Tasks internally to perform their actions.
 
+The wrappers at `io.integration.runTask()` expose the underlying Integration client as the first callback parameter (see examples on the right). They will have defaults set for options and `onError` handlers, but should otherwise be considered identical to raw `io.runTask()`.
+
```
nicktrn, line 10: New section on integration `io.runTask()` wrappers

docs/sdk/io/runtask.mdx
```diff
@@ -150,11 +170,11 @@ client.defineJob({
...
   run: async (payload, io, ctx) => {
     //runTask
-    const response = await io.runTask(
+    const response = await io.github.runTask(
       "create-card",
-      async () => {
+      async (client) => {
         //create a project card using the underlying GitHub Integration client
-        return io.github.client.rest.projects.createCard({
+        return client.rest.projects.createCard({
```
nicktrn, line 177: Updated `io.runTask()` example

nicktrn (commented):
Added a section on runTask wrappers as we have to make use of them now in the examples. (Underlying clients are not directly exposed anymore)

@matt-aitken You may want to have a brief look at this. Must be a remnant from the last big integration update.

ericallam (approved):

PR #511, opened 2023-09-27
CLI create-integration templates and shared configs

nicktrn:
Provides templates to make creating new integrations easier. May need some trimming down and more comments.

~~As many replacements are outside of strings, I appended `.txt` to hide errors. Might want to think about using a proper templating language in the future, e.g. nunjucks or liquid.~~

I used `*.j2` file extensions as the [jinja2 syntax highlighting plugin](https://github.com/samuelcolvin/jinjahtml-vscode) is much better than liquid and detects the underlying file type, not just templating language.

## Testing

Build and run via `pnpm start create-integration`

---

## Changelog

- Extend from common tsup and tsconfig
- Templating for `create-integration`
- Templates for everything
- Always ask for auth method
- Use [liquidjs](https://github.com/harttle/liquidjs) for templates

---

## Screenshots

cli


matt-aitken (approved):
The code looks good to me 👍

PR #522, opened 2023-09-29
Improves the performance of run resuming

ericallam:
When runs resume, we try and make sure that tasks that have already been completed are cached and reused. Worst case scenario the client needs to hit the API server once for a non-cached task that is indeed completed on the server, but this can get pretty expensive when there are a larger number of tasks.

This commit does 2 different things to help:

- noop tasks (logs and waits) are no longer “cached” using the eager-loading strategy, instead their idempotency keys are shoved into a bloom filter and the client tests for their inclusion in the bloom filter before running them (since they don’t have any concept of output, this works)
- Additional cached tasks are lazy loaded when a task is run. This allows us to progressively fetch additional tasks to be cached on the client, which will cut down on cache misses by a decent amount.

I'm also introducing in this PR a new way to do API versioning, using the `Trigger-Version` header, see here for the client side handling of this:

https://github.com/triggerdotdev/trigger.dev/blob/a9118be1a1d65d9ae94e1c6c9610db914a6311fd/packages/trigger-sdk/src/apiClient.ts#L111

And here for the server side:

https://github.com/triggerdotdev/trigger.dev/blob/a9118be1a1d65d9ae94e1c6c9610db914a6311fd/apps/webapp/app/routes/api.v1.runs.%24runId.tasks.ts#L88

This allows the client to be backwards (old clients work with new version of API) and forwards (new clients working with old versions of the API) compatible.

Oh yea, I also added a new function on IO called `io.yield()` which allows you to yield execution of the current run and resume it in a new function execution. Similar to `io.wait()` but does not create a task and resumes execution immediately. 

## ✅ Checklist

---

## Testing

I've been testing this using the `stressTest` in job-catalog, but more testing will need to take place before this goes live.

---


ericallam:
I've now tested the new client with an older version of the platform and made changes to yield to improve backwards compat with older versions of the Trigger.dev server.

packages/trigger-sdk/src/io.ts
```diff
@@ -699,6 +803,30 @@ export class IO {
     return this._taskStorage.run({ taskId: task.id }, executeTask);
   }
 
+  /**
+   * `io.yield()` allows you to yield execution of the current run and resume it in a new function execution. Similar to `io.wait()` but does not create a task and resumes execution immediately.
+   */
+  yield(key: string) {
+    if (!supportsFeature("yieldExecution", this._version)) {
```
ericallam: Rename this _serverVersion

packages/trigger-sdk/src/io.ts
```diff
@@ -549,19 +603,59 @@ export class IO {
...
+        cachedTasksCursor: this._cachedTasksCursor,
+      }
+    );
+
+    const task =
+      response.version === API_VERSIONS.LAZY_LOADED_CACHED_TASKS
+        ? response.body.task
+        : response.body;
+
+    if (response.version === API_VERSIONS.LAZY_LOADED_CACHED_TASKS) {
```
ericallam, line 644: this should use `supportsFeature`

PR #524, opened 2023-10-02
Features: Batching and graphile-worker upgrade to `0.16.0-rc.1`

### LINKED ISSUE #348: [TRI-1081] Feature idea: Batch Event Trigger

### LINKED ISSUE #396: [TRI-1152] chore: upgrade webapp to use graphile-worker `0.14.0-rc.0`

nicktrn:
Closes #396 
Closes #348 

## ✅ Checklist

- [x] More testing just to be safe
- [x] Create migration docs

---

### Migration

1. Create this `docker-compose.yml`

```yaml
version: "3"

volumes:
  data:

networks:
  app_network:
    external: false

x-db-url: &db-url postgresql://postgres:postgres@db:5432/postgres

x-env: &env
  LOGIN_ORIGIN: http://localhost:3030
  APP_ORIGIN: http://localhost:3030
  PORT: 3030
  REMIX_APP_PORT: 3030
  MAGIC_LINK_SECRET: secret
  SESSION_SECRET: secret
  ENCRYPTION_KEY: ae13021afef0819c3a307ad487071c06
  DATABASE_URL: &db-url
  DIRECT_URL: &db-url
  GRACEFUL_SHUTDOWN_TIMEOUT: 5000
  NODE_ENV: production
  RUNTIME_PLATFORM: docker-compose
  TRIGGER_LOG_LEVEL: debug

services:
  old-webapp:
    image: ghcr.io/triggerdotdev/trigger.dev:v2.2.21
    restart: "no"
    environment:
      <<: *env
      WORKER_ENABLED: false
      EXECUTION_WORKER_ENABLED: false
      TASK_OPERATION_WORKER_ENABLED: false
    ports:
      - 3065:3030
    depends_on:
      - db
    networks:
      - app_network

  old-workers:
    image: ghcr.io/triggerdotdev/trigger.dev:v2.2.21
    restart: "no"
    environment:
      <<: *env
      HTTP_SERVER_DISABLED: true
    depends_on:
      - db
    networks:
      - app_network

  new-webapp:
    image: ghcr.io/triggerdotdev/trigger.dev:batching-rc.2
    restart: always
    environment:
      <<: *env
      WORKER_ENABLED: false
      EXECUTION_WORKER_ENABLED: false
      TASK_OPERATION_WORKER_ENABLED: false
    ports:
      - 3030:3030
    depends_on:
      - db
    networks:
      - app_network

  new-workers:
    image: ghcr.io/triggerdotdev/trigger.dev:batching-rc.2
    restart: always
    environment:
      <<: *env
      HTTP_SERVER_DISABLED: true
    depends_on:
      - db
    networks:
      - app_network
      
  db:
    image: postgres:14
    restart: always
    volumes:
      - data:/var/lib/postgresql/data/
    environment:
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: postgres
      POSTGRES_DB: postgres
    networks:
      - app_network
    ports:
      - 3085:5432
```

2. Start up the DB

```bash
docker-compose up -d db
```

3. Start up the old webapp and workers

```bash
docker-compose up old-webapp old-workers
```

4. In another window, start up the new webapp and workers

```bash
docker-compose up new-webapp new-workers
```

5. To delete the DB and start again, maybe after altering some env vars

```bash
docker-compose down -v
```

### Batching

https://github.com/triggerdotdev/trigger.dev/blob/2afa19995e1cee973d25a278256e66e667f304b2/references/job-catalog/src/events.ts#L156-L211

---

## Changelog

- Upgrade to `v0.16.0-rc.1`
- Detect migration errors and point to docs
- Add migration delay to allow graceful shutdown to complete (rolling deploy)
- Insert custom `add_batch_job` function
- Add batched event triggers and webhook delivery
- Read replica support via optional `DATABASE_READ_REPLICA_URL` env var

---

## Screenshots

graphile-migration

💯


nicktrn:
`0.15.0` was released yesterday. The only change being that migration files are now embedded as strings. Just in case of unforeseen bugs, we should stay on `0.14.0-rc.0` for now.

apps/webapp/app/services/db/pgListen.server.ts
```diff
@@ -0,0 +1,48 @@
+import { logger } from "~/services/logger.server";
+import { Logger } from "@trigger.dev/core";
+import type { PoolClient } from "pg";
```
nicktrn: I've only added types for now. Will leave adding the package for when we actually need it. For now, we only listen to events within `zodWorker` where we can grab an existing `PoolClient`.

apps/webapp/app/services/db/pgListen.server.ts
```diff
@@ -0,0 +1,48 @@
...
+  public async call(channelName: string, callback: (payload: string) => Promise<void>) {
+    this.#logDebug("Registering notification handler", { channelName });
+
+    const isValidChannel = channelName.match(/^[a-zA-Z0-9:-_]+$/);
+
+    if (!isValidChannel) {
+      throw new Error(`Invalid channel name: ${channelName}`);
+    }
+
+    this.#poolClient.query(`LISTEN "${channelName}"`).then(null, (error) => {
```
nicktrn: `channelName` gets validated above to prevent SQL injections. Don't think we can pass it as a param to `query()`, which would prevent having to do this.

nicktrn:
Waiting for `v0.16` release and #689 before finishing this up.

apps/webapp/app/platform/zodWorker.server.ts
```diff
@@ -443,7 +573,8 @@ export class ZodWorker<TMessageCatalog extends MessageCatalogSchema> {
 
     await this.enqueue(helpers.job.task_identifier, payload, {
       runAt: helpers.job.run_at,
-      queueName: helpers.job.queue_name ?? undefined,
+      // TODO: fix queueName, maybe use reschedule helper
+      // queueName: helpers.job.job_queue_id ?? undefined,
```
nicktrn: `queue_name` is not directly exposed anymore and would have to be queried for. Could use [reschedule_jobs](https://worker.graphile.org/docs/admin-functions#rescheduling-jobs) instead but this will only work for unlocked jobs.

apps/webapp/app/platform/zodWorker.server.ts
```diff
@@ -51,10 +60,11 @@ const AddJobResultsSchema = z.array(GraphileJobSchema);
 
 export type ZodTasks<TConsumerSchema extends MessageCatalogSchema> = {
   [K in keyof TConsumerSchema]: {
-    queueName?: string | ((payload: z.infer<TConsumerSchema[K]>) => string);
+    queueName?: string | ((payload: z.infer<TConsumerSchema[K]>, jobKey?: string) => string);
```
nicktrn, line 63: This helps decrease payload size for batch jobs

apps/webapp/app/presenters/TestJobPresenter.server.ts
```diff
@@ -139,13 +139,18 @@ export class TestJobPresenter {
...
-      runs: job.runs.map((r) => ({
-        id: r.id,
-        number: r.number,
-        status: r.status,
-        created: r.createdAt,
-        payload: r.event.payload ? JSON.stringify(r.event.payload, null, 2) : undefined,
-      })),
+      runs: job.runs.map((r) => {
+        // For compatibility with old Job Runs where payload is only available on the related Event Record
+        const payload = r.payload !== null ? JSON.parse(r.payload) : r.event.payload;
```
nicktrn, line 144: We'll be able to simplify this in a lot of places once we migrate existing data and make JobRun.payload non-optional.

apps/webapp/app/services/endpointApi.server.ts
```diff
@@ -254,26 +254,34 @@ export class EndpointApi {
     return HttpSourceResponseSchema.parse(anyBody);
   }
 
-  async deliverWebhookRequest(options: {
+  // TODO: ensure backward compat
+  async deliverWebhookRequests(options: {
```
nicktrn: Quick comment so I don't forget to do this

apps/webapp/app/services/runs/createRun.server.ts
```diff
@@ -14,44 +14,61 @@ export class CreateRunService {
...
           jobId: job.id,
           versionId: version.id,
-          eventId: eventId,
+          eventId: firstEvent.id,
+          eventIds: eventRecords.map((event) => event.id),
           environmentId: environment.id,
           organizationId: environment.organizationId,
           projectId: environment.projectId,
           endpointId: endpoint.id,
-          externalAccountId: eventRecord.externalAccountId
```
nicktrn, line 51: Include this in `jobKey`

apps/webapp/app/services/runs/createRun.server.ts
```diff
@@ -14,44 +14,61 @@ export class CreateRunService {
...
-            ? eventRecord.externalAccountId
+          batched,
+          payload: JSON.stringify(
+            batched ? eventRecords.map((event) => event.payload) ?? [{}] : firstEvent.payload ?? {}
+          ),
+          externalAccountId: firstEvent.externalAccountId
+            ? firstEvent.externalAccountId
             : undefined,
-          isTest: eventRecord.isTest,
+          isTest: firstEvent.isTest,
```
nicktrn, line 75: Include in `jobKey` too

apps/webapp/app/services/sources/webhookDeliveryBatcher.server.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    const MAX_INTERVAL_IN_SECONDS = 10 * 60;
+
+    const deliverAfter = maxInterval ? Math.max(maxInterval, MAX_INTERVAL_IN_SECONDS) : undefined;
+
+    await workerQueue.enqueue(
+      "deliverMultipleWebhookRequests",
+      { webhookEnvironmentId, requestDeliveryIds }
+      // { runAt: deliverAfter ? deliverAfterToDate(deliverAfter) : undefined }
```
nicktrn: Should probably remove the `maxInterval` param as we already respect it for incomplete batches before they come through here.

apps/webapp/app/services/worker.server.ts
```diff
@@ -227,13 +246,37 @@ function getWorkerQueue() {
       },
     },
     tasks: {
+      "events.invokeDispatchBatcher": {
```
nicktrn, line 249: Couldn't think of a more confusing name.

apps/webapp/app/services/worker.server.ts
```diff
@@ -227,13 +246,37 @@ function getWorkerQueue() {
...
+        },
+      },
       "events.invokeDispatcher": {
         priority: 0, // smaller number = higher priority
         maxAttempts: 6,
         handler: async (payload, job) => {
           const service = new InvokeDispatcherService();
 
-          await service.call(payload.id, payload.eventRecordId);
+          await service.call(payload.id, [payload.eventRecordId]);
```
nicktrn, line 281: Takes an array now, but only batches if the dispatcher has batching enabled.

apps/webapp/app/services/endpointApi.server.ts
nicktrn: Will add batched delivery as a new action instead so we can support both more easily. Just a refactor, the mechanics won't change.

packages/trigger-sdk/src/triggerClient.ts
```diff
@@ -1322,6 +1371,38 @@ export class TriggerClient {
...
+        throw new Error("Context length needs to match number of event IDs.");
+      }
+
+      return {
+        ...sharedTriggerContext,
+        events: execution.eventIds.map((id, i) => ({
+          id,
+          name: execution.event.name,
+          context: parsedContext[i],
+          timestamp: execution.event.timestamp,
```
nicktrn, line 1442: Cheating here - timestamp comes from the first event. Timestamps would either be another thing to attach to JobRun, get for each execution, or remove.

Alternatively, we could create [many-to-many relations](https://www.prisma.io/docs/orm/prisma-schema/data-model/relations/many-to-many-relations#relational-databases) between JobRun and EventRecord. This would be a big change.

PR #530, opened 2023-10-02
fix: #526 CLI warn if using @trigger.dev/react package with Next.js pages directory

### LINKED ISSUE #526: [TRI-1334] [TRI-1331] CLI warning if using @trigger.dev/react package with Next.js pages directory
## Problem

We add the `@trigger.dev/react` package when you run `npx @trigger.dev/cli@latest init` in a Next.js project. The Next.js pages directory has a known issue with ESM modules.

We currently have an open issue but this still isn't great: [https://github.com/triggerdotdev/trigger.dev/issues/331](https://github.com/triggerdotdev/trigger.dev/issues/331)

## Solution

Show a clear warning in our CLI init command that gives you instructions on what to do, if you're using Next.js + pages router. 

### If on Next.js >13.1, add this to next.config file:

```
const nextConfig = {
  transpilePackages: ["@trigger.dev/react"],
};

module.exports = nextConfig;
```

### If on Next.js <13.1, add this to next.config file:

1. Install next-transpile-modules package

```
const withTM = require('next-transpile-modules')(['@trigger.dev/react']); // pass the modules you would like to see transpiled

module.exports = withTM({
  //...existing config object
});
```

### How to

* It makes most sense to add this to [the post-install step](https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/frameworks/nextjs/index.ts#L91).
* There's already a function to check for pages directory that can be used

  ```
  const nextJsDir = await detectPagesOrAppDir(path);
  ```
* [Here's some code](https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/frameworks/nextjs/index.ts#L111) that could potentially be adapted to check the Next.js version. But be careful when checking versions because special characters are used in package.json files, like "^" and "\*". So either you should get the actual version installed (not using the package.json) or use a semver package to do the comparison.
* Drawing a box using `boxen` would be a nice way to highlight this information, see [this code](https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/frameworks/express/index.ts#L39) for an example.

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1331](https://linear.app/triggerdotdev/issue/TRI-1331/cli-warning-if-using-triggerdevreact-package-with-nextjs-pages)</sub>

<sub>[TRI-1334](https://linear.app/triggerdotdev/issue/TRI-1334/[tri-1331]-cli-warning-if-using-triggerdevreact-package-with-nextjs)</sub>

neo773:
Closes #526

## ✅ Checklist

## Screenshots

SCR-20231003-dbpb


packages/cli/src/frameworks/nextjs/index.ts
```diff
@@ -89,6 +92,58 @@ export class NextJs implements Framework {
...
+    const nextJsDir = await detectPagesOrAppDir(path);
+    const nextConfigExists = await detectNextConfigFile(path);
+
+    const require = createRequire(import.meta.url);
+    const nextPath = require.resolve("next");
+    const reactPath = require.resolve("@trigger.dev/react");
+
+    if (nextConfigExists && nextJsDir === "pages" && typeof reactPath !== "undefined") {
+      const nextPackageJsonPath = pathModule.join(nextPath, "../../../", "package.json");
+      const nextPackageJsonData = JSON.parse(await fs.readFile(nextPackageJsonPath, "utf8"));
```
matt-aitken, line 105: It would be much better to use the current way we check for dependencies. 

This function gets the "next" dependency version, and then returns true if it exists. You could create a function that uses this same methodology where you can pass in a package name and it returns undefined or the version string.

https://github.com/triggerdotdev/trigger.dev/blob/6769d6b43930317045796cc31af7924f462df07a/packages/cli/src/frameworks/nextjs/index.ts#L111

packages/cli/src/frameworks/nextjs/index.ts
```diff
@@ -89,6 +92,58 @@ export class NextJs implements Framework {
           break;
       }
     }
+
+    const nextJsDir = await detectPagesOrAppDir(path);
```
matt-aitken, line 96: Let's create a separate function for all of this logic, and export it from this file so tests can be written.

Could you also wrap the function in a try catch block so if an error is thrown it doesn't break installs. It doesn't need to do anything in the catch.

matt-aitken (changes requested):

matt-aitken:
This is a bit stale now

PR #536, opened 2023-10-03
feat: allow CLI update to accept a version for trigger-dev package

### LINKED ISSUE #484: [TRI-1281] feat: Add an optional `to` param to `@trigger.dev/cli init update`

When running `npx @trigger.dev/cli@latest update` we should allow you to pass in a `to` option which would set the version to upgrade to (instead of the default which is "latest".

* Add a `to` option to the update command, default to "latest": [https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/cli/index.ts#L98](https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/cli/index.ts#L98)
* If the `to` flag isn't "latest" then get ncu to do the change without spitting out the confirmation. There's a flag called `upgrade` that should work. This file is where edits need to happen: [https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/commands/update.ts#L35](https://github.com/triggerdotdev/trigger.dev/blob/e350659e244e6976747f726984c83a780ba94882/packages/cli/src/commands/update.ts#L35)

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1281](https://linear.app/triggerdotdev/issue/TRI-1281/feat-add-an-optional-to-param-to-triggerdevcli-init-update)</sub>

hmacr:
Closes #484 

## ✅ Checklist

---

## Testing

- Verified that the update command accepts the version and uses it for dependency update.
- Verified that the updated command falls back to `latest` if the version is not provided.

---

## Changelog

- Allows CLI users to provide a version for trigger-dev package update.

💯

### DIFF
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -101,7 +101,6 @@ export async function updateCommand(projectPath: string, anyOptions: any) {
         newPackageJSON[tmp.type][packageName] = updatedDependencies[packageName];
       }
     });
-    console.log(newPackageJSON);
     await writeJSONFile(packageJSONPath, newPackageJSON);
     await installDependencies(projectPath);
   }
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -4,8 +4,24 @@ import { run, RunOptions } from "npm-check-updates";
 import { installDependencies } from "../utils/installDependencies.js";
 import { readJSONFileSync, writeJSONFile } from "../utils/fileSystem.js";
 import { logger } from "../utils/logger.js";
+import { z } from "zod";
 
-export async function updateCommand(projectPath: string) {
+export const UpdateCommandOptionsSchema = z.object({
+  to: z.string().optional(),
+});
+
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/update.ts
```diff
@@ -69,6 +87,40 @@ export async function updateCommand(projectPath: string) {
...
+
+  if (applyUpdates) {
+    const newPackageJSON = packageData;
+    packagesToUpdate.forEach((packageName) => {
+      const tmp = packageMaps[packageName];
+      if (tmp) {
+        newPackageJSON[tmp.type][packageName] = updatedDependencies[packageName];
+      }
+    });
+    console.log(newPackageJSON);
```
ericallam: Can you remove this console.log?

hmacr: Ah yes, missed it.

hmacr: Done!

PR #542, opened 2023-10-03
[TRI-1333] : Upgrade all packages/* and use fetch instead of node-fetch

### LINKED ISSUE #525: [TRI-1333] [TRI-1332] Upgrade all packages/* and use fetch instead of node-fetch
## Problem

Our packages currently support Node 16–17. This means we can't use `fetch` and we can't support some modern deployment platforms like Cloudflare/Vercel edge functions.

Node v18 is now LTS and v16 has entered "end-of-life" with no official support.

## Solution

* Set the engines.node property in all of the packages/\*/package.json files.
* Upgrade all packages/\* to support Node v18. 
* Change fetch to node-fetch
* Fix any other issues
* Ensure that typechecking has no errors: `pnpm run --filter "./packages/**" typecheck`

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1332](https://linear.app/triggerdotdev/issue/TRI-1332/upgrade-all-packages-and-use-fetch-instead-of-node-fetch)</sub>

<sub>[TRI-1333](https://linear.app/triggerdotdev/issue/TRI-1333/[tri-1332]-upgrade-all-packages-and-use-fetch-instead-of-node-fetch)</sub>

Rutam21:
Closes #525 

## ✅ Checklist

---

## Testing

Type checking has been done and no errors are found.

```bash
pnpm run --filter "./packages/**" typecheck
```

Testing

Smoke Testing has been done as suggested by @ericallam .

- **events job was run successfully and successful trigger was performed on the run. Find the screenshots below.**

events

- **delays job was run successfully and successful trigger was performed on the run. Find the screenshots below.**

delays

- **resend job was run successfully and successful trigger was performed on the run. Find the screenshots below.**

resend

- **resend job was run successfully and successful trigger was performed on the run. Find the screenshots below.**

### Jobs were also run successfully on the Trigger.dev cloud dashboard.

jobs

---

## Changelog

- All the Packages are updated to use the Node v18.0.0 LTS from now on.
- Certain other dependencies are updated using `pnpm`.
- `Node-Fetch` imports and dependency are removed in all packages and `fetch` is getting used as the native function as supported by Node 18 now.

---

## Screenshots

List of all modeified files are as follows.

File Change List

>Note: The integrations also use the older Node versions. If they need to be updated as well, I would love to work on that ticket too as I am familiar with such issues now.


Rutam21:
@matt-aitken Please review and suggest changes, if any. Thanks

ericallam:
Have you run and tested the code still works? The easiest way to do that is to run jobs in `references/job-catalog`:

https://github.com/triggerdotdev/trigger.dev/blob/main/references/job-catalog/README.md

Rutam21:
Hello @ericallam, So, I need to run all jobs `references/job-catalog` one by one and check if they run without any errors or running some of them is enough?
Steps for that would be as follows:

- Set .env file with the Trigger API Key
- Build the CLI
- Run each file in src
- Then run the Trigger Dev Command

Please let me know if it's the right way. Thanks.

ericallam:
No definitely don't run them all haha, I would just pick a few to run as a smoke test. The simplest is probably `events.ts`

Rutam21:
Great, I will 3-4 files in the src and then update with the screenshots here.
So, all we are looking for now is that they run and there's no error in the terminal, right? Or do I need to look at some other paramters as well?

ericallam:
Yea considering the change, seeing successful runs should be a good enough test. And since this is a client only change, you don't even have to run the server locally if you don't want to, you could just use cloud.trigger.dev

Rutam21:
Perfect, I will update the screenshots asap.

Rutam21:
@ericallam I have performed the tests and updated the test results in the comments above. Please review and let me know if it looks good. Thanks.

Rutam21:
@ericallam I have performed the tests and updated the test results in the comments above. Please review and let me know if it looks good. Thanks.

Rutam21:
@ericallam I have performed the tests and updated the test results in the comments above. Please review and let me know if it looks good. Thanks.

packages/trigger-sdk/package.json
```diff
@@ -56,6 +54,6 @@
     "typescript": "^4.8.4"
   },
   "engines": {
-    "node": ">=16.8.0"
+    "node": "^18.0.0"
```
ericallam: This limits the Node version to >= 18 and < 19, which isn't what we want. I'd just do `>=18.0.0`

ericallam (changes requested):
Also make sure to run `pnpm run typecheck` in the root of the repo to make sure all the types still are valid.

PR #556, opened 2023-10-04
docs: Enhance io.runTask() Callback Documentation

### LINKED ISSUE #553: [TRI-1362] docs: Add better documentation for how to use Task Callbacks
We currently only have a small amount of documentation for task callbacks in the [io.runTask()](https://trigger.dev/docs/sdk/io/runtask) docs but it would be great to flesh out the documentation in more detail, especially usage with just `io.runTask` instead of `io.replicate.runTask`, e.g.:

```ts
await io.runTask(
  "use-callback-url",
  async (task) => {
    // task.callbackUrl is the URL to call when the task is done
    // The output of this task will be the body POSTed to this URL
  },
  {
    name: "Use the callbackUrl to notify the caller when the task is done",
    callback: {
      enabled: true,
      timeoutInSeconds: 300, // If task.callbackUrl is not called within 300 seconds, the task will fail
    },
  }
);
```

Things the doc should cover:

 - When it makes sense to use callback URLs (a callback from an external system). Would be great to find another example of this other than Replicate
 - How to correctly type runTask so the output type is correct
 - How the request to the task URL works (POST and the entire JSON body becomes the task output)
 - How the timeout works

I think this would probably best live in Documentation -> Guides -> Task Callbacks

<sub>[TRI-1362](https://linear.app/triggerdotdev/issue/TRI-1362/docs-add-better-documentation-for-how-to-use-task-callbacks)</sub>

MeenuyD:
Closes #553 

Enhanced the io.runTask() documentation to cover task callbacks in detail. The update includes guidelines on when to use callback URLs, instructions for typing runTask() correctly for accurate output, an explanation of the POST request structure to the task URL, and a brief overview of the timeout functionality.

### DIFF
--- a/docs/documentation/guides/task-callbacks.mdx
+++ b/docs/documentation/guides/task-callbacks.mdx
@@ -1,0 +1,65 @@
+---
+title: "Task callbacks"
+sidebarTitle: "Task callbacks"
+description: "`io.runTask()` allows you to run a [Task](/documentation/concepts/tasks) from inside a Job run."
+---
+
+Callbacks in asynchronous programming are essential communication tools that enhance task functionality.
+
+In the context of io.runTask(), callbacks enable tasks to notify external systems once they are completed. 
+
+This asynchronous approach ensures seamless communication between different components of a system.
+
+
+
+## When it makes sense to use callback URLs
+
+Using callback URLs is invaluable in scenarios beyond Replicate.
+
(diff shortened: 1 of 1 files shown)


ericallam:
Thanks for your attempt but unfortunately I have to close this as the documentation is too far from being correct. I'd encourage you to take the time to understand the feature better or ask questions before submitting again. All the best 🫡

PR #563, opened 2023-10-05
feat: add internal flag to job-run table

### LINKED ISSUE #557: [TRI-1366] fix: JobRun needs an internal boolean column, to reflect the value in Job
We use internal jobs for things like webhook registration, and those jobs have `internal = true` set in the database. We also need to add an `internal` column to JobRun that is set to `true` when the job it's associated with is `internal = true`. This is just denormalization, so when we do usage queries (in the new usage page) we can easily eliminate internal runs from the usage. The migration for this should upgrade existing runs to the correct `internal` value.

<sub>[TRI-1366](https://linear.app/triggerdotdev/issue/TRI-1366/fix-jobrun-needs-an-internal-boolean-column-to-reflect-the-value-in)</sub>

hmacr:
Closes #557

## ✅ Checklist

---

## Testing

- Verified the migration is applied and the column is added to `JobRun` table.
- Verified the new job runs are using the `internal` flag from `Job` table (I did this by manually updating the `internal` flag for one of the jobs).

---

## Changelog

- Add new column `internal` in `JobRun` table (defaults to `false`)
- Set `JobRun` internal flag based on the job's definition in the `createRun` API.

💯

### DIFF
--- a/apps/webapp/app/services/runs/createRun.server.ts
+++ b/apps/webapp/app/services/runs/createRun.server.ts
@@ -70,6 +70,7 @@ export class CreateRunService {
             ? eventRecord.externalAccountId
             : undefined,
           isTest: eventRecord.isTest,
+          internal: job.internal,
         },
       });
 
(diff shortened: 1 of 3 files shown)


hmacr:
@ericallam I've moved the query as suggested.

PR #564, opened 2023-10-05
feat: add prisma SQL migration for job-run internal flag backfill

### LINKED ISSUE #557: [TRI-1366] fix: JobRun needs an internal boolean column, to reflect the value in Job

hmacr:
Closes #557

This builds on top of #563 and should ideally be merged after #563 goes live.

The intention is to separate the schema and API change from the backfill script. This way, the jobs & job runs that get created during the backfill execution can still have the correct `internal` flags set (the API takes care of it).

## ✅ Checklist

---

## Testing

- Verified the past runs of internal jobs have their job-run internal flag set to `true`.
- Verified that the migration runs only once.

---

## Changelog

- Add a backfill script that sets the `internal` flag for past `JobRun` entries.
- To avoid DB overload, chunks of jobs (100 - configurable) are used to load their associated job-runs. 

💯

### DIFF
--- a/apps/webapp/app/services/runs/createRun.server.ts
+++ b/apps/webapp/app/services/runs/createRun.server.ts
@@ -70,6 +70,7 @@ export class CreateRunService {
             ? eventRecord.externalAccountId
             : undefined,
           isTest: eventRecord.isTest,
+          internal: job.internal,
         },
       });
 
--- a/packages/database/prisma/migrations/20231005064823_add_job_run_internal/migration.sql
+++ b/packages/database/prisma/migrations/20231005064823_add_job_run_internal/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "JobRun" ADD COLUMN     "internal" BOOLEAN NOT NULL DEFAULT false;
--- a/packages/database/prisma/migrations/20231005123839_job_run_internal_backfill/migration.sql
+++ b/packages/database/prisma/migrations/20231005123839_job_run_internal_backfill/migration.sql
@@ -1,0 +1,4 @@
+UPDATE "JobRun"
+SET "internal" = "Job"."internal"
+FROM "Job"
+WHERE "JobRun"."jobId" = "Job"."id";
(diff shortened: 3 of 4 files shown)


hmacr:
Marking it as draft, since it has dependency on #563.
Needs to be rebased with the main branch once #563 is merged.

apps/webapp/prisma/migrations/jobRunInternal.ts
```diff
@@ -0,0 +1,74 @@
+import { prisma, PrismaClientOrTransaction } from "~/db.server";
+import chunk from "lodash.chunk";
+
+const MIGRATION_NAME = "2023-10-05-SetJobRunInternalFlag";
+const JOB_CHUNK_SIZE = 100;
```
hmacr: What should be the ideal chunk size?

ericallam:
I like the idea of doing the backfilling separately, but I think we should try do the backfill in SQL using Prisma migrations. Something like: 

```sql
UPDATE "JobRun"
SET "internal" = "Job"."internal"
FROM "Job"
WHERE "JobRun"."jobId" = "Job"."id";
```

hmacr:
Ah ok. I was worried that a large update would reduce the DB throughput for other processes, so wrote a backfill script.

I'll update the PR with the necessary SQL.

hmacr:
I've removed the backfill script and added the SQL to update the job-run internal flag.

The migration SQL worked as expected on local.

hmacr:
Closing since this code is moved to #563

PR #602, opened 2023-10-11
fix: broken links in documentation

vimode:
## ✅ Checklist

---

## Changelog

Fixed a few broken links. Some seem to have been moved to /documentation and some have their path changed. 

The link(s) in the description of the [Projects](https://trigger.dev/docs/documentation/concepts/projects) page look fine and shouldn't be broken but they are. 
They should resolve to the relative root (/docs/...) but instead they resolve to the domain root. 
I am not sure if its a bug of Mintlify that links in the page title & description or its just how titles work.
So I have updated it to link to the domain root rather than relative root. Let me know if that seems ok.

I also think if it helps, run `broken-links` from [mintlify](https://mintlify.com/docs/settings/broken-links#broken-links) as I wonder how many more broken links might exist as what I could gather the docs seem to have changed a lot.

Thank you.
💯


matt-aitken (approved):
This is great, thanks!

PR #617, opened 2023-10-13
Feat: add `maxDuration` setting for API/Trigger route in the CLI Init Command for Next.js

### LINKED ISSUE #613: [TRI-1398] feat: CLI init should add maxDuration to the api/trigger route for Next.js
On Vercel, even if you're on a paid plan, the max duration of an API route is not what you're paying unless you configure it.

The api/trigger file:

```ts
import { createAppRoute } from "@trigger.dev/nextjs";
import { client } from "@/trigger";

import "@/jobs";

//this route is used to send and receive data with Trigger.dev
export const { POST, dynamic } = createAppRoute(client);

//uncomment this to set a higher max duration (it must be inside your plan limits). Full docs: https://vercel.com/docs/functions/serverless-functions/runtimes#max-duration
//export const maxDuration = 60;
```

There are two different ways to specify the max duration, it depends on which version of Next.js you're on. Info here: https://vercel.com/docs/functions/serverless-functions/runtimes#max-duration

## Tasks
- [ ] For the appropriate Next.js versions add commented out lines to the api/trigger file (as in the above) using CLI init
- [ ] Log out a message to the console using `boxen` with a link to the docs. We do this elsewhere in the CLI. This will cover people who need to uncomment the lines or add it to their vercel.json file

<sub>[TRI-1398](https://linear.app/triggerdotdev/issue/TRI-1398/feat-cli-init-should-add-maxduration-to-the-apitrigger-route-for)</sub>

alexandredev3:
Closes #613 

## ✅ Checklist

---

## Testing
1. Created a project using `Next.js@13.0.3`.
2. Ran `pnpm exec trigger-cli init`. The `src/pages/api/trigger.ts` route was generated without the `maxDuration` commented out.
3. Updated the Next.js dependency version to `13.5.4` within the `pages` directory.
4. Ran `pnpm exec trigger-cli init` again. This time, the `src/pages/api/trigger.ts` route was created with the `maxDuration` commented out.
5. Made adjustments to the project to use the `app` directory.
6. Executed `pnpm exec trigger-cli init` once more. As a result, the `src/pages/api/trigger/route.ts` route was generated with the `maxDuration` commented out.

---

## Changelog

- Added `detectNextVersion` function to detect the consumer's Next.js version.
- Added a new template: `pagesApiRouteWithConfigObject.js` with the `maxDuration` commented out lines.
- The CLI's `init` command detects the consumer's Next.js version and adds `maxDuration` commented out lines to the api/trigger file.
- Log out a message to the console using `boxen` with a link to the vercel runtime `maxDuration` documentation.
- Added `createTriggerRoute` function to create both `AppRoute` and `PagesRoute` file. Both functions, `createTriggerAppRoute` and `createTriggerPageRoute`, does the same thing, with the only distinction being the `path` for the generated trigger route file path. The caller of `createTriggerRoute` is responsible for providing the path for the generated trigger route file.

---


packages/cli/src/frameworks/nextjs/index.ts
```diff
@@ -144,45 +174,34 @@ export async function detectPagesOrAppDir(path: string): Promise<"pages" | "app"
...
+  const packageJsonContent = await readPackageJson(path);
+  if (!packageJsonContent) {
+    return null;
   }
-  logger.success(`✔ Created API route at ${apiRoutePath}`);
 
-  await createJobsAndTriggerFile(path, endpointSlug, fileExtension, pathAlias, templatesDir);
+  const versionNumberPattern = /[\d.]+|latest/;
+  if (packageJsonContent.dependencies?.next !== undefined)
+    return packageJsonContent.dependencies?.next?.match(versionNumberPattern)![0];
```
matt-aitken: There is a risk here that the user has a weird value for their version and we get null. Then this will throw an exception.

I'm going to change this to this:

```suggestion
if (packageJsonContent.dependencies?.next !== undefined)
    return packageJsonContent.dependencies?.next?.match(versionNumberPattern)?.at(0) ?? null;
  if (packageJsonContent.devDependencies?.next !== undefined)
    return packageJsonContent.devDependencies?.next?.match(versionNumberPattern)?.at(0) ?? null;
```

matt-aitken (approved):
Great! I made the version checking a bit safer but other than this was spot on

PR #628, opened 2023-10-16
fix: append nanoid for dynamic trigger reg event

hmacr:
## ✅ Checklist

---

## Testing

Verified Dynamic Trigger registration & re-registration, ensured that a single event dispatcher is created for both cases.

---

## Changelog

Append a `nanoid` to uniquely identify the dynamic trigger source registration event.

---

💯

### DIFF
--- a/.changeset/strange-schools-boil.md
+++ b/.changeset/strange-schools-boil.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/sdk": patch
+---
+
+use idempotency-key as event-id for dynamic-trigger registrations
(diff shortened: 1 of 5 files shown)


ericallam:
This is a good idea, but I think we should use the value of an "Idempotency-Key" header by default as the eventId, with the registration.id + nanoid as a fallback if there is no Idempotency-Key header. Then we should update the `registerInterval` function to pass in the `task.idempotencyKey`:

https://github.com/triggerdotdev/trigger.dev/blob/21f62e780c4fef4a6c19d243762e99aed5be56b6/packages/trigger-sdk/src/io.ts#L535:

```ts
const registration = await this.runTask(
          "register-source",
          async (subtask1) => {
            return trigger.register(id, params, subtask1.idempotencyKey);
          },
          {
            name: "register-source",
          }
        );
```

hmacr:
> This is a good idea, but I think we should use the value of an "Idempotency-Key" header by default as the eventId, with the registration.id + nanoid as a fallback if there is no Idempotency-Key header. Then we should update the `registerInterval` function to pass in the `task.idempotencyKey`:

@ericallam The deprecated `io.registerTrigger` internally invokes the `DynamicTrigger.register` method. So, I've added the support to pass the task idempotency-key (when available) to the registration endpoint. Please check.

PR #680, opened 2023-10-23
Add notice on `sideEffects` in deployment guide

zcesur:
## ✅ Checklist

---

## Changelog

Added an extra step prompting users to include `./src/jobs/**/*.ts` in `sideEffects` if needed

I might have an unconventional monorepo setup since noone else seems to have this issue :man_shrugging: But only after doing this did I manage to convince Next.js to not tree shake away my jobs

---


ericallam:
Interesting... I wonder if you could share a little more of your file structure and contents so we can track down why this wasn't working. Could you share the contents of the trigger API route file?

zcesur:
Here ya go:
```ts
// apps/console/src/pages/api/trigger.ts
import { createPagesRoute } from "@trigger.dev/nextjs";

import { client } from "~/jobs";

const { handler, config } = createPagesRoute(client);
export { config };

export default handler;


// apps/console/src/jobs/client.ts
import { Trigger } from "@algora/trigger";
export const client = Trigger.init();


// packages/trigger/src/client.ts
import { TriggerClient } from "@trigger.dev/sdk";
// ...
export const init = () =>
  new TriggerClient({
    id: id,
    apiKey: env.TRIGGER_API_KEY,
    apiUrl: env.TRIGGER_API_URL,
  });
```

I had also made a dummy project as I was trying to track this down. Just pushed it on Github if you wanna reproduce the issue:
https://github.com/zcesur/trigger-sideeffects-repro

matt-aitken (approved):

PR #687, opened 2023-10-26
fix : Only allow logins from whitelisted_emails (for self-hosting)

### LINKED ISSUE #685: [TRI-1444] Only allow logins from whitelisted_emails (for self-hosting)
Currently there isn't a convenient way to lock down logins to a self-hosted Trigger.dev service.

We should add a new ENV var which allows a regex pattern to be used to only allow certain emails.

```
WHITELISTED_EMAILS="matt@gmail\.com|jane@yahoo\.com"
```

Only emails that match this would be able to sign-up or login.

It must be added to `env.server.ts` file, and it needs to be optional. Probably the easiest way to implement this would be to modify the `findOrCreateUser()` function in `user.server.ts` to first check the whitelist if it exists and just throw an error.

<sub>[TRI-1444](https://linear.app/triggerdotdev/issue/TRI-1444/only-allow-logins-from-whitelisted-emails-for-self-hosting)</sub>

HarshitVashisht11:
Closes #685 

## ✅ Checklist

---

## Testing

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,6 +24,11 @@ NODE_ENV=development
 # FROM_EMAIL=
 # REPLY_TO_EMAIL=
 
+# OPTIONAL VARIABLES
+WHITELISTED_EMAILS="matt@gmail\.com|jane@yahoo\.com"
+# Description: This environment variable defines a regex pattern for allowed email addresses. 
+# Only emails that match this pattern will be able to sign up or log in.
+
 # CLOUD VARIABLES
 POSTHOG_PROJECT_KEY=
 PLAIN_API_KEY= 
--- a/apps/webapp/app/models/user.server.ts
+++ b/apps/webapp/app/models/user.server.ts
@@ -2,7 +2,11 @@ import type { Prisma, User } from "@trigger.dev/database";
 import type { GitHubProfile } from "remix-auth-github";
 import { prisma } from "~/db.server";
 import { env } from "~/env.server";
+import { authenticator } from "~/services/auth.server";
+import { addEmailLinkStrategy } from "~/services/emailAuth.server";
 export type { User } from "@trigger.dev/database";
+
+addEmailLinkStrategy(authenticator);
 
 type FindOrCreateMagicLink = {
   authenticationMethod: "MAGIC_LINK";
@@ -23,16 +27,11 @@ type LoggedInUser = {
   isNewUser: boolean;
 };
 
+class EmailWhitelistError extends Error {}
+
 export async function findOrCreateUser(input: FindOrCreateUser): Promise<LoggedInUser> {
-  if (env.WHITELISTED_EMAILS) {
-    // Create a regular expression from the whitelist pattern
-    const emailWhitelistRegex = new RegExp(env.WHITELISTED_EMAILS);
-    
-    // Check if the user's email matches the whitelist pattern
-    if (!emailWhitelistRegex.test(input.email)) {
-      // If the email is not in the whitelist, throw an error
-      throw new Error('Email address is not allowed.');
-    }
+  if (!isEmailWhitelisted(input.email, env.WHITELISTED_EMAILS)) {
+    throw new EmailWhitelistError("Access to this instance is restricted.");
   }
   switch (input.authenticationMethod) {
     case "GITHUB": {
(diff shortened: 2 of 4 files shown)


HarshitVashisht11:
Hi @matt-aitken can you please review my PR

HarshitVashisht11:
Hi @matt-aitken, can you please review my PR

nicktrn:
Hi @HarshitVashisht11, thanks for your submission!

A few pointers:
- Simple testing instructions with sample values would have been great! And maybe a screenshot as this affects UX.
- The env var could do with being added to `.env.example` under `# OPTIONAL VARIABLES`. Commented out and with a short description of what it does and what values it accepts.
- The error text should probably be sanitized. Did you try to pass an invalid regex pattern? Maybe we could log the error so server admins can see the real issue. Users could be shown a more generic message, something like: "Access to this instance is restricted."
- The UX for this isn't great. We should display an error immediately instead of going back to the login page first.

[email-whitelist.webm](https://github.com/triggerdotdev/trigger.dev/assets/55853254/d507f677-ef25-44de-9a8b-3aca0e18e1ee)

HarshitVashisht11:
Thanks for reviewing my PR I will try to fix the issue and will provide a  working video as soon as possible

nicktrn:
Just a small hint. We use [remix-auth](https://github.com/sergiodxa/remix-auth) - maybe there's a better way to handle those errors.

HarshitVashisht11:
https://github.com/triggerdotdev/trigger.dev/assets/120767685/cb6ac059-e06e-4509-96c2-732a5ff05deb

Please Review my PR @nicktrn

.env.example
```diff
@@ -24,6 +24,11 @@ NODE_ENV=development
 # FROM_EMAIL=
 # REPLY_TO_EMAIL=
 
+# OPTIONAL VARIABLES
+WHITELISTED_EMAILS="matt@gmail\.com|jane@yahoo\.com"
+# Description: This environment variable defines a regex pattern for allowed email addresses. 
+# Only emails that match this pattern will be able to sign up or log in.
+
```
nicktrn, line 31: Please have a look further up the file, the Optional Variables section already exists.

The `WHITELISTED_EMAILS` line would have to be commented out as it's optional.

The description is good, but would have to go above the env var, not below. You can also drop the `Description:` prefix.

apps/webapp/app/models/user.server.ts
```diff
@@ -2,8 +2,12 @@ import type { Prisma, User } from "@trigger.dev/database";
 import type { GitHubProfile } from "remix-auth-github";
 import { prisma } from "~/db.server";
 import { env } from "~/env.server";
+import { authenticator } from "~/services/auth.server";
+import { addEmailLinkStrategy } from "~/services/emailAuth.server";
 export type { User } from "@trigger.dev/database";
 
+addEmailLinkStrategy(authenticator);
```
nicktrn, line 9: Why? We already do this elsewhere.

apps/webapp/app/models/user.server.ts
```diff
@@ -189,3 +188,11 @@ export async function grantUserCloudAccess({ id, inviteCode }: { id: string; inv
     },
   });
 }
+
+function isEmailWhitelisted(email: string, whitelist: string | undefined) {
+  if (whitelist) {
+    const regex = new RegExp(whitelist);
```
nicktrn, line 194: What happens if whitelist contains an invalid regular expression?

apps/webapp/app/routes/magic.tsx
```diff
@@ -7,6 +7,6 @@ export async function loader({ request }: LoaderArgs) {
 
   await authenticator.authenticate("email-link", request, {
     successRedirect: redirectTo ?? "/",
-    failureRedirect: "/login",
+    failureRedirect: "/login/magic",
```
nicktrn, line 10: This seems to make sense. ~~We should probably do this for the GitHub strategy as well.~~

@matt-aitken Can you see any issues? Was there a particular reason we'd want to always redirect to `/login` on errors instead of staying on the current strategy-specific page? (There's a back button on `/login/magic` for example.)

If we don't do this, user-facing errors won't actually be displayed until we navigate to `/login/magic` again.

nicktrn (commented):
Thanks for those edits!

Makes sense to abstract the regex check into a separate function.

I'm wondering if `findOrCreateUser` is actually the right place for this. We may just want to call the whitelist check in each strategy.

Feel free to experiment to see what works best, and what potential issues arise with each option.

nicktrn:
Thanks for your efforts @HarshitVashisht11! We decided to go with #721 in the end.

PR #692, opened 2023-10-27
Fix: [TRI-1444] Only allow logins from whitelisted_emails (for self-hosting)#685

AsbarK:
Closes #<issue>

## ✅ Checklist

## Steps

1. **Setting `WHITELISTED_EMAILS` in `.env.example`**:
   - I added the `WHITELISTED_EMAILS` variable to `.env.example` under `# OPTIONAL VARIABLES` with the value `matt@gmail\.com|jane@yahoo\.com`.
   - I ensured that the application reads this environment variable correctly.

## Testing

I have tested the changes in this pull request as follows:
1. ** Values to check wether the `WHITELISTED_EMAILS` validation works 
   - `matt@gmail.com` or `jane@yahoo.com`

## Screenshots

https://github.com/triggerdotdev/trigger.dev/assets/106991686/717704f3-5d3f-4fc1-9458-072b0614fe5c

💯


AsbarK:
hey @nicktrn  should i change the error message to some thing like `Access Denied!`

apps/webapp/app/models/user.server.ts
```diff
@@ -36,6 +37,33 @@ export async function findOrCreateUser(input: FindOrCreateUser): Promise<LoggedI
...
+          email: input.email,
+        },
+        update: { email: input.email },
+        create: { email: input.email, authenticationMethod: "MAGIC_LINK" },
+      });
+
+      return {
+        user,
+        isNewUser: !existingUser,
+      };
```
nicktrn, line 60: No need to duplicate all that code. Better to negate your if condition.

apps/webapp/app/routes/magic.tsx
```diff
@@ -7,6 +7,6 @@ export async function loader({ request }: LoaderFunctionArgs) {
 
   await authenticator.authenticate("email-link", request, {
     successRedirect: redirectTo ?? "/",
-    failureRedirect: "/login",
+    failureRedirect: "/login/magic",
```
nicktrn, line 10: Haven't tested this, but should work. What about any other auth strategies?

.env.example
```diff
@@ -13,6 +13,8 @@ APP_ORIGIN=http://localhost:3030
 NODE_ENV=development
 
 # OPTIONAL VARIABLES
+# This is used for Restricting logins other than whitelisted emails for self-hosted authentication.
+WHITELISTED_EMAILS="matt@gmail\.com|jane@yahoo\.com"
```
nicktrn, line 17: This is optional and should be commented out.

nicktrn (commented):
Hi @AsbarK, thanks for your submission. Have left some comments. No need for double mentions in both the PR and Issue next time.

nicktrn:
> should i change the error message to some thing like Access Denied

We can change the error message later, I think what's more important is to check what happens when you pass an invalid RegEx string in that env var.

AsbarK:
thanks for reviewing  i will impliment the changes and let you know.

apps/webapp/app/routes/magic.tsx
AsbarK: hey @nicktrn , For GitHub when someone clicks the continue with GitHub button, the button should only allow the users if the GitHub  email matches the regs?

nicktrn: Actually, leave it like this. If people want to restrict access via email whitelist, they probably shouldn't enable GitHub auth.

PR #699, opened 2023-10-28
[TRI-1449] Display warning message when duplicate job IDs are detected

### LINKED ISSUE #691: [TRI-1449] Display warning message when duplicate job IDs are detected
It's quite easy to accidentally reuse a job ID:

```ts
// foo.ts
client.defineJob({
  id: "do-something",
});

// bar.ts
client.defineJob({
  id: "do-something",
})
```

We should detect when this happens and display a warning in the users local server console.

<sub>[TRI-1449](https://linear.app/triggerdotdev/issue/TRI-1449/display-warning-message-when-duplicate-job-ids-are-detected)</sub>

alexandredev3:
Closes #691 

## ✅ Checklist

---

## Testing

- Tried to define two jobs with identical IDs:
  ```ts
    client.defineJob({
      id: "same-job-id",
    });

    client.defineJob({
      id: "same-job-id",
    });
  ``` 
Got this warning message printed out to the console: `Warning: The Job "same-job-id" you're attempting to define has already been defined. Please assign a different ID to the job.`

---

## Changelog

- Added the `colorette` library; couldn't use `chalk` because it's a pure ESM package while `trigger-sdk` is being compiled to CJS.

---

💯


packages/trigger-sdk/src/triggerClient.ts
```diff
@@ -464,10 +465,17 @@ export class TriggerClient {
     };
   }
 
-  defineJob<
+  async defineJob<
```
ericallam: This shouldn't be an async method

packages/trigger-sdk/src/triggerClient.ts
```diff
@@ -464,10 +465,17 @@ export class TriggerClient {
...
     TTrigger extends Trigger<EventSpecification<any>>,
     TIntegrations extends Record<string, TriggerIntegration> = {},
   >(options: JobOptions<TTrigger, TIntegrations>) {
+    const existingRegisteredJob = this.#registeredJobs[options.id];
+    if (existingRegisteredJob) {
+      console.warn(
```
ericallam, line 488: This looks good, but I think it should be something like what you have but with a bit that makes it clear the warning is from Trigger.dev: `[@trigger.dev/sdk] Warning: the Job ...`

ericallam (changes requested):

alexandredev3:
> This shouldn't be an async method
> This looks good, but I think it should be something like what you have but with a bit that makes it clear the warning is from `Trigger.dev: [@trigger.dev/sdk] Warning: the Job ...`

@ericallam Done!

ericallam (approved):

PR #708, opened 2023-10-30
feat: view teammates' dev env job runs

### LINKED ISSUE #435: [TRI-1226] Allow seeing teammates' dev runs
When on the runs page for a job you currently only see runs associated with your personal Dev environment – you DO NOT see runs from your teammates Dev environment.

## Options

1. Keep it as-is, but add a filter dropdown where you can select environments. Teammate Dev environments would be unchecked by default when you load the page.
2. Show all Dev environments by default. Filters can be added later.

I think option 2 is the correct choice.

## Display style

Your own Dev environment would remain as-is: `[DEV]`.
Teammate's would appear like this: `[DEV: Amy]`.

## Changes required

* Add an optional name param on the `EnvironmentLabel`.
* Change the `RunListPresenter` so it doesn't filter the environments in the where clause. Get the name of the user for the environment.
* The RunTable would need to be adapted to pass in the name (note this only applies to DEV environments).
* On the Run page, update the environment badge

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1226](https://linear.app/triggerdotdev/issue/TRI-1226/allow-seeing-teammates-dev-runs)</sub>

hmacr:
Closes #435 

## ✅ Checklist

---

## Testing

Verified the environment labels in the dashboard.

---

## Changelog

- Remove env-filter when fetching run-list.
- Add optional `username` in `EnvironmentLabel`.
- Determine `username` to display based on current-user in `RunListPresenter` and `RunPresenter`.

---

## Screenshots

**Current User - Run List**
Screenshot 2023-10-30 at 8 56 37 PM
<br>

**Other User - Run List**
Screenshot 2023-10-30 at 8 56 13 PM
<br>

**Current User - Run**
Screenshot 2023-10-30 at 8 56 44 PM
<br>

**Other User - Run**
Screenshot 2023-10-30 at 8 56 27 PM

💯


apps/webapp/app/presenters/RunListPresenter.server.ts
```diff
@@ -138,4 +133,27 @@ export class RunListPresenter {
       },
     };
   }
+
+  // we want to remove `null` from username
+  #getUsername(user?: {
```
ericallam: I think this is repeated in RunPresenter, it would be good to extract this out into a single function. Also, is there any backup in case this #getUsername returns undefined?

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.triggers_.external.$triggerParam_.runs.$runParam/route.tsx
```diff
@@ -124,6 +126,7 @@ export default function Page() {
           id: trigger.id,
         }),
       }}
+      currUser={user}
```
ericallam: We prefer to write out the entire word, so can you rename this to `currentUser`?

ericallam (dismissed):

apps/webapp/app/presenters/RunListPresenter.server.ts
hmacr: `#getUsername` shouldn't return `null` because all the registered users will have a name set.
Either way, if it does, the username will not be appended to the env-label in RunList and Run pages (Dev: _username_ vs Dev) which should be fine I hope.

hmacr:
@ericallam I've addressed the review comments, please check.

matt-aitken (approved):

matt-aitken (approved):
This is great @hmacr!

PR #712, opened 2023-10-31
fix: set auth-method for OAuth integration clients

hmacr:
## Issue
The additional tabs (Connections, and Scopes) for OAuth connections were not displayed.
This is because the nullish coalescing operator takes precedence over ternary operator. So, the auth-method type and name were not set properly when the auth-method is OAuth.

Operator Precedence Explanation: [Stack Overflow](https://stackoverflow.com/a/67575538/21819000)

## ✅ Checklist

---

## Testing

Verified that the additional tabs are displayed for OAuth integrations

---

## Changelog

- Wrap the auth-source ternary check in parenthesis to increase operator precedence

---

## Screenshots

**Before**
<br>
Screenshot 2023-10-31 at 12 56 38 PM
<br>
**After**
<br>
Screenshot 2023-10-31 at 12 54 41 PM

💯

### DIFF
--- a/apps/webapp/app/presenters/IntegrationClientPresenter.server.ts
+++ b/apps/webapp/app/presenters/IntegrationClientPresenter.server.ts
@@ -121,11 +121,11 @@ export class IntegrationClientPresenter {
       },
       authMethod: {
         type:
-          integration.authMethod?.type ?? integration.authSource === "RESOLVER" ? "local" : "local",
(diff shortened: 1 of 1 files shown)


matt-aitken:
This is great @hmacr – I've manually incorporated these changes into main because we made quite a lot of other changes so the merge was painful

PR #720, opened 2023-11-03
HTTP endpoints, HTTP triggers and new sidebar

matt-aitken:
This is a monster Pull Request because it includes the new app sidebar as well as the HTTP endpoint/trigger work.

## HTTP endpoints & HTTP triggers
Sometimes you want to subscribe to changes from an API, and we don't have an Integration for it yet. That's when you can use `defineHttpEndpoint` to receive webhooks, verify them, and create an HTTP Trigger.

```ts
//create an HTTP endpoint
const caldotcom = client.defineHttpEndpoint({
  id: "cal.com",
  source: "cal.com",
  icon: "caldotcom",
  verify: async (request) => {
    //this helper function makes verifying most webhooks easy
    return await verifyRequestSignature({
      request,
      headerName: "X-Cal-Signature-256",
      secret: process.env.CALDOTCOM_SECRET!,
      algorithm: "sha256",
    });
  },
});

client.defineJob({
  id: "http-caldotcom",
  name: "HTTP Cal.com",
  version: "1.0.0",
  enabled: true,
  //create a Trigger from the HTTP endpoint above. The filter is optional.
  trigger: caldotcom.onRequest({ filter: { body: { triggerEvent: ["BOOKING_CANCELLED"] } } }),
  run: async (request, io, ctx) => {
    //note that when using HTTP endpoints, the first parameter is the request
    //you need to get the body, usually it will be json so you do:
    const body = await request.json();
    await io.logger.info(`Body`, body);
  },
});
```

There's new sections of the dashboard that relate to this, as well as some new abilities like being able to link to Triggers from Job pages – currently only used by the HTTP trigger.

## New sidebar
This change has a massive blast radius because it involved moving and renaming quite a lot of files, which then impacts the related imports. Generally it's simpler now because we ditched the dropdown breadcrumb menus, which had some nasty queries in weird routes.

## Todo
Docs
- [x] Merge main in
- [x] Add top-level concept for HTTP Endpoint
- [x] Add trigger
- [x] References
  - [x] TriggerClient define
  - [x] TriggerHttpEndpoint


nicktrn:
Haven't had  the pleasure to look at this Lovecraftian work in depth yet, but wanted to suggest we squash those Prisma migrations before merging. I recall a recent issue when we didn't.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam/route.tsx
```diff
@@ -35,9 +36,16 @@ export const loader = async ({ request, params }: LoaderFunctionArgs) => {
 
     telemetry.project.identify({ project });
 
-    return typedjson({
-      project,
-    });
+    const session = await setCurrentProjectId(project.id, request);
```
matt-aitken, line 39: This is where the current projectId is set, so the new sidebar can always show a project

apps/webapp/app/routes/_app.orgs.$organizationSlug/route.tsx
```diff
@@ -1,63 +1,95 @@
...
+    useTypedLoaderData<typeof loader>();
+  const user = useUser();
+
+  //the side menu won't change projects when using the switcher unless we use the hook (on project pages)
+  const project = useOptionalProject() ?? currentProject;
+
   return (
     <>
-      <Outlet />
+      <div className="grid grid-cols-[14rem_1fr] overflow-hidden">
```
matt-aitken, line 69: The new SideMenu lives at the Org level. Plus the breadcrumb and the new loading indicator.

apps/webapp/app/routes/api.v1.httpendpoints.$httpEndpointId.env.$envType.$shortcode.ts
```diff
@@ -0,0 +1,56 @@
+import type { ActionFunctionArgs, LoaderFunctionArgs } from "@remix-run/server-runtime";
+import {
+  HandleHttpEndpointService,
+  HttpEndpointParamsSchema,
+} from "~/services/httpendpoint/HandleHttpEndpointService";
+import { logger } from "~/services/logger.server";
+
+export async function action({ request, params }: ActionFunctionArgs) {
```
matt-aitken: Webhook data is received here for HTTP endpoints.

apps/webapp/app/services/endpointApi.server.ts
```diff
@@ -240,6 +241,36 @@ export class EndpointApi {
     return HttpSourceResponseSchema.parse(anyBody);
   }
 
+  async deliverHttpEndpointRequestForResponse(options: {
```
matt-aitken, line 244: This is used when the user wants an immediate `Response` to their `Request` (by implementing `responseWith` on their HTTP endpoint.

apps/webapp/app/services/triggers/registerHttpEndpoint.server.ts
```diff
@@ -0,0 +1,112 @@
+import { HttpEndpointMetadata } from "@trigger.dev/core";
+import { z } from "zod";
+import { $transaction, Prisma, PrismaClientOrTransaction, prisma } from "~/db.server";
+import { ExtendedEndpoint, findEndpoint } from "~/models/endpoint.server";
+import { getSecretStore } from "../secrets/secretStore.server";
+import { generateSecret } from "../sources/utils.server";
+import { logger } from "../logger.server";
+
+export class RegisterHttpEndpointService {
```
matt-aitken, line 9: Where the HttpEndpoint gets indexed

apps/webapp/app/services/httpendpoint/HandleHttpEndpointService.ts
```diff
@@ -0,0 +1,241 @@
...
+
+export class HandleHttpEndpointService {
+  #prismaClient: PrismaClient;
+
+  constructor(prismaClient: PrismaClient = prisma) {
+    this.#prismaClient = prismaClient;
+  }
+
+  public async call(params: HttpEndpointParams, request: Request) {
+    const httpEndpoint = await this.#prismaClient.triggerHttpEndpoint.findUnique({
```
matt-aitken, line 31: Where the Http endpoint requests get processed

apps/webapp/app/utils/createHttpSourceRequest.ts
```diff
@@ -0,0 +1,14 @@
+import { HttpSourceRequest } from "@trigger.dev/core";
+import { requestUrl } from "./requestUrl.server";
+
+export async function createHttpSourceRequest(request: Request): Promise<HttpSourceRequest> {
+  const url = requestUrl(request);
+  return {
+    headers: Object.fromEntries(request.headers) as Record<string, string>,
+    url: url.href,
+    method: request.method,
+    rawBody: ["POST", "PUT", "PATCH"].includes(request.method)
```
matt-aitken: This was taken from the HttpSource code. Is this the best way to deal with a body? 

In theory a `GET` can have a body… also it's possible that a `POST` doesn't have a body, in which case this would throw an error I think. What about `Delete`?

What if instead we just try catch request.arrayBuffer() and if there's data we put it in the `rawBody`? Feels more comprehensive?

@ericallam @nicktrn

apps/webapp/tailwind.config.js
```diff
@@ -87,6 +87,14 @@ module.exports = {
             fontWeight: "500",
           },
         ],
+        "2sm": [
```
matt-aitken, line 90: @samejr I can't tell from the name of this how big it is in relation to the other sizes… naming is hard, but I think we can do better than this.

apps/webapp/app/utils/createHttpSourceRequest.ts
nicktrn: > What if instead we just try catch request.arrayBuffer() and if there's data we put it in the rawBody?

I think this may have to be done to cover all edge cases. We shouldn't count on external services being sensible.

apps/webapp/app/services/triggers/registerHttpEndpoint.server.ts
```diff
@@ -0,0 +1,112 @@
...
+          endpointId: endpoint.id,
+          httpEndpointId: httpEndpoint.id,
+          skipTriggeringRuns: httpEndpointMetadata.skipTriggeringRuns,
+          source: httpEndpointMetadata.source,
+        },
+      });
+
+      //create the secret
+      //we don't upsert because we don't want to change an existing one
+      const secretStore = getSecretStore(httpEndpoint.secretReference.provider);
```
ericallam: the transaction should be used here as well

packages/trigger-sdk/src/security.ts
```diff
@@ -0,0 +1,41 @@
+import crypto from "crypto";
+import { VerifyResult } from "./types";
+
+export async function verifyRequestSignature({
```
ericallam, line 5: Add jsdocs and add to the docs

packages/trigger-sdk/src/triggerClient.ts
```diff
@@ -515,9 +570,20 @@ export class TriggerClient {
...
   }
 
+  /**
+   * An [HTTP endpoint](https://trigger.dev/docs/documentation/concepts/http-endpoints) allows you to create a [HTTP Trigger](https://trigger.dev/docs/documentation/concepts/triggers/http), which means you can trigger your Jobs from any webhooks.
+   * @param options The Endpoint options
+   * @returns An HTTP Endpoint, that can be used to create an HTTP Trigger.
+   * @link https://trigger.dev/docs/documentation/concepts/http-endpoints
+   */
+  defineHttpEndpoint(options: EndpointOptions) {
+    const endpoint = httpEndpoint(options);
```
ericallam, line 589: Check for duplicate IDs

references/job-catalog/src/httpTrigger.ts
```diff
@@ -0,0 +1,159 @@
...
+    } catch (error) {
+      return {
+        success: false,
+        reason: error instanceof Error ? error.message : JSON.stringify(error),
+      };
+    }
+  },
+});
+
+client.defineJob({
```
ericallam, line 132: test using the same endpoint on multiple jobs

ericallam (approved):

apps/webapp/app/utils/createHttpSourceRequest.ts
matt-aitken: Ok this is now done

references/job-catalog/src/httpTrigger.ts
matt-aitken: Yep this is working, added an example to the catalog

PR #721, opened 2023-11-03
FIX: [TRI-1444] Allow only logins from WHITELISTED_EMAILS fixes #685

### LINKED ISSUE #685: [TRI-1444] Only allow logins from whitelisted_emails (for self-hosting)

abelghalem:
Closes #685 

## ✅ Checklist

---

## Testing

1. Set `WHITELISTED_EMAILS` in .env with a regex with allowed emails.
2. Start the webapp with the command `pnpm run dev --filter webapp`.
3. Trying to log in with an unauthorized email should show an error.
4. Trying to log in with an authorized email should log in or create the account.

---

## Changelog

User can now set a WHITELISTED_EMAILS environment variable that must be a valid Regex to list the authorized emails.

---

## Screenshots

image

💯

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,8 @@ APP_ORIGIN=http://localhost:3030
 NODE_ENV=development
 
 # OPTIONAL VARIABLES
+# This is used for validating emails that are allowed to log in. Every email that do not match this regex will be rejected.
+# WHITELISTED_EMAILS="authorized@yahoo\.com|authorized@gmail\.com"
 # This is used for logging in via GitHub. You can leave these commented out if you don't want to use GitHub for authentication.
 # AUTH_GITHUB_CLIENT_ID=
 # AUTH_GITHUB_CLIENT_SECRET=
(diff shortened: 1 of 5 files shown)


matt-aitken (approved):
This is great!

PR #732, opened 2023-11-10
Fix for #731, integrations getting stuck if OAuth then moved to API key

### LINKED ISSUE #731: [TRI-1575] Creating an integration with OAuth then switching to API Key makes it stuck
## To reproduce
1. Create a SendGrid integration with just an id (no API key) and reference it in a Job.
2. Save and check it's appeared in the dashboard
3. You'll notice that on the integrations page it says it requires action. Don't do anything there.
4. Then go back to your code and add `apiKey: process.env.SENDGRID_API_KEY` to the definition.
5. Save so it syncs

When you go back to the dashboard and refresh, the integration is in the errored state still and is stuck. When you click it you go to a blank panel.

## Solution
We need to set the `setupStatus` to `"COMPLETE"` when upserting LOCAL or AUTH_RESOLVER integration types.

<sub>[TRI-1575](https://linear.app/triggerdotdev/issue/TRI-1575/creating-an-integration-with-oauth-then-switching-to-api-key-makes-it)</sub>

matt-aitken:
Closes #731 

## ✅ Checklist

---

## Testing

- [ ] Create an integration that's OAuth (like SendGrid)
- [ ] Sync it to the dashboard
- [ ] Check it's in an incomplete state on the integrations page
- [ ] In your code add an `apiKey`
- [ ] Save and sync
- [ ] The integration should now be setup as "LOCAL"

### DIFF
--- a/apps/webapp/app/services/jobs/registerJob.server.ts
+++ b/apps/webapp/app/services/jobs/registerJob.server.ts
@@ -446,6 +446,7 @@ export class RegisterJobService {
             title: jobIntegration.metadata.name,
             authSource: "LOCAL",
             connectionType: "DEVELOPER",
+            setupStatus: "COMPLETE",
             definition: {
               connectOrCreate: {
                 where: {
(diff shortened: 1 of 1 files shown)


matt-aitken:
This needs testing

Issue #769, opened 2023-12-02
Shopify webhook  - Failed sha256 verification


jaaneh:
I opened this originally in discussions, but unsure if this is better suited as an issue so re-opening it here. Feel free to close one and continue conversation in the other. Discussion https://github.com/triggerdotdev/trigger.dev/discussions/768

---

Hey there,

I've been trying to get up an going with the Shopify integration, however every time I trigger an event through Shopify it leads to a "Failed sha256 verification" error.

![image](https://github.com/triggerdotdev/trigger.dev/assets/27323317/d5924c04-9cd4-4822-a9a4-961f20e54fa2)

The webhook succeeds obviously when using the test event in Trigger, and it returns the payload (using the example below). I have set up webhook in Shopify, and I am also using the correct secrets/env vars.

I am self-hosting Trigger, and have input the DEV URL to Shopify as the app is not hosted yet.

Seeing as Shopify doesn't have a way to insert the secret that Trigger gives you, and they instead give us one, I assume Trigger registers this automatically when registering the trigger? Still, with the correct URL set in Shopify it fails to verify.

Here's an example piece of code that doesn't work:

```ts
import "@shopify/shopify-api/adapters/node"
import { Shopify } from "@trigger.dev/shopify"
import { client } from "../"

const shopify = new Shopify({
  id: "shopify",
  apiKey: process.env.SHOPIFY_API_KEY!,
  apiSecretKey: process.env.SHOPIFY_API_SECRET_KEY!,
  adminAccessToken: process.env.SHOPIFY_ADMIN_ACCESS_TOKEN!,
  hostName: process.env.SHOPIFY_SHOP_DOMAIN!
})

client.defineJob({
  id: "shopify/on-order-created",
  name: "Shopify: On Order Created",
  version: "0.1.0",
  trigger: shopify.on("orders/create"),
  run: async (payload, io, ctx) => {
    return { payload }
  }
})
```

jaaneh:
Hi @nicktrn. Appreciate the reply.

I am running `2.2.10` on the SDK, Express and Shopify packages. The self-hosted is also on the latest `2.2.17`.

I've made sure the `apiSecretKey` is correct and matches. Please see below:

![image](https://github.com/triggerdotdev/trigger.dev/assets/27323317/21d8aff5-6789-467a-b5ce-6a3de8e90e42)
[...]

PR #773, opened 2023-12-05
docs: Add comment to when using as Express middleware

jaaneh:
Related #769 

## ✅ Checklist

---

## Changelog

Add a comment explaining that for Express, any JSON parser middlewares need to be registered after the Trigger middleware.

---

## Screenshots

image

💯

### DIFF
--- a/docs/_snippets/manual-setup-express.mdx
+++ b/docs/_snippets/manual-setup-express.mdx
@@ -81,6 +81,7 @@ const app: Express = express();
 
 //add the middleware
 app.use(createMiddleware(client));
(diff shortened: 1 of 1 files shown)


jaaneh:
Actually, this may only be related to certain integrations. Perhaps it's easier to document this on those integrations pages rather than the main Express platform page? Not 100% sure if it's easier to just require it to be above for all for ease, or if some integrations actually require it to come after.

matt-aitken (approved):
This is a really good clarification, thanks @jaaneh

PR #791, opened 2023-12-12
Auto update endpoint URL on redirect

### LINKED ISSUE #786: [TRI-1739] Vercel 308 redirect causes endpoint to not be reachable
When prod/staging endpoints are configured with a root domain (e.g. `https://example.com/api/trigger`), and a redirect is configured in vercel to redirect `example.com` -> `www.example.com`, we fail to follow the redirect and are unable to connect to the endpoint. We should handle this situation and make sure redirects are followed (and better yet, update the endpoint to the correct URL, since the redirect is permanent (308)

<sub>[TRI-1739](https://linear.app/triggerdotdev/issue/TRI-1739/vercel-308-redirect-causes-endpoint-to-not-be-reachable)</sub>

ericallam:
Closes #786

## ✅ Checklist

---

## Testing

Tested against a URL redirecting using a 308 on Vercel (root -> www) and made sure the endpoint URL was updated and the indexing completed successfully.

---

## Changelog

We will now detect redirects when doing an endpoint indexing request and update the endpoint URL to "follow" the redirect, and then retry the indexing. If that happens more than 5 times, we will abort. If the response is a redirect with no location URL, we will abort.

### DIFF
--- a/apps/webapp/app/services/endpointApi.server.ts
+++ b/apps/webapp/app/services/endpointApi.server.ts
@@ -123,6 +123,7 @@ export class EndpointApi {
         "x-trigger-api-key": this.apiKey,
         "x-trigger-action": "INDEX_ENDPOINT",
       },
+      redirect: "manual",
     });
 
     return {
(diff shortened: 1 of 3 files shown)


nicktrn (approved):
Nice and simple :+1:

PR #800, opened 2023-12-19
feat:runs filtering by status and environment

### LINKED ISSUE #520: [TRI-1320] [TRI-1105] Ability to filter runs by status and environment
### Problem

The runs list can get very long and there's no way to quickly find a run by status (success/fail/etc) and the environment (dev/staging/prod).

### Solution

Add filtering for

* Status
* Environment (include other users' dev environments in the list)

<sub>From [SyncLinear.com](https://synclinear.com) | [TRI-1105](https://linear.app/triggerdotdev/issue/TRI-1105/ability-to-filter-runs-by-status-and-environment)</sub>

<sub>[TRI-1320](https://linear.app/triggerdotdev/issue/TRI-1320/[tri-1105]-ability-to-filter-runs-by-status-and-environment)</sub>

Abhi1992002:
Closes #520 

## ✅ Checklist

---

## Changelog

Add 2 filters , environment and status 

---

## Video

_[video]_

https://github.com/triggerdotdev/trigger.dev/assets/122007096/bc1b863e-e00d-4d43-a986-40965869fb67

💯


apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.runs/route.tsx
```diff
@@ -48,6 +110,70 @@ export default function Page() {
...
+      setselectedStatus(status);
+    }
+
+    if (environment && environment !== ExtendedRuntimeEnvironment.ALL) {
+      filteredRuns = filteredRuns.filter((run) => run.environment.type === environment);
+      setSelectedEnvironment(environment);
+    }
+
+    setFilteredList(filteredRuns);
+  }, [location.search, list.runs, setFilteredList, setselectedStatus, setSelectedEnvironment]);
```
matt-aitken: This is client-side filtering which isn't the right choice for this. You can have millions of runs which we can't load on the client, so the filtering needs to happen in `RunListPresenter` by passing in the filter data and then incorporating that filter into the database query.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.runs/route.tsx
```diff
@@ -48,6 +110,70 @@ export default function Page() {
...
+
+  const handleFilterChange = (filterType: string, value: string) => {
+    url.set(filterType, value);
+
+    const filters = ["status", "environment"]; // Add more filters as needed
+    const queryString = filters
+      .map((filter) => (url.has(filter) ? `${filter}=${url.get(filter)}` : null))
+      .filter((filter) => filter !== null)
+      .join("&");
+
```
matt-aitken: I think you could just use `url.toString()` here. It turns a URLSearchParams object into a valid query string: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString

matt-aitken (changes requested):
This is good but we need to use server-side filtering not client-side filtering of the runs. Client-side filtering only makes sense when you have all the data on the client, which isn't possible for us as some users have millions of runs.

Abhi1992002:
sry! mistakenly commit without reading all your comments, i will add your prescribed changes and then add a new commit

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.runs/route.tsx
```diff
@@ -48,6 +130,43 @@ export default function Page() {
...
   const project = useProject();
+  const user = useUser();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const url = new URLSearchParams(location.search);
+
+  const [selectedEnvironment, setSelectedEnvironment] = useState<string>(
+    ExtendedRuntimeEnvironment.ALL
+  );
+  const [selectedStatus, setselectedStatus] = useState<string>(ExtendedJobRunStatus.ALL);
```
matt-aitken: Why do you need `useState` when you have the state in the URL?

By having both useState and the URL, you get into the tricky position where you have to sync the state between them. For example, does this work properly if you refresh the page after having set filters? My guess is it doesn't work. 

I think it would be better to just derive the selectedEnvironment from the URL, and then update the state by setting the URL which you're already doing. Then URLs can be shared between team members and no state syncing is needed.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.runs/route.tsx
```diff
@@ -48,6 +130,43 @@ export default function Page() {
...
+    const environment = url.get("environment");
+
+    if (status && status !== "ALL") {
+      setselectedStatus(status);
+    }
+
+    if (environment && environment !== "ALL") {
+      setSelectedEnvironment(environment!);
+    }
+  });
```
matt-aitken: You could totally remove this if you ditched the useState

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.runs/route.tsx
matt-aitken: When you select an option you get a glitchy flash, which I think is caused by this

Uploading CleanShot 2023-12-20 at 11.34.09.mp4…

matt-aitken (approved):
This is great, thanks @Abhi1992002

PR #815, opened 2023-12-27
feat: Add $not to eventFilters

### LINKED ISSUE #810: [TRI-1795] Add $not to eventFilters
EventFilters support lots of different nice matching, but they don't support `not`.

Add it to the schema:
https://github.com/triggerdotdev/trigger.dev/blob/main/packages/core/src/schemas/eventFilter.ts

Add the implementation:
https://github.com/triggerdotdev/trigger.dev/blob/main/packages/core/src/eventFilterMatches.ts

Add tests:
https://github.com/triggerdotdev/trigger.dev/blob/main/packages/core/test/eventFilterMatches.test.ts

<sub>[TRI-1795](https://linear.app/triggerdotdev/issue/TRI-1795/add-dollarnot-to-eventfilters)</sub>

meetwithyash:
Closes #810

## ✅ Checklist

---

## Changelog

- not filter added in event filters
---

### DIFF
--- a/packages/core/test/eventFilterMatches.test.ts
+++ b/packages/core/test/eventFilterMatches.test.ts
@@ -252,6 +252,30 @@ describe("eventFilterMatches", () => {
     expect(eventFilterMatches(payload, filter)).toBe(true);
   });
 
+  it("should return true when payload matches an not condition", () => {
+    const payload = {
+      name: "John",
+      age: 30,
+      score: 100,
+      isAdmin: false,
+      hobbies: ["reading", "swimming"],
+      address: {
+        street: "123 Main St",
+        city: "Anytown",
+        state: "CA",
+        zip: "12345",
+      },
+    };
+    const filter: EventFilter = {
+      hobbies: [{ $not: "gaming" }],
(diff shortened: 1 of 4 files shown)


packages/core/test/eventFilterMatches.test.ts
```diff
@@ -252,6 +252,30 @@ describe("eventFilterMatches", () => {
     expect(eventFilterMatches(payload, filter)).toBe(true);
   });
 
+  it("should return true when payload matches an not condition", () => {
```
matt-aitken, line 255: This is great! We should add a false test as well because it's relatively easy to make a code change in the future where all conditions get evaluated to true.

matt-aitken (changes requested):
Could you add a false test as well?

packages/core/test/eventFilterMatches.test.ts
meetwithyash: @matt-aitken , Pushed update with other test case, Thanks!

matt-aitken (approved):

PR #820, opened 2024-01-01
use --env cli flag for deno 

guy-borderless:
using the `--env` cli flag requires fewer permissions (no need for env and fs access)


matt-aitken (approved):
Thanks! This is cleaner

PR #832, opened 2024-01-10
feat: runs filtering by relative timeframe

Kritik-J:
## ✅ Checklist

---

## Changelog

Add ability to filter runs based on relative timeframe

---

## Video
[screen-capture.webm](https://github.com/triggerdotdev/trigger.dev/assets/82360525/b93765c8-8df7-4a56-ab9c-efe68bd8d96c)

💯


apps/webapp/app/components/runs/RunStatuses.tsx
```diff
@@ -169,3 +177,35 @@ export const filterableStatuses: Record<FilterableStatus, JobRunStatus[]> = {
...
+  "1 hour",
+  "3 hours",
+  "6 hours",
+  "1 day",
+  "3 days",
+  "7 days",
+  "10 days",
+  "14 days",
+  "30 days",
+];
```
matt-aitken: Given that these are details that are only related to the relative time filtering component, I think it makes more sense for them to not be in here.

In fact, I think it'd be better to just have a const and infer the type from that (in the other file).

Like this:

```ts
const timeFrameValues = [
  {
    label: "5 mins",
    value: 5 * 60 * 1000,
  },
  {
    label: "15 mins",
    value: 15 * 60 * 1000,
  },
  {
    label: "30 mins",
    value: 30 * 60 * 1000,
  },
  {
    label: "1 hour",
    value: 60 * 60 * 1000,
  },
  {
    label: "3 hours",
    value: 3 * 60 * 60 * 1000,
  },
  {
    label: "6 hours",
    value: 6 * 60 * 60 * 1000,
  },
  {
    label: "1 day",
    value: 24 * 60 * 60 * 1000,
  },
  {
    label: "3 days",
    value: 3 * 24 * 60 * 60 * 1000,
  },
  {
    label: "7 days",
    value: 7 * 24 * 60 * 60 * 1000,
  },
  {
    label: "10 days",
    value: 10 * 24 * 60 * 60 * 1000,
  },
  {
    label: "14 days",
    value: 14 * 24 * 60 * 60 * 1000,
  },
  {
    label: "30 days",
    value: 30 * 24 * 60 * 60 * 1000,
  },
];

type RelativeTimeFrameItem = typeof timeFrameValues[number];
```

apps/webapp/app/components/runs/RunTimeFrameFilter.tsx
```diff
@@ -0,0 +1,110 @@
...
+      return 10 * 24 * 60 * 60 * 1000;
+    case "14 days":
+      return 14 * 24 * 60 * 60 * 1000;
+    case "30 days":
+      return 30 * 24 * 60 * 60 * 1000;
+    default:
+      const _exhaustiveCheck: never = timeframe;
+      throw new Error(`Non-exhaustive match for value: ${timeframe}`);
+  }
+}
```
matt-aitken: As per previous comment, I'd just use a const with the label and values in one array which simplifies all of this.

matt-aitken (changes requested):
I think it'd be simpler to use a const with both the label and value for the relative time periods together.

Other than that this looks really good 🥳

matt-aitken (approved):
This is great, thanks a lot @Kritik-J

PR #836, opened 2024-01-11
fix: stripe webhook validation

michrosenberger:
The SubtleCryptoProvider cannot be used in a synchronous context. This is a problem for non-node environments where the Stripe client falls back to SubtleCryptoProvider.

Running the trigger.dev Stripe integration on Cloudflare Pages results in the following error:
```
{
  "error": {
    "name": "Error",
    "message": "SubtleCryptoProvider cannot be used in a synchronous context.\\nUse `await constructEventAsync(...)` instead of `constructEvent(...)`"
  },
  "timestamp": "2024-01-11T19:44:47.237Z",
  "name": "trigger.dev",
  "message": "[@trigger.dev/stripe] Error while validating webhook signature",
  "level": "error"
}
```

Even worse, without verbose logging enabled, this is just silently ignored and no event or job-run is triggered from trigger.dev

## ✅ Checklist

---

## Changelog

Fix Stripe integration webhook validation.

💯

### DIFF
--- a/.changeset/shy-needles-push.md
+++ b/.changeset/shy-needles-push.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/stripe": patch
+---
+
+Fix SubtleCryptoProvider webhook validation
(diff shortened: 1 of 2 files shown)


ericallam:
Has this been tested to still work on Node.js?

michrosenberger:
Yes, tested and still works fine on Node.js.

matt-aitken (approved):

Issue #856, opened 2024-01-17
bug: Unable to Add OAuth Integrations when Self-Hosting


erin-allison:
### Provide environment information

This environment is running using the docker-compose file from `https://github.com/triggerdotdev/docker`

```
  System:
    OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (16) x64 AMD Ryzen 7 7800X3D 8-Core Processor
    Memory: 29.29 GB / 30.93 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 18.18.2 - /usr/local/bin/node
    npm: 9.8.1 - /usr/local/bin/npm
    pnpm: 7.18.1 - /usr/local/bin/pnpm
```

### Describe the bug

When self-hosting Trigger.dev, an error is encountered when attempting to set up integrations using OAuth2 tokens.

![image](https://github.com/triggerdotdev/trigger.dev/assets/75811601/211c5a8b-d428-40bf-a8b4-d69b9236bda6)
![image](https://github.com/triggerdotdev/trigger.dev/assets/75811601/79023451-0757-4f3e-ab47-b2846540ca7f)
![image](https://github.com/triggerdotdev/trigger.dev/assets/75811601/6e129be1-3efc-48ec-9d55-9d274e9401cd)

### Reproduction repo

https://github.com/triggerdotdev/docker

### To reproduce

1. Deploy the compose file from `https://github.com/triggerdotdev/docker`
2. Authenticate and navigate to the integrations page
3. Select a Trigger.dev integration which supports OAuth2 (such as Slack)
4. Select OAuth2 (Bot), Developers and then fill in any value for ID, Name and Client ID/Secret
5. Click the "Connect to __" button
6. Note that an error is presented instead of being navigated to the service for authentication.

### Additional information

This issue appears to be related to how the `hasCustomClient` field is not sent with the form submission to the server. If that field isn't present, the route handler doesn't pass the provided client ID & secret to the auth repository for the rest of the process.

![image](https://github.com/triggerdotdev/trigger.dev/assets/75811601/a7b192f6-2817-471f-805e-351aa92b10ac)

https://github.com/triggerdotdev/trigger.dev/blob/2c328ba93df5167aa74712866979561307017008/apps/webapp/app/routes/resources.connection.%24organizationId.oauth2.ts#L70-L74
[...]

PR #859, opened 2024-01-19
fix: Allow self-hosted trigger.dev installations to add OAuth integrations

erin-allison:
Fixes triggerdotdev/trigger.dev#856

## ✅ Checklist

---

## Testing

Using an installation of trigger.dev with a clean database I:
- Attempted to add a Slack integration with the "Use my OAuth App" checkbox unchecked
- Added a Slack integration using OAuth credentials for my company's Slack server
- Set its status to "MISSING_FIELDS" (`UPDATE "Integration" SET "setupStatus" = 'MISSING_FIELDS'`)
- Open the "update" form using the "Integrations Missing Fields" workflow
- Note: I can't figure out how to properly trigger that workflow, so I can't fully step through this case, but it does make it to where it attempts to write to the secret store, which means the client form and server request validation function properly.

---

## Changelog

Updated `ConnectToOAuthForm` and `UpdateOAuthForm` components (and associated server route handlers) to enable the "Use my OAuth App" checkbox on self-hosted installations, but validate its state on form submission.

---

## Screenshots

image
image

### DIFF
--- a/apps/webapp/app/components/integrations/ConnectToOAuthForm.tsx
+++ b/apps/webapp/app/components/integrations/ConnectToOAuthForm.tsx
@@ -52,7 +52,7 @@ export function ConnectToOAuthForm({
       onValidate({ formData }) {
         return parse(formData, {
           // Create the schema without any constraint defined
-          schema: createSchema(),
+          schema: createSchema({isManagedCloud}),
         });
       },
     });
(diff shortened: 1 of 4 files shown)


matt-aitken:
I've created a separate pull request #872 that fixes this issue. It makes it work as it was originally meant to work, it was as simple as the value of hasCustomClient wasn't being submitted by the form…

PR #872, opened 2024-01-26
Fix custom oauth client

### LINKED ISSUE #856: bug: Unable to Add OAuth Integrations when Self-Hosting

matt-aitken:
Closes #856 

The value of the checkbox wasn't coming through because we'd set it to "disabled". I had assumed that the value came through still in the form data… but it doesn't. So now we set the checkbox to readonly instead, I modified the checkbox component so it behaves nicely and is styled properly too.

I have tested with custom OAuth clients.

Note that `CLOUD_ENV=development` should NOT be used by anyone self-hosting Trigger.dev. It controls a bunch of things specific to the cloud, including communication with some private services. We intentionally don't document this environment variable anywhere. If you don't have this set the form should behave correctly for you, and force you to add custom OAuth clients each time.

### DIFF
--- a/apps/webapp/app/components/integrations/ConnectToOAuthForm.tsx
+++ b/apps/webapp/app/components/integrations/ConnectToOAuthForm.tsx
@@ -127,7 +127,7 @@ export function ConnectToOAuthForm({
             id="hasCustomClient"
             label="Use my OAuth App"
             variant="simple/small"
-            disabled={requiresCustomOAuthApp}
+            readOnly={requiresCustomOAuthApp}
             onChange={(checked) => setUseMyOAuthApp(checked)}
             {...conform.input(hasCustomClient, { type: "checkbox" })}
             defaultChecked={requiresCustomOAuthApp}
@@ -135,8 +135,9 @@ export function ConnectToOAuthForm({
           {useMyOAuthApp && (
             <div className="ml-6 mt-2">
               <Paragraph variant="small" className="mb-2">
-                Set the callback url to <CodeBlock code={callbackUrl} showLineNumbers={false} />
+                Set the callback url to
               </Paragraph>
+              <CodeBlock code={callbackUrl} showLineNumbers={false} />
               <div className="flex flex-col gap-2">
                 <div className="flex gap-2">
                   <InputGroup fullWidth>
--- a/apps/webapp/app/components/integrations/UpdateOAuthForm.tsx
+++ b/apps/webapp/app/components/integrations/UpdateOAuthForm.tsx
@@ -117,7 +117,7 @@ export function UpdateOAuthForm({
             id="hasCustomClient"
             label="Use my OAuth App"
             variant="simple/small"
-            disabled={requiresCustomOAuthApp}
+            readOnly={requiresCustomOAuthApp}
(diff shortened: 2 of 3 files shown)


matt-aitken:
@erin-allison this fix is simpler than your PR and is now behaving as I originally designed it too 🤦  Basically the checkbox was set to disabled so wasn't coming through in the form data. I've tested it works.

Make sure you remove CLOUD_ENV from your environment variables, self-hosters shouldn't be using this.

erin-allison:
Yep, this way works too - I just know that there is no OOTB "read-only" option on checkboxes, and wasn't familiar enough with the UI side of things to try and implement this approach

erin-allison:
Yep, this way works too - I just know that there is no OOTB "read-only" option on checkboxes, and wasn't familiar enough with the UI side of things to try and implement this approach

matt-aitken:
It's crazy that there isn't built-in support for readonly on a checkbox…

PR #878, opened 2024-01-30
Set endpoint URLs to null, instead of deleting them

matt-aitken:
Hard deleting endpoints has quite a few bad side effects:
- It deletes all associated runs
- It deletes all associated webhooks
- It deletes other assets that are associated

This PR changes the behaviour so "deleting" an endpoint actually just sets the URL to null. Most of the changes were to guard against a null URL and not continue.

### DIFF
--- a/apps/webapp/app/components/run/RunOverview.tsx
+++ b/apps/webapp/app/components/run/RunOverview.tsx
@@ -117,6 +117,7 @@ export function RunOverview({ run, trigger, showRerun, paths, currentUser }: Run
             {showRerun && run.isFinished && (
               <RerunPopover
                 runId={run.id}
+                runPath={paths.run}
                 runsPath={paths.runsPath}
                 environmentType={run.environment.type}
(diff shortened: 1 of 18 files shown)


apps/webapp/app/components/run/RunOverview.tsx
```diff
@@ -347,6 +350,7 @@ function RerunPopover({
       <PopoverContent className="flex min-w-[20rem] max-w-[20rem] flex-col gap-2 p-0" align="end">
         <Form method="post" action={`/resources/runs/${runId}/rerun`} {...form.props}>
           <input {...conform.input(successRedirect, { type: "hidden" })} defaultValue={runsPath} />
+          <input {...conform.input(failureRedirect, { type: "hidden" })} defaultValue={runPath} />
```
matt-aitken, line 353: If we fail, we redirect back to this page and show an error toast message

nicktrn (approved):

PR #879, opened 2024-01-30
Faster jobs page

matt-aitken:
The Jobs page was slow to load because getting the last run for each job was very slow.

Prisma was doing a weird query where it was returning all of the runs for a project, and then in code it was getting the latest one for each job. The query took about 1 second but then the code took a very long time for organizations with millions of runs.

The new query only gets the latest run for each org, then does the join in code.

I've also added a composite index to make getting the latest run faster. Plus use defer on the runs page.


swag42dev:
https://github.com/triggerdotdev/trigger.dev/issues/886
@matt-aitken there is issue connected with theese changes.

PR #880, opened 2024-02-01
Rename and delete projects

### LINKED ISSUE #638: [TRI-1414] Ability to delete a Project
You can't delete a Project from the UI at the moment.

## Considerations
By deleting a project that has Jobs that are deployed to Staging or Prod, it will break user's apps.
- Indexing their Jobs will fail because the API keys won't exist anymore
- They won't be sent any events to Trigger Jobs
- All API and SDK calls will fail for that deleted Project

We need to make sure that users are fully aware of what will happen if they choose to delete it. That should include showing them what will be impacted, e.g. a list of Jobs with the environments that they're active in, and the number of runs they've had in the past week.

We could also have soft-deleting (i.e. archiving) as the first step. This would have the same impact as deleting except the data would still exist (with an archived flag). These would live in the UI and could be fully deleted from there, and automatically deleted after being in the archive for a certain period.

<sub>[TRI-1414](https://linear.app/triggerdotdev/issue/TRI-1414/ability-to-delete-a-project)</sub>

matt-aitken:
Closes #638

Sidebar changes
- The title in the dropdown is the project name
- The section title above the project is just "Project"
- The section title above the organization is just "Organization"
- Added "Project settings" link

Project settings
- Rename project
- Delete project

Deleting a project
- Disables all endpoints by setting the URL to null
- Disables and soft-deletes all jobs
- Disables any linked schedules for the jobs
- Marks the project as deleted
- Clears the projectId from the session

### DIFF
--- a/apps/webapp/app/components/navigation/SideMenu.tsx
+++ b/apps/webapp/app/components/navigation/SideMenu.tsx
@@ -4,7 +4,6 @@ import {
   ArrowRightOnRectangleIcon,
   ChartBarIcon,
   CursorArrowRaysIcon,
-  EllipsisHorizontalIcon,
   ShieldCheckIcon,
 } from "@heroicons/react/20/solid";
 import { UserGroupIcon, UserPlusIcon } from "@heroicons/react/24/solid";
@@ -15,6 +14,7 @@ import { useFeatures } from "~/hooks/useFeatures";
 import { MatchedOrganization } from "~/hooks/useOrganizations";
 import { MatchedProject } from "~/hooks/useProject";
 import { User } from "~/models/user.server";
+import { useV3Enabled } from "~/root";
 import { useCurrentPlan } from "~/routes/_app.orgs.$organizationSlug/route";
 import { cn } from "~/utils/cn";
 import {
@@ -56,9 +56,8 @@ import {
   PopoverSectionHeader,
 } from "../primitives/Popover";
 import { StepNumber } from "../primitives/StepNumber";
-import { MenuCount, SideMenuItem } from "./SideMenuItem";
 import { SideMenuHeader } from "./SideMenuHeader";
-import { useV3Enabled } from "~/root";
+import { MenuCount, SideMenuItem } from "./SideMenuItem";
 
 type SideMenuUser = Pick<User, "email" | "admin"> & { isImpersonating: boolean };
 type SideMenuProject = Pick<
@@ -106,11 +105,7 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
             showHeaderDivider ? " border-border" : "border-transparent"
           )}
         >
-          <ProjectSelector
-            organization={organization}
-            organizations={organizations}
-            project={project}
-          />
+          <ProjectSelector organizations={organizations} project={project} />
           <UserMenu user={user} />
         </div>
         <div
@@ -118,7 +113,7 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
           ref={borderRef}
         >
           <div className="mb-6 flex flex-col gap-1 px-1">
-            <SideMenuHeader title={project.name || "No project found"}>
+            <SideMenuHeader title={"Project"}>
               <PopoverMenuItem
                 to={projectSetupPath(organization, project)}
                 title="Framework setup"
@@ -170,7 +165,7 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
             />
           </div>
           <div className="mb-1 flex flex-col gap-1 px-1">
-            <SideMenuHeader title={organization.title}>
+            <SideMenuHeader title={"Organization"}>
               <PopoverMenuItem to={newProjectPath(organization)} title="New Project" icon="plus" />
               <PopoverMenuItem
                 to={inviteTeamMemberPath(organization)}
@@ -209,7 +204,7 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
           </div>
         </div>
         <div className="flex flex-col gap-1 border-t border-border p-1">
-          {currentPlan?.subscription?.isPaying === true ? (
+          {currentPlan?.subscription?.isPaying === true && (
             <Dialog>
               <DialogTrigger asChild>
                 <Button
@@ -265,16 +260,14 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
                 </div>
               </DialogContent>
             </Dialog>
-          ) : (
-            <SideMenuItem
-              name="Join our Discord"
-              icon={DiscordIcon}
-              to="https://trigger.dev/discord"
-              data-action="join our discord"
-              target="_blank"
-            />
           )}
-
+          <SideMenuItem
+            name="Join our Discord"
+            icon={DiscordIcon}
+            to="https://trigger.dev/discord"
+            data-action="join our discord"
+            target="_blank"
+          />
           <SideMenuItem
             name="Documentation"
             icon="docs"
@@ -317,11 +310,9 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
 
 function ProjectSelector({
   project,
-  organization,
   organizations,
 }: {
   project: SideMenuProject;
-  organization: MatchedOrganization;
   organizations: MatchedOrganization[];
 }) {
   const [isOrgMenuOpen, setOrgMenuOpen] = useState(false);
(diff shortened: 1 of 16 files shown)


matt-aitken:
I've dealt with deleting the last project in an org now. It takes you to the new project page, without a cancel button

apps/webapp/app/components/navigation/SideMenu.tsx
```diff
@@ -4,7 +4,6 @@ import {
   ArrowRightOnRectangleIcon,
```
matt-aitken, line 4: The changes in this file are improvements to the sidebar, mostly unrelated to deleting/renaming projects.
- The dropdown title is now the Project name (instead of the org name)
- The section titles are now "Project" and "Organization" because it was confusing before.

apps/webapp/app/components/navigation/SideMenu.tsx
```diff
@@ -265,16 +268,14 @@ export function SideMenu({ user, project, organization, organizations }: SideMen
...
-          ) : (
-            <SideMenuItem
-              name="Join our Discord"
-              icon={DiscordIcon}
-              to="https://trigger.dev/discord"
-              data-action="join our discord"
-              target="_blank"
-            />
           )}
-
```
matt-aitken, line 277: Paying users get the option of both Discord and Slack. Before it was only Slack if they're paying, which sucked because some people were happily using Discord.

apps/webapp/app/presenters/OrganizationsPresenter.server.ts
```diff
@@ -2,13 +2,16 @@ import { PrismaClient } from "@trigger.dev/database";
...
   getCurrentProjectId,
   setCurrentProjectId,
 } from "~/services/currentProject.server";
 import { logger } from "~/services/logger.server";
 import { newProjectPath } from "~/utils/pathBuilder";
 import { ProjectPresenter } from "./ProjectPresenter.server";
+import { redirectWithErrorMessage } from "~/models/message.server";
+import { match } from "assert";
 
 export class OrganizationsPresenter {
```
matt-aitken, line 16: This file deals with selecting a project, as well as organizations. This is because we always have a selected project due to the nature of our sidebar/design.

Where possible we now redirect users to the correct page rather than just showing them an error.

apps/webapp/app/presenters/OrganizationsPresenter.server.ts
```diff
@@ -75,16 +82,21 @@ export class OrganizationsPresenter {
...
         const session = await setCurrentProjectId(bestProject.id, request);
         throw redirect(request.url, {
           headers: { "Set-Cookie": await commitCurrentProjectSession(session) },
         });
       }
 
-      //use the project param to find the project
-      const project = await prisma.project.findFirst({
+      //get all the projects
+      const projects = await prisma.project.findMany({
```
matt-aitken, line 99: This allows us to check if there are any projects, as well as find the exact project.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.settings/route.tsx
```diff
@@ -0,0 +1,245 @@
+import { conform, useForm } from "@conform-to/react";
```
matt-aitken, line 1: A new page where you can renamed your project or delete it.

To delete you need to type in the slug to confirm you want to delete it.

CleanShot 2024-02-02 at 11 33 39

apps/webapp/app/routes/_app.orgs.$organizationSlug_.projects.new/route.tsx
```diff
@@ -1,24 +1,59 @@
 import { conform, useForm } from "@conform-to/react";
```
matt-aitken, line 1: It's now possible to have an Org with no projects. So we've moved the new project page outside of the org route.

If you have zero projects there's a panel explaining why you need to add a project. When we allow deleting orgs we can have a link to that page in the callout too.

apps/webapp/app/routes/resources.environments.$environmentParam.endpoint.$endpointParam.ts
```diff
@@ -1,6 +1,6 @@
 import { ActionFunctionArgs, json } from "@remix-run/server-runtime";
 import { z } from "zod";
-import { DeleteEndpointIndexService } from "~/services/endpoints/deleteEndpointService";
```
matt-aitken, line 3: Simple rename because I realised it's deleting an endpoint, not an index

apps/webapp/app/services/deleteProject.server.ts
```diff
@@ -0,0 +1,142 @@
+import { PrismaClient } from "@trigger.dev/database";
+import { prisma } from "~/db.server";
+import { DisableJobService } from "./jobs/disableJob.server";
+import { AuthenticatedEnvironment } from "./apiAuth.server";
+import { DeleteJobService } from "./jobs/deleteJob.server";
+import { DeleteEndpointService } from "./endpoints/deleteEndpointService";
+import { logger } from "./logger.server";
+import { DisableScheduleSourceService } from "./schedules/disableScheduleSource.server";
+
+type Options = { projectId: string; userId: string } | { projectSlug: string; userId: string };
```
matt-aitken, line 10: The slug variant is used at the moment. The id one will be used when deleting an org and iterating through the projects.

apps/webapp/app/services/deleteProject.server.ts
```diff
@@ -0,0 +1,142 @@
...
+        },
+        data: {
+          enabled: false,
+        },
+      });
+      const eventDispatchers = await this.#prismaClient.eventDispatcher.findMany({
+        where: {
+          environmentId: environment.id,
+        },
+      });
```
matt-aitken, line 74: You can't do `updateMany` and return them with Prisma at the moment. It just returns the number of items that were modified.

packages/database/prisma/migrations/20240131105237_project_deleted_at/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- DropIndex
+DROP INDEX "idx_jobrun_jobid_createdat";
```
matt-aitken, line 2: It removed the index from the previous PR because I hadn't defined it in the Prisma schema file… fixed below

packages/database/prisma/migrations/20240202115155_added_job_run_index_back_in_using_prisma_schema/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX "idx_jobrun_jobId_createdAt" ON "JobRun"("jobId", "createdAt" DESC);
```
matt-aitken, line 2: The index added back in because I've now used the Prisma schema to add it

apps/webapp/app/models/runtimeEnvironment.server.ts
```diff
@@ -7,6 +7,9 @@ export async function findEnvironmentByApiKey(apiKey: string) {
   const environment = await prisma.runtimeEnvironment.findUnique({
     where: {
       apiKey,
+      project: {
```
matt-aitken: This will block API calls if the project is deleted

apps/webapp/app/models/runtimeEnvironment.server.ts
```diff
@@ -30,22 +36,3 @@ export async function findEnvironmentByPublicApiKey(apiKey: string) {
...
-  });
-
-  if (!organization) {
-    return;
-  }
-
-  const environment = organization.environments.find((environment) => environment.slug === slug);
-
-  return environment;
-}
```
matt-aitken: This wasn't being used anywhere so it's best to delete it

apps/webapp/app/models/runtimeEnvironment.server.ts
ericallam: I wouldn't add this to the query, but just check the return value. This could introduce an inefficient query in a hot path so easier to just check afterwards and return undefined.

apps/webapp/app/services/deleteProject.server.ts
```diff
@@ -0,0 +1,142 @@
...
+          },
+          data: {
+            deletedAt: new Date(),
+          },
+        });
+
+        //disable scheduled sources
+        for (const eventDispatcher of eventDispatchers) {
+          await service.call({
+            key: job.id,
```
ericallam, line 100: Are we sure this key value is correct here?

apps/webapp/app/models/runtimeEnvironment.server.ts
matt-aitken: Good idea

apps/webapp/app/services/deleteProject.server.ts
matt-aitken: It's taken from the disabling job code here: https://github.com/triggerdotdev/trigger.dev/blob/a9ff32418e4be194a76a91a195c782f7ae98e28f/apps/webapp/app/services/jobs/disableJob.server.ts#L99

PR #892, opened 2024-02-08
Fix edge cases with deleting last org/project

matt-aitken:
There were a few bugs with project/org deleting
- The code that selects the best project to show if there's no cookie and the user lands on the root of the site wasn't factoring in deleted projects
- If you delete your last project (across all orgs) an error was happening
- If an org had any deleted projects (ever) it couldn't be deleted

You could create multiple orgs and projects with the same name by hitting entering twice on the new org page

### DIFF
--- a/apps/webapp/app/components/navigation/SideMenu.tsx
+++ b/apps/webapp/app/components/navigation/SideMenu.tsx
@@ -357,23 +357,31 @@ function ProjectSelector({
           <Fragment key={organization.id}>
             <PopoverSectionHeader title={organization.title} />
             <div className="flex flex-col gap-1 p-1">
-              {organization.projects.map((p) => {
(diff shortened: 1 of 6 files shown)


apps/webapp/app/components/navigation/SideMenu.tsx
```diff
@@ -357,23 +357,31 @@ function ProjectSelector({
...
+                      isSelected={isSelected}
+                      icon="folder"
+                    />
+                  );
+                })
+              ) : (
+                <PopoverMenuItem
+                  to={newProjectPath(organization)}
+                  title="New project"
+                  icon="plus"
```
matt-aitken, line 382: If you delete the last project in an org and then navigate somehow to a different org you end up with an org you can't get to in the nav. This allows you to create a project for that org.

PR #903, opened 2024-02-19
v3 queue system

ericallam:
Implements a simple fair & multi-tenant queueing system for v3 in Redis. This PR also adds "internal" telemetry that can either be logged out or shipped to an OTLP endpoint, depending on the env vars (by default telemetry will be logged out to the console). 

The follow instrumentations are enabled currently:

- HTTP
- Prisma
- Express

I've also added support in the dev CLI for env variable support


apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+type MessagePayload = z.infer<typeof MessagePayload>;
+
+const SemanticAttributes = {
+  QUEUE: "marqs.queue",
+  PARENT_QUEUE: "marqs.parentQueue",
+  MESSAGE_ID: "marqs.messageId",
+  CONCURRENCY_KEY: "marqs.concurrencyKey",
+};
+
+// TODO: heartbeats from the workers to ensure they're still alive
```
matt-aitken: @ericallam didn't you say this was done now?

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+
+/**
+ * MarQS - Modular Asynchronous Reliable Queueing System (pronounced "markus")
+ */
+export class MarQS {
```
matt-aitken, line 56: Wouldn't it be better if the M was for "Multitenant"?

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+            : constants.SHARED_QUEUE;
+
+        // Read the parent queue for matching queues
+        const messageQueue = await this.#getRandomQueueFromParentQueue(
+          parentQueue,
+          (queue, score) => this.#calculateMessageQueueWeight(queue, score)
+        );
+
+        if (!messageQueue) {
+          abort();
```
matt-aitken, line 140: Would it be useful to add some logging before all of these aborts?

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+        const rawMessage = await this.redis.get(`${constants.MESSAGE_PART}:${messageId}`);
+
+        if (!rawMessage) {
+          return;
+        }
+
+        const message = MessagePayload.safeParse(JSON.parse(rawMessage));
+
+        if (!message.success) {
+          return;
```
matt-aitken, line 357: Is this a bug if this happens? If so we should log the error from Zod.

apps/webapp/app/v3/marqs.server.ts
ericallam: No because it's going to be VERY common for dequeue to abort, and the whole point is to prevent unnecessary logging.

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+      return;
+    }
+
+    for (let i = 0; i < messages.length; i++) {
+      const message = messages[i];
+
+      const messageData = await this.redis.get(`${constants.MESSAGE_PART}:${message}`);
+
+      if (!messageData) {
+        await this.redis.zrem(constants.MESSAGE_VISIBILITY_TIMEOUT_QUEUE, message);
```
matt-aitken, line 496: If there's no message data what does that mean?

If it's a bug then we should log

matt-aitken: A code comment would be useful if it's not a bug

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+
+        continue;
+      }
+
+      const parsedMessage = MessagePayload.safeParse(JSON.parse(messageData));
+
+      if (!parsedMessage.success) {
+        await this.redis.zrem(constants.MESSAGE_VISIBILITY_TIMEOUT_QUEUE, message);
```
matt-aitken, line 504: Same goes here, if it's a bug that the message wasn't valid we should log the errors from Zod

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+
+    const concurrencyLimit =
+      (await this.redis.get(`${concurrencyQueueName}:${constants.CONCURRENCY_LIMIT_PART}`)) ?? 100;
+    const currentConcurrency = await this.redis.scard(
+      `${queue}:${constants.CURRENT_CONCURRENCY_PART}`
+    );
+
+    const capacity = Number(concurrencyLimit) - Number(currentConcurrency);
+
+    const capacityWeight = capacity / Number(concurrencyLimit);
```
matt-aitken: We should guard against the concurrencyLimit being zero (or negative).

If it's zero then `capacityWeight` will be Infinity.

apps/webapp/app/v3/marqs.server.ts
```diff
@@ -0,0 +1,906 @@
...
+      maxVisibilityTimeout: string,
+      callback?: Callback<void>
+    ): Result<void, Context>;
+  }
+}
+
+export const marqs = singleton("marqs", getMarQSClient);
+
+function getMarQSClient() {
+  if (env.REDIS_HOST && env.REDIS_PORT) {
```
matt-aitken: Should we log an error if these variables aren't set?

We could warn people that for v3 they will need to have Redis configured?

apps/webapp/app/v3/marqs/environmentQueueConsumer.server.ts
```diff
@@ -0,0 +1,384 @@
...
+
+    if (!message) {
+      setTimeout(() => this.#doWork(), 1000);
+      return;
+    }
+
+    const messageBody = MessageBody.safeParse(message.data);
+
+    if (!messageBody.success) {
+      // TODO: do some kind of DLQ thing?
```
matt-aitken: Should we log something here, with the message id and error message?

apps/webapp/app/v3/marqs/environmentQueueConsumer.server.ts
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    const backgroundTask = backgroundWorker.tasks.find(
+      (task) => task.slug === existingTaskRun.taskIdentifier
+    );
+
+    if (!backgroundTask) {
+      // TODO: some kind of DLQ thing?
+      await marqs?.acknowledgeMessage(message.messageId);
+
```
matt-aitken, line 251: For now should we log a message?

apps/webapp/app/v3/marqs/environmentQueueConsumer.server.ts
```diff
@@ -0,0 +1,384 @@
...
+        attempts: {
+          take: 1,
+          orderBy: { number: "desc" },
+        },
+        tags: true,
+      },
+    });
+
+    if (!lockedTaskRun) {
+      await marqs?.acknowledgeMessage(message.messageId);
```
matt-aitken, line 281: I think a log here makes sense too?

apps/webapp/app/v3/marqs/environmentQueueConsumer.server.ts
```diff
@@ -0,0 +1,384 @@
...
+    };
+
+    const payload = {
+      execution,
+      traceContext: lockedTaskRun.traceContext as Record<string, unknown>,
+    };
+
+    // TODO: send trace context down to the CLI
+    this._sender.send("BACKGROUND_WORKER_MESSAGE", {
+      backgroundWorkerId: backgroundWorker.friendlyId,
```
ericallam: handle when the websocket connection is closed

matt-aitken (approved):

PR #905, opened 2024-02-21
v3: Implemented in-process task retries in dev w/cache support

ericallam:
Task retrying now works in dev and uses the same process across attempts (as long as the CLI isn't exited).


ericallam (commented):
Good job me

PR #908, opened 2024-02-23
v3: batch trigger

ericallam:
Ability to batch trigger and batch trigger and wait for the results. Introduces a new concept of a `TaskRunDependency` which creates a relationship from either a `BatchRun` or a `TaskRunAttempt` and a subsequent `TaskRun`, this is so we can "resolve" the `BatchRun` or `TaskRunAttempt` once a `TaskRun` completes (in the case of a `BatchRun`, only when all `BatchRunItem`'s are resolved.
CleanShot 2024-02-26 at 11 20 05

CleanShot 2024-02-26 at 11 20 28


ericallam (commented):
This good

PR #911, opened 2024-02-24
Fix: Remove fixed public schema name from migration to enable custom schema names

terehov:
Closes #<issue>

## ✅ Checklist

---

## Testing

Migration "20240130165343_add_composite_index_to_job_run_for_job_id_and_created_at" had a fixed "public" schema in the migration, preventing from running it against a custom schema name, like proposed in the docs. 

---

## Changelog

Removed "public" from migration.


matt-aitken:
Unfortunately it’s not as easy as this to fix the problem.

This will work for anyone starting from scratch but it will fail, block all new migrations and potentially wipe the database of anyone who has already performed this migration.

Prisma records migrations that have been run in a table along with a hash of the migration file. That hash won’t match and so it will fail.

I think the solution is to delete this migration file entirely and write a SQL script that runs before the migrations that deletes the record of the migration running. The index added in this migration was re-added in a later one anyway.

PR #923, opened 2024-03-04
v3: no bundle builds for the dev CLI

ericallam:
- Use `bundle: false` for our dev CLI so otel instrumentation works and packages are instead installed to `.trigger/node_modules`.


ericallam (commented):
👍

PR #1002, opened 2024-04-05
v3: crash observability

nicktrn:
This adds:
- New `CRASHED` status
- Displaying of crash errors for both runs and deployments
- Fixes for prod cancellation and reconnects


matt-aitken (approved):
Great 🤩

PR #1003, opened 2024-04-05
v3: prod image upgrade and fixes

nicktrn:
This adds:
- Better support for third party packages and binaries
- Write permissions for workdir
- Switch from alpine to bookworm-slim
- Small fix to final attempt shutdown

### DIFF
--- a/apps/kubernetes-provider/src/index.ts
+++ b/apps/kubernetes-provider/src/index.ts
@@ -417,7 +417,7 @@ class KubernetesTaskOperations implements TaskOperations {
     type: THookType,
     cause: THookType extends "postStart" ? PostStartCauses : PreStopCauses
   ) {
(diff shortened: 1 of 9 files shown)


apps/kubernetes-provider/src/index.ts
```diff
@@ -417,7 +417,7 @@ class KubernetesTaskOperations implements TaskOperations {
     type: THookType,
     cause: THookType extends "postStart" ? PostStartCauses : PreStopCauses
   ) {
-    return ["/bin/sh", "-c", `sleep 1; wget -q -O- 127.0.0.1:8000/${type}?cause=${cause}`];
+    return ["/bin/sh", "-c", `for i in $(seq 1 5); do sleep 1; wget -q -O- 127.0.0.1:8000/${type}?cause=${cause} && break; done`];
```
matt-aitken: For things like this which are a bit hard to read (if you're not awesome at the command line) it'd be good to have a code comment on the line above saying what it does.

PR #1030, opened 2024-04-13
v3: fix dependency trigger and wait

nicktrn:
- Fix repeated trigger waits by removing unique constraints from schema and converting to arrays
- docs: Add alternatives to running (batch) trigger and wait in parallel
- Update docker lifecycle command to use busybox
- Replace clear filters trash icon with `x`


apps/webapp/app/v3/services/resumeAttempt.server.ts
```diff
@@ -78,6 +87,8 @@ export class ResumeAttemptService extends BaseService {
         return;
```
matt-aitken, line 87: What happens to the run attempt when this function returns undefined?

docs/v3/triggering.mdx
```diff
@@ -165,6 +165,43 @@ export const myTask = task({
...
+
+<CodeGroup>
+```ts /trigger/batch.ts
+export const batchTask = task({
+  id: "batch-task",
+  run: async (payload: string) => {
+    const results = await childTask.batchTriggerAndWait({
+      items: [{ payload: "item1" }, { payload: "item2" }],
+      // optional: how many to run in parallel
+      batchOptions: { maxConcurrency: 1 },
```
matt-aitken: I didn't know we had this! Does it work?

apps/webapp/app/v3/services/resumeAttempt.server.ts
```diff
@@ -24,7 +25,7 @@ export class ResumeAttemptService extends BaseService {
         },
         include: {
           taskRun: true,
-          taskRunDependency: {
+          dependencies: {
```
matt-aitken, line 28: @ericallam you need to look at this file. I don't have enough knowledge of the system right now.

apps/webapp/app/v3/services/resumeAttempt.server.ts
nicktrn: We don't currently do anything with the return value as resumes are two separate messages. Tasks saying ready for resume, platform sending the resume data. There's room to fail runs in a bunch of places (and shut down the workers), but it's tough because a single no-op message could mean the entire run fails.

docs/v3/triggering.mdx
nicktrn: We do have this, but it's a mystery (I didn't actually check)

apps/webapp/app/v3/services/resumeAttempt.server.ts
nicktrn: Previously, an attempt could only have a single dependency at a time. Meaning we had to clear it to add a subsequent one. (This is where we ran into unique constraint errors.)

The mechanics are still the same now, you can't wait for multiple dependencies at once. Only the latest dependency in the array is considered.

docs/v3/triggering.mdx
matt-aitken: @ericallam does this work?

apps/webapp/app/v3/services/resumeAttempt.server.ts
matt-aitken: Makes sense, let's leave it for now then I think until we are more certain on the real world solution to this (if there even is a problem)

docs/v3/triggering.mdx
ericallam: If it does it would be a surprise to me!

apps/webapp/app/v3/services/resumeAttempt.server.ts
ericallam: I think this just needs testing on test cloud, it looks okay to me but hard to tell just by looking at it

matt-aitken (approved):

PR #1033, opened 2024-04-14
v3: fix raw queries for custom schemas

nicktrn:
Raw queries can fail when using a custom postgres schema without adjusting the [search path](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH). This fixes that issue by:
- Extracting the schema from `DATABASE_URL`
- Using that schema in all raw queries


apps/webapp/app/presenters/v3/DeploymentListPresenter.server.ts
```diff
@@ -106,9 +106,9 @@ export class DeploymentListPresenter {
   u."avatarUrl" AS "userAvatarUrl", 
   wd."deployedAt"
 FROM
```
matt-aitken, line 108: I think there's a missing one on line 100 @nicktrn

matt-aitken (changes requested):
I think there's one missing line

matt-aitken (approved):

PR #1034, opened 2024-04-16
v3: prod worker graceful shutdown

nicktrn:
- Give workers 10 minutes to finish their current attempt when receiving SIGTERM
- Fail run with timeout error if it doesn't exit in that timeframe
- Fix an issue where large timeout delays could exceed 32 bit signed integer limits


matt-aitken:
Some questions on this:
- In what situations would workers get a SIGTERM?
- What's the worst outcome for a customer running a task and they received a SIGTERM?

nicktrn:
Should probably have prefaced this by stating the previous default, which was to quit immediately. This is more of a precaution to limit situations that are difficult to debug. With these changes we'll know when this happens, and we have control over the shutdown process.

>In what situations would workers get a SIGTERM?

Manually or automatically terminating containers / pods.
- Manual: May be required when runs appear "stuck" or otherwise unresponsive.
- Automatic:  Could be for a number of reasons. OOM conditions or other critical issues, and when scaling down with tasks still on the affected node. I'm trying to reduce the chances of any of this happening.

>What's the worst outcome for a customer running a task and they received a SIGTERM?

Worst case would be the entire run fails and has to be replayed.

Could up the grace period to an hour for now?

packages/core/src/v3/schemas/common.ts
```diff
@@ -49,6 +50,7 @@ export const TaskRunInternalError = z.object({
     "TASK_RUN_CANCELLED",
     "TASK_OUTPUT_ERROR",
     "HANDLE_ERROR_ERROR",
+    "GRACEFUL_EXIT_TIMEOUT"
```
matt-aitken, line 53: What happens if someone is on an older version of the SDK and receives one of these new errors from the platform? Or isn't that possible?

matt-aitken (commented):
I had one question

packages/core/src/v3/schemas/common.ts
nicktrn: Those errors get sent from the worker to the platform, so that wouldn't be possible.

matt-aitken (approved):

Issue #1014, opened 2024-04-09
[TRI-2206] Dev option so logs come through to the local console  (v3)


matt-aitken:
When running the v3 CLI logs only go to the platform, they're not visible in your local terminal.

We should add a setting to `trigger.config`, e.g. `logToConsole: boolean`. 

The logs would need to be prefixed with the run_id (multiple runs can be happening at once)

<sub>[TRI-2206](https://linear.app/triggerdotdev/issue/TRI-2206/dev-option-so-logs-come-through-to-the-local-console-v3)</sub>

matt-aitken:
This was added to `trigger.config`. Use `enableConsoleLogging: true`

PR #1036, opened 2024-04-16
Scheduled tasks

matt-aitken:
Schedule tasks
- Define scheduled tasks in your `schedules.task()`
- Manage schedules in the dashboard
- Do a test run of a schedule in the dashboard
- Manage schedules using the SDK
- Manage schedules using the REST API

Other fixes
- Idempotency key when triggering a task
- Replaying runs that have a superjson payload type (e.g. have Date in)
- Canceling a run wasn't removing a run from your concurrency count


ericallam:
Snuck in a fix for #1014 as well:

CleanShot 2024-04-16 at 21 47 42@2x

PR #1043, opened 2024-04-18
v3: Fixes for using `(batch)triggerAndWait` with idempotency keys

ericallam:
Fixes various issues with triggerAndWait and batchTriggerAndWait

When using idempotency keys, triggerAndWait and batchTriggerAndWait will still work even if the existing runs have already been completed (or even partially completed, in the case of batchTriggerAndWait)

- TaskRunExecutionResult.id is now the run friendlyId, not the attempt friendlyId
- A single TaskRun can now have many batchItems, in the case of batchTriggerAndWait while using idempotency keys
- A run’s idempotencyKey is now added to the ctx as well as the TaskEvent and displayed in the span view
- When resolving batchTriggerAndWait, the runtimes no longer reject promises, leading to an error in the parent task

Additionally, this PR contains a fix for tasks getting incorrectly throttled because of the default queue concurrency in `MarQS`, which has now been removed (if a task queue does not have a concurrency limit set, then it will be limited by either the env concurrency limit or the org concurrency limit)

### DIFF
--- a/apps/webapp/app/models/taskRun.server.ts
+++ b/apps/webapp/app/models/taskRun.server.ts
@@ -1,0 +1,108 @@
+import {
+  TaskRunError,
+  TaskRunExecutionResult,
+  TaskRunFailedExecutionResult,
+  TaskRunSuccessfulExecutionResult,
+} from "@trigger.dev/core/v3";
+import {
+  BatchTaskRunItemStatus,
+  TaskRun,
+  TaskRunAttempt,
+  TaskRunAttemptStatus,
+  TaskRunStatus,
+} from "@trigger.dev/database";
+import { assertNever } from "assert-never";
+
+const SUCCESSFUL_STATUSES = [TaskRunStatus.COMPLETED_SUCCESSFULLY];
+const FAILURE_STATUSES = [
+  TaskRunStatus.CANCELED,
+  TaskRunStatus.INTERRUPTED,
+  TaskRunStatus.COMPLETED_WITH_ERRORS,
+  TaskRunStatus.SYSTEM_FAILURE,
+  TaskRunStatus.CRASHED,
+];
+
+export type TaskRunWithAttempts = TaskRun & {
+  attempts: TaskRunAttempt[];
+};
+
+export function executionResultForTaskRun(
+  taskRun: TaskRunWithAttempts
+): TaskRunExecutionResult | undefined {
+  if (SUCCESSFUL_STATUSES.includes(taskRun.status)) {
+    // find the last attempt that was successful
+    const attempt = taskRun.attempts.find((a) => a.status === TaskRunAttemptStatus.COMPLETED);
+
+    if (!attempt) {
+      return undefined;
+    }
+
+    return {
+      ok: true,
+      id: taskRun.friendlyId,
+      output: attempt.output ?? undefined,
+      outputType: attempt.outputType,
+    } satisfies TaskRunSuccessfulExecutionResult;
+  }
+
+  if (FAILURE_STATUSES.includes(taskRun.status)) {
+    if (taskRun.status === TaskRunStatus.CANCELED) {
+      return {
+        ok: false,
+        id: taskRun.friendlyId,
+        error: {
+          type: "INTERNAL_ERROR",
+          code: "TASK_RUN_CANCELLED",
+        },
+      } satisfies TaskRunFailedExecutionResult;
+    }
+
+    const attempt = taskRun.attempts.find((a) => a.status === TaskRunAttemptStatus.FAILED);
+
+    if (!attempt) {
+      return undefined;
+    }
+
+    const error = TaskRunError.safeParse(attempt.error);
+
+    if (!error.success) {
+      return {
+        ok: false,
+        id: taskRun.friendlyId,
+        error: {
+          type: "INTERNAL_ERROR",
+          code: "CONFIGURED_INCORRECTLY",
+        },
+      } satisfies TaskRunFailedExecutionResult;
+    }
+
+    return {
+      ok: false,
+      id: taskRun.friendlyId,
+      error: error.data,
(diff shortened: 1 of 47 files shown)


apps/webapp/app/models/taskRun.server.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export function executionResultForTaskRun(
+  taskRun: TaskRunWithAttempts
+): TaskRunExecutionResult | undefined {
+  if (SUCCESSFUL_STATUSES.includes(taskRun.status)) {
+    // find the last attempt that was successful
+    const attempt = taskRun.attempts.find((a) => a.status === TaskRunAttemptStatus.COMPLETED);
+
+    if (!attempt) {
+      return undefined;
```
matt-aitken, line 44: Is it an error with the system if this happens? Should we log something?

apps/webapp/app/models/taskRun.server.ts
```diff
@@ -0,0 +1,108 @@
...
+          type: "INTERNAL_ERROR",
+          code: "TASK_RUN_CANCELLED",
+        },
+      } satisfies TaskRunFailedExecutionResult;
+    }
+
+    const attempt = taskRun.attempts.find((a) => a.status === TaskRunAttemptStatus.FAILED);
+
+    if (!attempt) {
+      return undefined;
```
matt-aitken, line 76: Same question here, is this a bug with the system and should we log something?

apps/webapp/app/models/taskRun.server.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const error = TaskRunError.safeParse(attempt.error);
+
+    if (!error.success) {
+      return {
```
matt-aitken, line 89: Do we want to log the `attempt.error` payload here so we can try figure out why there's a mismatch?

apps/webapp/app/presenters/v3/ApiBatchResultsPresenter.server.ts
```diff
@@ -0,0 +1,46 @@
...
+                  },
+                },
+              },
+            },
+          },
+        },
+      });
+
+      if (!batchRun) {
+        return undefined;
```
matt-aitken, line 35: Is this a bug with the system and should we log something?

apps/webapp/app/presenters/v3/ApiRunResultPresenter.server.ts
```diff
@@ -0,0 +1,33 @@
...
+          attempts: {
+            orderBy: {
+              createdAt: "desc",
+            },
+          },
+        },
+      });
+
+      if (!taskRun) {
+        return undefined;
```
matt-aitken, line 27: Is this a bug with the system and should we log something?

apps/webapp/app/v3/services/resumeTaskRunDependencies.server.ts
```diff
@@ -34,14 +34,16 @@ export class ResumeTaskRunDependenciesService extends BaseService {
       return;
     }
 
-    const { batchItem, dependency } = taskAttempt.taskRun;
+    const { batchItems, dependency } = taskAttempt.taskRun;
 
-    if (!batchItem && !dependency) {
+    if (!batchItems.length && !dependency) {
       return;
```
matt-aitken, line 40: Is this a bug with the system and should we log something?

apps/webapp/app/v3/services/triggerTask.server.ts
```diff
@@ -34,18 +34,20 @@ export class TriggerTaskService extends BaseService {
...
-            idempotencyKey,
-          },
-        },
-      });
+      const existingRun = idempotencyKey
+        ? await this._prisma.taskRun.findUnique({
+            where: {
+              runtimeEnvironmentId_idempotencyKey: {
+                runtimeEnvironmentId: environment.id,
+                idempotencyKey,
```
matt-aitken, line 44: Now idempotencyKey can be null will this work?

packages/core/src/v3/runtime/prodRuntimeManager.ts
```diff
@@ -139,10 +139,14 @@ export class ProdRuntimeManager implements RuntimeManager {
...
-    if (completion.ok) {
+    if (!wait.reject) {
       wait.resolve(completion);
     } else {
-      wait.reject(completion);
+      if (completion.ok) {
+        wait.resolve(completion);
+      } else {
+        wait.reject(completion);
+      }
```
matt-aitken, line 149: @nicktrn I don't know enough about this to review it

matt-aitken (approved):

packages/core/src/v3/runtime/prodRuntimeManager.ts
nicktrn: Looks like `batchTriggerAndWait()` should never throw, so we always return and array of all run results, failed and successful. This is different to `triggerAndWait()` where we throw on failure. Good fix!

apps/webapp/app/models/taskRun.server.ts
nicktrn: This looks similar to what we do in a bunch of places. To fail or not to fail? We might want to have a chat at some point.

nicktrn (approved):

apps/webapp/app/v3/services/triggerTask.server.ts
ericallam: Yup because we check if idempotency is truthy before calling this

apps/webapp/app/v3/services/resumeTaskRunDependencies.server.ts
ericallam: Nope, this just means theres no dependencies to resume

PR #1090, opened 2024-05-07
Improvement: Added more context to Shopify webhook registering errors

ramadanomar:
## About Issue

When initializing the Shopify Client with invalid parameters (excluding apiKey), errors are not explicitly thrown, resulting in a silent failure in the background. The user only sees an error message stating, "Error: Missing webhook ID for delete operation." This occurs because the current logic attempts to delete a webhook subscription first if its creation fails, possibly to avoid the "address already in use" exception. However, this logic is poorly documented, and only users running a trigger client with verbose mode enabled will see the debug message, "ERROR CREATING WEBHOOK RESTARTING WITH DELETE FIRST." In cases of invalid secret keys, this approach is ineffective and does not resolve the issue.

## ✅ Checklist

---

## Testing

Localhost & Shopify Dev Store

---

## Changelog

- Added input validation for constructor parameters to ensure required arguments are correct.
- Verified that the store domain uses the .myshopify.com domain instead of a custom primary domain.
- Integrated logger.error statements into the webhook created/update/delete logic for improved error tracking.

---

### Old Behaviour
image
### New Behaviour
image

💯

### DIFF
--- a/integrations/shopify/src/index.ts
+++ b/integrations/shopify/src/index.ts
@@ -63,6 +63,26 @@ export class Shopify implements TriggerIntegration {
       throw `Can't create Shopify integration (${options.id}) as apiKey was undefined`;
     }
 
+    if (Object.keys(options).includes("apiSecretKey") && !options.apiSecretKey) {
+      throw `Can't create Shopify integration (${options.id}) as apiSecretKey was undefined`;
+    }
+
+    if (Object.keys(options).includes("adminAccessToken") && !options.adminAccessToken) {
+      throw `Can't create Shopify integration (${options.id}) as adminAccessToken was undefined`;
+    }
+
+    if (Object.keys(options).includes("hostName") && !options.hostName) {
+      throw `Can't create Shopify integration (${options.id}) as hostName was undefined`;
+    }
+
+    // Regular expression to ensure the hostname ends with `.myshopify.com`
+    const shopifyDomainPattern = /^[a-zA-Z0-9-]+\.myshopify\.com$/;
+
+    // Verify that the user has entered the .myshopify.com domain and not a custom primary domain
+    if (!shopifyDomainPattern.test(options.hostName)) {
+      throw `Can't create Shopify integration (${options.id}) because hostName should be a ".myshopify.com" domain, not a custom primary domain.`;
+    }
+
     this._options = options;
     this._shopDomain = this._options.hostName.replace("http://", "").replace("https://", "");
   }
--- a/integrations/shopify/src/index.ts
+++ b/integrations/shopify/src/index.ts
@@ -75,16 +75,18 @@ export class Shopify implements TriggerIntegration {
       throw `Can't create Shopify integration (${options.id}) as hostName was undefined`;
     }
 
-    // Regular expression to ensure the hostname ends with `.myshopify.com`
+    this._options = options;
+    // Extract the shop domain if user has entered the full URL
+    this._shopDomain = this._options.hostName
+      .replace(/^https?:\/\//, "") // Remove http:// or https://
+      .replace(/\/$/, ""); // Remove trailing slash if it exists (e.g. `example.myshopify.com/`)
+
+    // Regular expression to ensure the shopDomain is a valid `.myshopify.com` domain
     const shopifyDomainPattern = /^[a-zA-Z0-9-]+\.myshopify\.com$/;
 
-    // Verify that the user has entered the .myshopify.com domain and not a custom primary domain
-    if (!shopifyDomainPattern.test(options.hostName)) {
-      throw `Can't create Shopify integration (${options.id}) because hostName should be a ".myshopify.com" domain, not a custom primary domain.`;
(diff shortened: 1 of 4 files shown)


integrations/shopify/src/index.ts
```diff
@@ -63,6 +63,26 @@ export class Shopify implements TriggerIntegration {
...
+    if (Object.keys(options).includes("hostName") && !options.hostName) {
+      throw `Can't create Shopify integration (${options.id}) as hostName was undefined`;
+    }
+
+    // Regular expression to ensure the hostname ends with `.myshopify.com`
+    const shopifyDomainPattern = /^[a-zA-Z0-9-]+\.myshopify\.com$/;
+
+    // Verify that the user has entered the .myshopify.com domain and not a custom primary domain
+    if (!shopifyDomainPattern.test(options.hostName)) {
+      throw `Can't create Shopify integration (${options.id}) because hostName should be a ".myshopify.com" domain, not a custom primary domain.`;
```
nicktrn: Does this definitely not work when specifying your custom domain instead? I'm worried this might break existing setups. What if someone specified the hostname with a slash at the end, e.g. `example.myshopify.com/`? What about `https://example.myshopify.com`? (We actually account for that one on line 67)

We should probably make the pattern less strict and add an example to the error message.

```suggestion
      throw `Can't create Shopify integration (${options.id}) because hostName should be a ".myshopify.com" domain, not a custom primary domain. For example: my-domain.myshopify.com`;
```

integrations/shopify/src/index.ts
```diff
@@ -63,6 +63,26 @@ export class Shopify implements TriggerIntegration {
...
+    if (Object.keys(options).includes("adminAccessToken") && !options.adminAccessToken) {
+      throw `Can't create Shopify integration (${options.id}) as adminAccessToken was undefined`;
+    }
+
+    if (Object.keys(options).includes("hostName") && !options.hostName) {
+      throw `Can't create Shopify integration (${options.id}) as hostName was undefined`;
+    }
+
+    // Regular expression to ensure the hostname ends with `.myshopify.com`
+    const shopifyDomainPattern = /^[a-zA-Z0-9-]+\.myshopify\.com$/;
```
nicktrn, line 79:
```suggestion
    const shopifyDomainPattern = /.+\.myshopify\.com\/?$/;
```

nicktrn (changes requested):
Thanks for this! Just one concern about backwards compat re the hostname option. Let me know what you think. Only accept the code suggestions if you think they make sense. :pray:

integrations/shopify/src/index.ts
ramadanomar: [According to the shopify docs
](https://shopify.dev/docs/api/admin-rest#endpoints)
> All Admin REST API endpoints follow this pattern:
> https://{store_name}.myshopify.com/admin/api/2024-04/{resource}.json

If OAuth will ever be used then i think hostName can be a custom domain, but from my experience with shopify rest api, if you dont use the stores .myshopify.com domain (initial one when created, so even if you change your .myshopify.com domain) [you will get a 301 redirect with no info about your current request.](https://community.shopify.com/c/graphql-basics-and/custom-domain-to-call-api/m-p/2387000)

If backwards compatibility is an issue we can remove this check completly and have it mentioned somewhere in the integration docs. Trailing slashes are a no go. Their specs are mentioned [here](https://github.com/Shopify/shopify-app-js/blob/main/packages/apps/shopify-api/docs/reference/shopifyApi.md)

But from experience things can get pretty confusing. If we want to allow custom domains we would have to pass a `customShopDomains`. 

tbh spending too much time on this is not even worth it as shopify is planning and will release new features only on the GraphQL API (for example if you want to access the new 2k variant limit instead of 100 you will have to use graphql api)

Even without the checks people should still get the error message logged when response.ok fails so it should be more than enough to debug it

TL/DR: i can remove it completly and maybe mention somewhere in the docs (not necessary tho)

ramadanomar: [Somewhere after Additionally, you will also have to provide your shop domain.](https://trigger.dev/docs/integrations/apis/shopify#personal-access-tokens)

nicktrn: Okay, that all makes sense. Let's keep this simple. Could you keep everything as is but move the domain validation after line 67 where we remove protocol prefixes, and validate `this._shopDomain` instead?

ramadanomar: I'm done with the changes but i also tested if i use another .myshopify.com domain (you can change it once for every store i think). The webhook appears as created but it's not registering correctly.

image
image

It's because it always returns a 301 redirect. i can also update it to throw an error warning the user to use the primary domain. My guess would be where we check for res.ok 

```
 if (!response.ok) {
          await handleWebhookError("WEBHOOK_CREATE", request, response);
}
```

I'm not sure if this is the actual best place to put that check.

ramadanomar: image

I think this should do?

nicktrn (approved):
Thanks for the changes! Made some small edits and will merge later if all goes well.

PR #1103, opened 2024-05-15
v2: Graphile auto-cleanup and auto-endpoint disabling

ericallam:
- Failed graphile jobs are no longer saved and cleaned up 7 days later, instead they are automatically deleted by *not* throwing in the run task
- Auto disable endpoints that have more than a day of failed indexes in a row
- No longer schedule webhooks to be delivered if an endpoint is disabled


ericallam (commented):
Looks good to me!

PR #1122, opened 2024-05-22
chore: update stripe integration

ryanleecode:
Update stripe client to the latest version. This is necessary since if you create a new Stripe account you are forced to use `2024-04-10` but the current integration is stuck on  `2022-11-15`

---

## Testing

It compiles and works. Tried `onCheckoutSessionCompleted` and `onCustomerSubscriptionUpdated` and there were no issues.

---

## Changelog

Update stripe to ~15.7.0

---


ericallam:
@ryanleecode this looks great! Can you add a changeset file to this PR as well? In the root of the repo just run `pnpm run changeset:add` and then do a `patch` upgrade for the stripe integration package 👍

ryanleecode:
Done

ryanleecode:
@ericallam Any blockers left for this PR?

PR #1127, opened 2024-05-23
fix: update remix deps to 2.9.2

jacobparis:
All of the breaking changes between Remix 2.1 to 2.9 occurred in Vite which you're not using yet, so this went smoothly.

I did have to install @conform-to/dom to peg it to its 0.* version because the other conform packages were installing it from 1.* and that was incompatible. We can fix that later when I take a look at updating Conform.

Closes #<issue>

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/triggerdotdev/trigger.dev/blob/main/CONTRIBUTING.md)
- [x] The PR title follows the convention.
- [x] I ran and tested the code works

---

## Testing

_[Describe the steps you took to test this change]_

---

## Changelog

_[Short description of what has changed]_

---

## Screenshots

_[Screenshots]_

💯

PR #1128, opened 2024-05-24
Switch to vite

jacobparis:
**Depends on:** https://github.com/triggerdotdev/trigger.dev/pull/1127

This PR switches the webapp to Vite which requires changing how the dev server works

## The Vite Config

The biggest issue with Vite is its dependency handling. Vite LOVES esm, and it gets top performance by using native browser imports to esm chunks. Unfortunately, that means it spends time transpiling (what it calls "optimizing") dependencies that are CJS or aren't declared quite the right way. 

Most of this should be done on initial load automatically now via the `server.warmup.clientFiles` field.

But other deps don't need to be optimized, so we can explicitly exclude them to speed up navigations.

```ts
optimizeDeps: {
    // Exclude if dep is ESM or runs server-side (just speeds up Vite on navigations)
    exclude: [
      // Include @prisma/client until @trigger-dev/database is ESM
      // https://github.com/prisma/prisma/issues/21474
      "@prisma/client",
      "@opentelemetry/api-logs",
      "@opentelemetry/api",
      "@opentelemetry/core",
      "@opentelemetry/exporter-trace-otlp-http",
      …
```

## The Express Server 

It uses the [Remix Express Vite Plugin](https://github.com/kiliman/remix-express-vite-plugin) which builds the `express.server.ts` along with the Remix app. This gets bundled by vite, so it's ok to import files from the app directory here. Any modifications to the express `app` server get made here.

Vite takes that bundled express configuration with the bundled Remix app and starts the server, which is where things get tricky

Since all the bundling is done at this point, any changes we want to make to `server` (what app.listen() returns) DO NOT get bundled, so importing modules that use tsconfig paths or other build configurations we may take for granted elsewhere in the app cannot happen. 

There are two places where these changes get made:

In development, we do it right at the Vite plugin in `vite.config.ts`
```ts
expressDevServer({
      exportName: "express",
      async configureServer(server) {
        if (!server) return;

        const { registerSocketIo } = await import("./app/socket.server");
        registerSocketIo(server);
      },
    }),
```

In production, there's a method in `express.server.ts`

```ts
createServer(app) {
    const server = http.createServer(app);

    registerSocketIo(server);

    return server;
  },
```

The `registerSocketIo` function attaches socketIo and the development `wss` server. To get around having to bundle all the event handlers for these, which reach deeply into the utils of Trigger, I replaced `wss` with a stubbier server as a singleton that does not have the event handlers hooked up. 

```ts
export const wss = singleton("wss", () => {
  return new WebSocketServer({ noServer: true });
});

export const io = singleton("socketIo", () => {
  if (!env.REDIS_HOST || !env.REDIS_PORT) {
    console.warn("No redis config found, skipping socket.io");
    return new Server();
  }

  const pubClient = new Redis({
    port: env.REDIS_PORT,
    host: env.REDIS_HOST,
    username: env.REDIS_USERNAME,
    password: env.REDIS_PASSWORD,
    enableAutoPipelining: true,
    ...(env.REDIS_TLS_DISABLED === "true" ? {} : { tls: {} }),
  });
  const subClient = pubClient.duplicate();

  const io = new Server({
    adapter: createAdapter(pubClient, subClient, {
      key: "tr:socket.io:",
      publishOnSpecificResponseChannel: true,
    }),
  });

  return io;
});
```

Since these are singletons, we can import them within the app (which is all bundled code with no restrictions) and attach listeners to the same instance of the server. This will happen on the first request to the app (in `entry.server.ts`), and very likely on the healthcheck ping during deploy if your infrastructure works that way.

```ts
// entry.server.ts
import { initializeWebSocketServer } from "./v3/handleWebsockets.server";
initializeWebSocketServer();

// ./v3/handleWebsockets.server
export function initializeWebSocketServer() {
  return singleton("wss:register", () => {
    wss.on("connection", handleWebSocketConnection);

    authenticatedConnections = new Map();

    new Gauge({
      name: "dev_authenticated_connections",
      help: "Number of authenticated dev connections",
      collect() {
        this.set(authenticatedConnections.size);
      },
      registers: [metricsRegister],
    });

    return wss;
  });
}

async function handleWebSocketConnection(ws: WebSocket, req: IncomingMessage) {
  …
```

Closes #<issue>

## ✅ Checklist

---

## Testing

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯


matt-aitken:
When I run the site locally I get this warning:

```
(node:15501) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGTERM listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

I assume this is referring to the fact we use `process.on("SIGTERM", ...` in quite a few places. This is a new warning I think. Is this because things are running multiple times?

matt-aitken:
Closing because it's been modified in #1167

PR #1133, opened 2024-05-28
v3: Runs List Management SDK and SDK improvements

ericallam:
- Implements the `runs.list` API endpoint and function
- Overhaul the `runs.retrieve` response to add additional helpful things like output, more granular status to match the dashboard, and helpful boolean helpers like `isCompleted`
- API functions now return an `ApiPromise` with helpers like `.asResponse()` and `.withResponse()`
- List endpoints return special "Page Promises" that can be used with an async iterator


apps/webapp/app/presenters/v3/ApiRetrieveRunPresenter.server.ts
```diff
@@ -32,14 +43,68 @@ export class ApiRetrieveRunPresenter extends BasePresenter {
...
+            });
+
+            $output = await parsePacket(outputPacket);
+          }
+        }
+      }
+
+      const apiStatus = ApiRetrieveRunPresenter.apiStatusFromRunStatus(taskRun.status);
+
       return {
```
matt-aitken, line 78: If the overall run has failed with `COMPLETED_WITH_ERRORS`, do we want to bubble an error up to the run level?

docs/v3/management/overview.mdx
```diff
@@ -0,0 +1,220 @@
...
+      console.error(`Unknown error: ${error.message}`);
+    }
+  }
+}
+```
+
+## Auto-pagination
+
+All list endpoints in the management API support auto-pagination.
+You can use `for await … of` syntax to iterate through items across all pages:
```
matt-aitken, line 165: Does this mean if you don't use return or break you'll go through every single run that matches your filters?

If so:
a) this is awesome
b) we should add a warning to make this clear and advise people that this could be a lot of runs and that they should break or return from the loop as early as possible, or filter tightly. Probably putting that in the example code makes sense.

matt-aitken (approved):

apps/webapp/app/presenters/v3/ApiRetrieveRunPresenter.server.ts
ericallam: Just the last attempt error?

matt-aitken: Yeah I guess so, that's what we do in the UI I think?

PR #1138, opened 2024-05-30
Support custom queue when triggering a task

### LINKED ISSUE #1137: [TRI-2618] Overriding the queue when calling trigger/triggerAndWait isn't working
## Reproduction

```ts
const run3 = await simpleChildTask.trigger(
  { message: "Hello, World again!" },
  {
    queue: {
      name: "my-other-queue",
      concurrencyLimit: 3,
    },
  }
);
```

The queue here isn't used, it just uses the queue on the task. This is because it's not being passed through, it is with batchTrigger or batchTriggerAndWait.

## Solution
We need to pass the option through. We also need to merge the properties because it's possible to specify a concurrencyLimit but not a name which is bad.

<sub>[TRI-2618](https://linear.app/triggerdotdev/issue/TRI-2618/overriding-the-queue-when-calling-triggertriggerandwait-isnt-working)</sub>

matt-aitken:
Closes #1137 

When calling trigger or batchTrigger (and their await variants) you can pass in a custom queue. This wasn't properly supported before.

This PR includes a test script (queues) that can be run in the v3-catalog.

### DIFF
--- a/apps/webapp/app/v3/services/triggerTask.server.ts
+++ b/apps/webapp/app/v3/services/triggerTask.server.ts
@@ -5,11 +5,11 @@ import {
   packetRequiresOffloading,
 } from "@trigger.dev/core/v3";
 import { createHash } from "node:crypto";
-import { $transaction } from "~/db.server";
+import { $transaction, prisma } from "~/db.server";
 import { AuthenticatedEnvironment } from "~/services/apiAuth.server";
+import { marqs, sanitizeQueueName } from "~/v3/marqs/index.server";
 import { eventRepository } from "../eventRepository.server";
 import { generateFriendlyId } from "../friendlyIdentifiers";
-import { marqs, sanitizeQueueName } from "~/v3/marqs/index.server";
 import { uploadToObjectStore } from "../r2.server";
 import { BaseService } from "./baseService.server";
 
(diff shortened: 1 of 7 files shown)


apps/webapp/app/v3/services/triggerTask.server.ts
```diff
@@ -183,6 +187,41 @@ export class TriggerTaskService extends BaseService {
...
+                  name: queueName,
+                  concurrencyLimit,
+                  runtimeEnvironmentId: environment.id,
+                  projectId: environment.projectId,
+                  rateLimit: body.options.queue.rateLimit,
+                  type: "NAMED",
+                },
+              });
+
+              if (typeof taskQueue.concurrencyLimit === "number") {
```
ericallam, line 216: What happens here if they previous had a concurrency limit set, and then removed it? These limits wouldn't get cleared. Looks like we might have the same issue here: https://github.com/triggerdotdev/trigger.dev/blob/main/apps/webapp/app/v3/services/createBackgroundWorker.server.ts#L175

matt-aitken: Good point, I'll take a look at that

ericallam (approved):

PR #1149, opened 2024-06-06
v2: MarQS powered job executions

ericallam:
This PR is the start of the conversion of v2 run executions being powered by MarQS instead of graphile worker.  This is currently shipped behind two flags, both of which need to be turned on to use MarQS instead of graphile:

- Set `V2_MARQS_ENABLED=1` env var
- Set the `v2MarqsEnabled = true` for an organization

When `V2_MARQS_ENABLED=1` is set, concurrency settings will be double-written to both MarQS and the legacy graphile system. Enabling MarQS v2 first before enabling orgs allows for the gradual migration to MarQS instead of having to do a "stop the world" migration.

This PR also contains a much improved MarQS "queue choice" algorithm, that will be used for v3 and v2. Other relevant environment variables for operation v2 MarQS:

- The existing `DEFAULT_ORG_EXECUTION_CONCURRENCY_LIMIT` env var is used by MarQS v2
- `V2_MARQS_DEFAULT_ENV_CONCURRENCY` controls the default concurrency per RuntimeEnvironment. This should be set very high as the current graphile system doesn't have the concept of environment-level concurrency, only org.
- `V2_MARQS_CONSUMER_POOL_ENABLED` controls whether to start a consumer pool.
- `V2_MARQS_CONSUMER_POOL_SIZE` controls the number of consumers the pool should manage
- `V2_MARQS_CONSUMER_POLL_INTERVAL_MS` controls how often each consumer polls
- `V2_MARQS_QUEUE_SELECTION_COUNT` controls how many queues MarQS will consider on each dequeueing attempt
- `V2_MARQS_VERBOSE` setting this to `"1"` will log out a bunch of MarQS stuff


apps/webapp/app/v3/marqs/index.server.ts
```diff
@@ -265,9 +285,14 @@ export class MarQS {
...
-        await RequeueTaskRunService.enqueue(
+        await this.options.visibilityTimeoutStrategy.heartbeat(
           messageData.messageId,
-          new Date(Date.now() + this.visibilityTimeoutInMs)
+          this.visibilityTimeoutInMs
+        );
+
+        await this.options.visibilityTimeoutStrategy.heartbeat(
+          messageData.messageId,
+          this.visibilityTimeoutInMs
```
matt-aitken: It looks like this is now happening twice in a row? Maybe a merge issue?

apps/webapp/app/v3/marqs/index.server.ts
```diff
@@ -351,6 +379,11 @@ export class MarQS {
 
         const message = await this.readMessage(messageData.messageId);
 
+        await this.options.visibilityTimeoutStrategy.heartbeat(
```
matt-aitken: Again this is in twice

matt-aitken (approved):

PR #1164, opened 2024-06-13
marqs: add parent queue global concurrency limits

ericallam:
This allows configuring a global concurrency limit for a parent queue (e.g. `marqs:sharedQueue`). If a value is not set in `marqs:sharedQueue:concurrency`, the value from `process.env.DEFAULT_PARENT_QUEUE_EXECUTION_CONCURRENCY_LIMIT` will be used, with a default of 250.


ericallam:
- [ ] Need to remove from parent queue when it's a dev parent queue

matt-aitken:
We didn't find this was needed and the new RunQueue is being worked on now

PR #1165, opened 2024-06-13
Schedule limits and timezone support

matt-aitken:
## Limits
We now limit the number of schedules that can be created in a project. This setting is controlled on a per-organization level and defaults to 5.

We show the limit in the UI. If you're at the limit (or over) when you try and create a new schedule using the UI or SDK an error will be returned.

## Timezones

You can now create a schedule with a CRON pattern and a timezone.

This allows you to schedule a task to run at 8am every weekday in New York time. It will always be 8am New York time no matter the time of year because it automatically adjusts for daylight savings time.

This is very powerful when combined with being able to create schedules for your users. You could show a dropdown of timezones in your app and then create a schedule for your user to receive notifications, summaries, or whatever that are generated and sent by a task.

### In the dashboard
You can select a timezone in the dashboard when creating a schedule. This is useful if you want to have something for your team always be in their timezone.

CleanShot 2024-06-13 at 18 10 54

### With the SDK

Create a schedule with a timezone:

```ts
const createdSchedule = await schedules.create({
  //The id of the scheduled task you want to attach to.
  task: firstScheduledTask.id,
  //The schedule in CRON format.
  cron: "0 0 * * *",
  // Optional, it defaults to "UTC". In IANA format, e.g. "America/New_York".
  // In this case, the task will run at midnight every day in New York time.
  // If you specify a timezone it will automatically work with daylight saving time.
  timezone: "America/New_York",
  //Optionally, you can specify your own IDs (like a user ID) and then use it inside the run function of your task.
  //This allows you to have per-user CRON tasks.
  externalId: "user_123456",
  //You can only create one schedule with this key.
  //If you use it twice, the second call will update the schedule.
  //This is useful because you don't want to create duplicate schedules for a user.
  deduplicationKey: "user_123456-todo_reminder",
});
```

We've also added an SDK function to get all possible timezones that we support.

```ts
const timezones = await schedules.timezones();
```

You can then use this list to render a piece of UI for your users to select from. You can optionally exclude "UTC" from this list if you want:

```ts
const { timezones } = await schedules.timezones({ excludeUtc: true });
```

## Other changes

- `deduplicationKey` is now required when creating a schedule using the SDK. People were creating a lot of duplicate schedules and this should prevent that.
- Fixed some broken docs links from the schedule page to the references

## Testing

- [x] Creating/editing/viewing in the dashboard
- [x] Creating/editing/retrieving with the new SDK
- [ ] Creating/editing/retrieving with the old SDK


packages/core/src/v3/index.ts
matt-aitken, line 1: Zodfetch is already a separate export from core

ericallam (approved):

PR #1175, opened 2024-06-21
feature: add no-trigger-core-import

jacobparis:
Disallows importing from `@trigger.dev/core` and `@trigger.dev/core/v3` directly (favouring more specific exports) and autofixes it

## ✅ Checklist

---

## Testing

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/packages/eslint-plugin/docs/rules/no-trigger-core-import.md
+++ b/packages/eslint-plugin/docs/rules/no-trigger-core-import.md
@@ -1,0 +1,69 @@
+# Prevent importing from `@trigger.dev/core` directly (`trigger-dev/no-trigger-core-import`)
+
+<!-- end auto-generated rule header -->
+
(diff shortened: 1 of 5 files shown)


ericallam:
@jacobparis I don't think it's a good idea to remove the `type` imports, I'd prefer if we used type imports instead of stripping them out. Is there a reason this was needed?

jacobparis:
> @jacobparis I don't think it's a good idea to remove the `type` imports, I'd prefer if we used type imports instead of stripping them out. Is there a reason this was needed?

It's not strictly needed but the eslint autofixing with types sometimes doubles them like `import type { type ArticleSchema }` which causes syntax errors that need to be manually fixed and I wanted to present a solution that didn't require manual intervention

A clean solution would be to run the autofix to remove all type imports and then change the lint rule to run again and add them all from a clean state, which should avoid it getting tangled

But I have no strong opinions here

matt-aitken:
This is great, I've turned off the vscode setting for now so that people don't fix these issues before we merge the Vite branch.

What's going to happen:
1. Merge this PR
2. In the Vite branch pull main in
3. Run the lint command and turn the vscode setting back on
4. Test the app locally
5. Create a build from the branch, deploy and test thoroughly on test cloud
6. Merge the Vite branch into main and deploy to test cloud
7. Go live

PR #1185, opened 2024-06-25
fix: prevent filter tooltip from opening on autofocus

kaiiiiiiiii:
## ✅ Checklist

---

## Testing

I navigated between pages and also did a couple of full page reloads.

---

## Changelog

Change prevents the "Runs" filter tooltip from auto opening on a page reload or page navigation. Focus on the filter element is not affected.

---

## Videos

Before:

https://github.com/triggerdotdev/trigger.dev/assets/4711233/7181ff86-460d-4509-842c-f0a4abe8efb2

After:

https://github.com/triggerdotdev/trigger.dev/assets/4711233/b4419a1e-85a6-4455-a481-ca99ea545bdd

Pardon me if I missed anything, first project PR :)


nicktrn:
Thanks for this @kaiiiiiiiii! I wonder if we need that autofocus here at all. Without it you can't press enter to open the menu, but the `F` shortcut works just fine. I think it looks a bit strange to have the filter button highlighted on entering the page. @matt-aitken?

matt-aitken:
We've just removed the autofocus instead

PR #1200, opened 2024-07-04
v3: Cancel awaited subtasks and reliable rate-limit recovery

ericallam:
- When cancelling a task that has in progress or queued subtasks (using `triggerAndWait` or `batchTriggerAndWait`), those subtasks will now also be cancelled (and any of their subtasks). Tasks triggered with `trigger` or `batchTrigger` will not be cancelled.
- Changing from sliding window to token bucket in the API rate limiter, to help smooth out traffic
- Adding spans to the API Client core & SDK functions
- Added waiting spans when retrying in the API Client
- Retrying in the API Client now respects the x-ratelimit-reset
- Retrying ApiError’s in tasks now respects the x-ratelimit-reset
- Added AbortTaskRunError that when thrown will stop retries
- Added idempotency keys SDK functions and automatically injecting the run ID when inside a task
- Added the ability to configure ApiRequestOptions (retries only for now) globally and on specific calls
- Implement the maxAttempts TaskRunOption (it wasn’t doing anything before)

### DIFF
--- a/apps/webapp/app/v3/services/cancelTaskRun.server.ts
+++ b/apps/webapp/app/v3/services/cancelTaskRun.server.ts
@@ -104,18 +104,6 @@ export class CancelTaskRunService extends BaseService {
       await this.#cancelRemainingRunWorkers(cancelledTaskRun);
     }
 
-    const cancelService = new CancelTaskRunService();
-
-    // Cancel any dependent task runs
-    for (const attempt of cancelledTaskRun.attempts) {
-      for (const dependency of attempt.dependencies) {
-        await cancelService.call(dependency.taskRun, {
-          ...opts,
-          reason: `Parent task run was cancelled`,
-        });
-      }
-    }
-
     return {
       id: cancelledTaskRun.id,
     };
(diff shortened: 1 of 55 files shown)


ericallam (commented):
self-reviewed 👍

ericallam:
- [ ] Add ability to trigger with an idempotency key without the run ID (use optional second param in create)

apps/webapp/app/v3/services/cancelTaskRun.server.ts
```diff
@@ -93,6 +104,18 @@ export class CancelTaskRunService extends BaseService {
...
+    // Cancel any dependent task runs
+    for (const attempt of cancelledTaskRun.attempts) {
+      for (const dependency of attempt.dependencies) {
+        await cancelService.call(dependency.taskRun, {
+          ...opts,
+          reason: `Parent task run was cancelled`,
+        });
+      }
+    }
+
```
matt-aitken: Should this be here?

matt-aitken (approved):

PR #1214, opened 2024-07-13
Added extraCACerts config

### LINKED ISSUE #1210: feat: Allow injecting ca cert in project

I was playing with the self-hosted deployment in an isolated environment with a private registry. Found out that it fails to even build due to private certification authority. 

Build works without problem but can not be run as the entrypoint tries to connect to self hosted trigger.dev api which is secured by self signed/private cert chain.

Allowing to inject private root cert which will allow securely interacting with the internal services would be helpful.

A config option to include a ca file which would then be added to `NODE_EXTRA_CA_CERTS=/app/ca.crt` environment variable during docker build and worker initialization.

Allow replacing `Containerfile.prod` along with build/runtime variable using config instead of through the project setting. project settings can not even be fetched in this scenario

### Additional information

For others who want to be able to do this in current scenario (tested beta-48), do this

1. install `trigger.dev@beta` as dev dependency instead of running from npx, 

2. add `ca.crt` file in your project and update config `additionalFiles: ["./ca.crt"]` to copy this.

3. in file `node_modules\trigger.dev\dist\Containerfile.prod` update the section for ENV with following
```yaml
ENV TRIGGER_PROJECT_ID=${TRIGGER_PROJECT_ID} \
    # ... ....
    NODE_ENV=production \
    NODE_EXTRA_CA_CERTS=/app/lib/ca.crt # this line
```

4. Update `node_modules\trigger.dev\dist\workers\prod\entry-point.js`, update function `gatherProcessEnv` at the very end of file like this
```js
function gatherProcessEnv() {
  const env = {

    NODE_ENV: process.env.NODE_ENV ?? "production",
// ... other config
    NODE_EXTRA_CA_CERTS: process.env.NODE_EXTRA_CA_CERTS, //// THIS LINE
  };
  return Object.fromEntries(Object.entries(env).filter(([key, value]) => value !== void 0));
}
```

step 3 and 4 needs to be done inside `node_modules`, this can be patched with `patch-packages` or `yarn patch`.  caveat is that as of writing this issue, the library is getting updated almost every day.

I can definitely create a pull request over next few days which should fix this if acceptable.

gautamsi:
Closes #1210 

## ✅ Checklist

---

## Testing
1. Generate a self signed cert and use this to protect trigger.dev deployment
2. create a task which tries to access a self hosted api with self signed certificate
3. try to deploy the project to self-hosted instance, see it fail
4. create a ca.crt file containing signing certificate public key in pem format
5. copy this file to project dir
6. add `extraCACerts` config `./ca.crt`
7. deploy the task to self-hosted instance, this time it should deploy and run successfully
---

## Changelog
---
Added config option `extraCACerts` to `ProjectConfig` type. This copies the ca file along with `additionalFiles` and sets `NODE_EXTRA_CA_CERTS` environment variable in built image as well as running the task.

## Screenshots

_[Screenshots]_

💯


packages/cli-v3/src/commands/deploy.ts
```diff
@@ -330,6 +331,7 @@ async function _deployCommand(dir: string, options: DeployCommandOptions) {
         loadImage: options.loadImage,
         buildPlatform: options.buildPlatform,
         noCache: options.noCache,
+        extraCACerts: resolvedConfig.config.extraCACerts?.replace(/^\./,"/app") ?? "",
```
gautamsi: ideally it should be used with `path.resolve`. Should I move this into `resolveConfig`? I am thinking of copying this value into `additionalFiles` and convert this value to absolute path by `resolve('/app",config.additionalCACerts)`

nicktrn: Agreed on `path.resolve()` and on moving everything to `resolveConfig()` :+1: 

Would be great if `extraCACerts` could accept paths outside the project dir, just like `additionalFiles` does: https://github.com/triggerdotdev/trigger.dev/blob/f91a118958fdfcbc6a3f8bafe8c9867db7e26640/packages/cli-v3/src/commands/deploy.ts#L1878-L1884

Essentially, just strip all `..` segments. The posix path sep is due to the requirement of glob patterns. (Yes, there are potential issues with this, but it's what additionalFiles currently does.)

`additionalFiles` will handle the raw paths fine, but we'll have to transform the `resolvedConfig.config.extraCACerts` output.

packages/cli-v3/src/commands/deploy.ts
```diff
@@ -779,6 +781,7 @@ type BuildAndPushImageOptions = {
   loadImage: boolean;
   buildPlatform: string;
   noCache: boolean;
+  extraCACerts: string;
```
nicktrn: This should probably be optional.

packages/cli-v3/src/commands/deploy.ts
```diff
@@ -837,6 +840,8 @@ async function buildAndPushImage(
       `TRIGGER_CONTENT_HASH=${options.contentHash}`,
       "--build-arg",
       `TRIGGER_PROJECT_REF=${options.projectRef}`,
+      "--build-arg",
+      `NODE_EXTRA_CA_CERTS=${options.extraCACerts}`,
```
nicktrn: What happens when this is an empty string or undefined?

packages/cli-v3/src/commands/deploy.ts
```diff
@@ -961,6 +966,8 @@ async function buildAndPushSelfHostedImage(
...
```
nicktrn: Same as above.

packages/cli-v3/src/commands/deploy.ts
```diff
@@ -1821,6 +1828,10 @@ export async function copyAdditionalFiles(
   tempDir: string
 ): Promise<AdditionalFilesReturn> {
   const additionalFiles = config.additionalFiles ?? [];
+  const extraCACerts = config.extraCACerts ?? '';
+  if (extraCACerts) {
+    additionalFiles.push(extraCACerts);
+  }
```
nicktrn: I think we should just handle this in `resolveConfig()` as well. It's going to get confusing with the path resolution step otherwise where `config.extraCACerts !== buildOptions.extraCACerts`

nicktrn (changes requested):
Thanks for this @gautamsi!

Good call with `resolveConfig()`. I think it will be a lot cleaner doing everything in there, that is:

1. Add the raw `extraCACerts` path to `additionalFiles`
2. Tranform it so `resolvedConfig.extraCACerts` now points at the deployed destination path -> `/app/...`

Re the last point: We could even just strip `..` and keep the path relative to `/app` if `NODE_EXTRA_CA_CERTS` works fine with relative paths. `/app` is the work dir in the resulting image so that should be fine.

gautamsi:
I will update this soon

packages/cli-v3/src/commands/deploy.ts
gautamsi: this variable by default is empty, nodejs only loads this variable if this has value during startup.

`When set, the well known "root" CAs (like VeriSign) will be extended with the extra certificates in file. The file should consist of one or more trusted certificates in PEM format. A message will be emitted (once) with [process.emitWarning()](https://nodejs.org/api/process.html#processemitwarningwarning-options) if the file is missing or malformed, but any errors are otherwise ignored.`

packages/cli-v3/src/commands/deploy.ts
gautamsi: I was unable to test relative path scenario, I say I was unable to set config correctly, if set correctly, the new regex would replace any instance of `./` or `../../../` to empty string making it relative to `/app` directory.

packages/cli-v3/src/commands/deploy.ts
gautamsi: converted this to check and apply arg instead of applying undefined.

packages/cli-v3/src/commands/deploy.ts
gautamsi: handled in resolveConfig
https://github.com/triggerdotdev/trigger.dev/blob/f606d3e072224e2496651e24bfe3e0bfdd30b642/packages/cli-v3/src/utilities/configFiles.ts#L246-L259

nicktrn: Love a good regexp

nicktrn (approved):

nicktrn:
Thanks for the changes! Will get this merged and tested. Should be in the next release.

gautamsi:
verified beta-49 works with this feature :)

nicktrn:
I did test this locally briefly and relative paths work as well :+1:

PR #1220, opened 2024-07-16
v3: fix websocket timeouts

nicktrn:
WebSockets could previously time out when no messages were sent for a period of 60s. This is due to load balancer connection timeouts. The only real fix here was to add a ping / pong mechanism that will keep connections alive during periods of rest, e.g. when waiting within or between attempts.

Pings will be automatically sent from the server every 45s. Clients will reply with pongs, but we just log those for now.

Whenever a dev CLI disconnects it will also cancel all in-progress dev runs. This shouldn't happen anymore.

This fix also prevents uncaught exceptions related to schema parsing.

### DIFF
--- a/apps/webapp/app/v3/authenticatedSocketConnection.server.ts
+++ b/apps/webapp/app/v3/authenticatedSocketConnection.server.ts
@@ -91,6 +91,13 @@ export class AuthenticatedSocketConnection {
             }
           }
         },
+        PING: async () => {
+          logger.debug("[AuthenticatedSocketConnection] Received ping", {
+            id: this.id,
+            envId: this.authenticatedEnv.id,
+          });
+          await this._sender.send("PONG", {});
+        },
       },
     });
   }
--- a/apps/webapp/app/v3/authenticatedSocketConnection.server.ts
+++ b/apps/webapp/app/v3/authenticatedSocketConnection.server.ts
@@ -126,13 +126,6 @@ export class AuthenticatedSocketConnection {
             }
           }
         },
-        PING: async () => {
-          logger.debug("[AuthenticatedSocketConnection] Received ping", {
-            id: this.id,
-            envId: this.authenticatedEnv.id,
-          });
-          await this._sender.send("PONG", {});
-        },
       },
     });
   }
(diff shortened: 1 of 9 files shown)


apps/webapp/app/v3/authenticatedSocketConnection.server.ts
```diff
@@ -50,8 +53,42 @@ export class AuthenticatedSocketConnection {
     ws.addEventListener("close", this.#handleClose.bind(this));
     ws.addEventListener("error", this.#handleError.bind(this));
 
+    ws.on("ping", (data) => {
+      logger.debug("[AuthenticatedSocketConnection] Received ping", {
```
ericallam, line 57: I think lets remove these logs, will add a ton of additional events to baselime that is probably not worth it

apps/webapp/app/v3/authenticatedSocketConnection.server.ts
```diff
@@ -50,8 +53,42 @@ export class AuthenticatedSocketConnection {
...
+    });
+
+    this._pingService = new HeartbeatService({
+      heartbeat: async () => {
+        if (ws.readyState !== WebSocket.OPEN) {
+          logger.debug("[AuthenticatedSocketConnection] Websocket not open, skipping ping");
+          return;
+        }
+
+        logger.debug("[AuthenticatedSocketConnection] Sending ping", {
```
ericallam, line 79: Same with this one. If you want to keep these maybe add a `logger.dev` or something that only logs when NODE_ENV=development

apps/webapp/app/v3/authenticatedSocketConnection.server.ts
nicktrn: We should never receive any pings, would be good to know when we do

apps/webapp/app/v3/authenticatedSocketConnection.server.ts
nicktrn: Think I'll just comment this out for now

apps/webapp/app/v3/authenticatedSocketConnection.server.ts
nicktrn: I'll remove the pong one

PR #1225, opened 2024-07-18
add xdg command check for linux servers

HorizonCode:
## ✅ Checklist

## Changelog

adds a check if the current system is linux, then check if xdg is available on there, if not the open function is not called.

---

## Screenshots

image

💯

### DIFF
--- a/.changeset/green-pens-battle.md
+++ b/.changeset/green-pens-battle.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": patch
+---
+
+fix open on linux-server systems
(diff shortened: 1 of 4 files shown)


nicktrn:
Thanks for this @HorizonCode and sorry about the delay. Just saw that installing `xdg-utils` might fix this, even if `xdg-open` is a no-op on most servers. I say most, because `xdg-open https://example.com` will actually work if you have a text-based browser like w3m or lynx installed.

HorizonCode:
i know that this should work if xdg-utils is installed, but this should fix it when its not installed, and not throw an error, maybe add a alert in the login thing when the check failed, that xdg-utils is not installed instead of just crashing the whole application

nicktrn (approved):
Thanks for the PR and sorry it took a while to get this in. Tested locally, works great! Only linted and added some logging.

PR #1226, opened 2024-07-18
Declarative schedules

matt-aitken:
This is for this roadmap feature: https://feedback.trigger.dev/p/static-cron-attached-to-tasks

We've called them "declarative" schedules as they're declared in your code on a task like this:

```ts
export const firstScheduledTask = schedules.task({
  id: "first-scheduled-task",
  //every two hours (UTC timezone)
  cron: "0 */2 * * *",
  run: async (payload, { ctx }) => {
    //do something
  },
});
```

Or you can specify a timezone if you don't want UTC:

```ts
export const secondScheduledTask = schedules.task({
  id: "second-scheduled-task",
  cron: {
    //5am every day Tokyo time
    pattern: "0 5 * * *",
    timezone: "Asia/Tokyo",
  },
  run: async (payload) => {},
});
```

These schedules are synced when you run the `dev` or `deploy` CLI commands. If you create, edit or delete the cron property it will be reflected when this sync happens.

You can't delete, disable or enable them using the other SDK functions that work with imperative schedules.

### DIFF
--- a/apps/webapp/app/v3/services/createBackgroundWorker.server.ts
+++ b/apps/webapp/app/v3/services/createBackgroundWorker.server.ts
@@ -224,11 +224,6 @@ export async function createBackgroundTasks(
   }
 }
 
-//todo syncStaticSchedules
-//1. update
-//2. create
-//delete
-// - get all static ones and see if any are missing
 export async function syncStaticSchedules(
   tasks: TaskResource[],
   worker: BackgroundWorker,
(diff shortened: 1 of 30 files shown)


apps/webapp/app/v3/services/createBackgroundWorker.server.ts
```diff
@@ -220,3 +228,130 @@ export async function createBackgroundTasks(
...
+      const instance = schedule.instances.at(0);
+      if (instance) {
+        await registerNextService.call(instance.id);
+      } else {
+        logger.error("Missing instance for declarative schedule", {
+          schedule,
+        });
+      }
+    } else {
+      const newSchedule = await prisma.taskSchedule.create({
```
ericallam, line 314: Use UpsertTaskSchedule here and handle failures

PR #1232, opened 2024-07-22
Run tags

matt-aitken:
You can trigger with tags:

```ts
const { id } = await simpleChildTask.trigger(
    { message: "Hello from triggerRunsWithTags" },
    { tags: "user:abcdef", "org:12345" }
  );
 ```
 
 And set them inside a run:
 
 ```ts
await tags.add("product:1");
```


apps/webapp/app/routes/api.v1.runs.$runId.tags.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const existingTags =
+      (await getTagsForRunId({
+        friendlyId: parsedParams.data.runId,
+        environmentId: authenticationResult.environment.id,
+      })) ?? [];
+
+    //remove duplicate tags from the new tags
+    const bodyTags = typeof body.data.tags === "string" ? [body.data.tags] : body.data.tags;
+    const newTags = bodyTags.filter((tag) => !existingTags.map((t) => t.name).includes(tag));
```
matt-aitken: Do the trim here, add it to the filter

packages/core/src/v3/apiClient/index.ts
```diff
@@ -289,6 +292,19 @@ export class ApiClient {
     );
   }
 
+  setTags(runId: string, body: SetTagsRequestBody, requestOptions?: ZodFetchOptions) {
```
matt-aitken: Change to addTags

packages/core/src/v3/schemas/api.ts
```diff
@@ -110,6 +122,12 @@ export const GetBatchResponseBody = z.object({
 
 export type GetBatchResponseBody = z.infer<typeof GetBatchResponseBody>;
 
+export const SetTagsRequestBody = z.object({
```
matt-aitken: AddTags

PR #1239, opened 2024-07-29
fix: normalize email case to prevent duplicate accounts

### LINKED ISSUE #1144: Logging in with the same email but different case creates a new account
If you login with test@gmail.com and Test@gmail.com you will create two accounts.

Kritik-J:
Closes #1144 

## ✅ Checklist

---

## Changelog

Normalize email case to prevent duplicate accounts

💯

### DIFF
--- a/apps/webapp/app/utils/email.ts
+++ b/apps/webapp/app/utils/email.ts
@@ -1,0 +1,5 @@
+export const normalizeEmail = (email: string): string => {
+  const [localPart, domain] = email.split("@");
+  const normalizedEmail = `${localPart}@${domain.toLowerCase()}`;
+  return normalizedEmail;
+};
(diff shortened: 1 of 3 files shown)


apps/webapp/app/utils/email.ts
```diff
@@ -0,0 +1,5 @@
+export const normalizeEmail = (email: string): string => {
+  const [localPart, domain] = email.split("@");
+  const normalizedEmail = `${localPart}@${domain.toLowerCase()}`;
```
rossanmol, line 3: Why not lowercase the whole email, but just domain?

Kritik-J: Refer - https://github.com/triggerdotdev/trigger.dev/issues/1144#issuecomment-2144628481

nicktrn:
We didn't forget about this, but we're not sure about changing the current behaviour. We'll also have to consider what happens to existing users with "duplicate" accounts that have different email casings.

PR #1243, opened 2024-08-01
Added TaskRun completedAt column

matt-aitken:
Add TaskRun `completedAt` column in preparation for some changes that are coming. It gets set when a TaskRun is in one of the finished statuses. We have not done a migration that sets this column for existing runs because that operation would be too expensive when deploying.

### DIFF
--- a/apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
+++ b/apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
@@ -566,7 +566,6 @@ export class SharedQueueConsumer {
                 lockedById: null,
                 status: lockedTaskRun.status,
                 startedAt: existingTaskRun.startedAt,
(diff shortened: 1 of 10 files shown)


apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
```diff
@@ -566,6 +566,7 @@ export class SharedQueueConsumer {
                 lockedById: null,
                 status: lockedTaskRun.status,
                 startedAt: existingTaskRun.startedAt,
+                //should we set the completedAt to null here?
```
matt-aitken: @ericallam @nicktrn should we set `completedAt` to null here?

nicktrn: Voting for `null`

PR #1250, opened 2024-08-08
FinalizeRunService

matt-aitken:
Centralize where we update the TaskRun to a final state and do marqs.ack.

In the future we will:
- enqueue run failure alerts from here
- update Redis concurrency for Task/global

### DIFF
--- a/apps/webapp/app/v3/failedTaskRun.server.ts
+++ b/apps/webapp/app/v3/failedTaskRun.server.ts
@@ -5,6 +5,7 @@ import { marqs } from "~/v3/marqs/index.server";
 import { TaskRunStatus } from "@trigger.dev/database";
 import { createExceptionPropertiesFromError, eventRepository } from "./eventRepository.server";
 import { BaseService } from "./services/baseService.server";
+import { FinalizeTaskRunService } from "./services/finalizeTaskRun.server";
 
 const FAILABLE_TASK_RUN_STATUSES: TaskRunStatus[] = ["EXECUTING", "PENDING", "WAITING_FOR_DEPLOY"];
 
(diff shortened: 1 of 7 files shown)


apps/webapp/app/v3/failedTaskRun.server.ts
```diff
@@ -40,7 +39,13 @@ export class FailedTaskRunService extends BaseService {
     // No more retries, we need to fail the task run
     logger.debug("[FailedTaskRunService] Failing task run", { taskRun, completion });
 
-    await marqs?.acknowledgeMessage(taskRun.id);
+    const finalizeService = new FinalizeTaskRunService();
```
ericallam, line 42: `BaseService` already has a constructor that takes a `PrismaClientOrTransaction`, so you should probably just use that instead of passing the transaction as an option to the call method.

apps/webapp/app/v3/services/completeAttempt.server.ts
```diff
@@ -50,17 +51,31 @@ export class CompleteAttemptService extends BaseService {
         id: execution.attempt.id,
       });
 
-      // Update the task run to be failed
-      await this._prisma.taskRun.update({
+      const run = await this._prisma.taskRun.findUnique({
```
ericallam: use findFirst here

apps/webapp/app/v3/services/completeAttempt.server.ts
```diff
@@ -96,28 +111,27 @@ export class CompleteAttemptService extends BaseService {
...
+          output: completion.output,
+          outputType: completion.outputType,
+          usageDurationMs: completion.usage?.durationMs,
+        },
+      });
 
-    await marqs?.acknowledgeMessage(taskRunAttempt.taskRunId);
+      await finalizeService.call({
+        tx,
+        id: taskRunAttempt.taskRun.id,
```
ericallam: `taskRunAttempt.taskRunId`

PR #1263, opened 2024-08-20
Fix for short runs inside (batch)triggerAndWait

matt-aitken:
If short runs were inside `triggerAndWait` or `batchTriggerAndWait` there was a race condition where a checkpoint wouldn't be created in time. In some cases the run wasn't running in the cluster anymore and they got stuck frozen forever.

Changes
- Don't attempt to continue these runs in the cluster if there's no checkpoint.
- When we create the checkpoint try and continue these runs (they won't continue if the sub-runs aren't finished).
- Remove some code that was failing attempts to prevent infinite recursion. It was causing errors in certain conditions where runs would have otherwise succeeded.

### DIFF
--- a/apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
+++ b/apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
@@ -617,16 +617,6 @@ export class SharedQueueConsumer {
           return;
         }
 
-        // if (messageBody.data.completedAttemptIds.length < 1) {
-        //   logger.error("No attempt IDs provided", {
-        //     queueMessage: message.data,
-        //     messageId: message.messageId,
-        //   });
-
-        //   await this.#ackAndDoMoreWork(message.messageId);
-        //   return;
-        // }
-
         const resumableRun = await prisma.taskRun.findUnique({
           where: {
             id: message.messageId,
--- a/apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
+++ b/apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
@@ -617,6 +617,16 @@ export class SharedQueueConsumer {
           return;
         }
 
+        // if (messageBody.data.completedAttemptIds.length < 1) {
+        //   logger.error("No attempt IDs provided", {
+        //     queueMessage: message.data,
+        //     messageId: message.messageId,
+        //   });
+
+        //   await this.#ackAndDoMoreWork(message.messageId);
+        //   return;
+        // }
+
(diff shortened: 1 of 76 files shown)


apps/webapp/app/v3/marqs/sharedQueueConsumer.server.ts
```diff
@@ -617,15 +617,15 @@ export class SharedQueueConsumer {
...
-        if (messageBody.data.completedAttemptIds.length < 1) {
-          logger.error("No attempt IDs provided", {
-            queueMessage: message.data,
-            messageId: message.messageId,
-          });
-
-          await this.#ackAndDoMoreWork(message.messageId);
-          return;
-        }
+        // if (messageBody.data.completedAttemptIds.length < 1) {
```
matt-aitken: Delete the code

apps/webapp/app/v3/services/createCheckpoint.server.ts
```diff
@@ -140,10 +151,72 @@ export class CreateCheckpointService extends BaseService {
...
+
+            return {
+              success: true,
+              checkpoint,
+              event: checkpointEvent,
+              keepRunAlive: false,
+            };
+          }
+
+          const childAttempt = await this._prisma.taskRunAttempt.findFirst({
```
matt-aitken: First check that the TaskRun is finished, get the most recent finished status attempt

apps/webapp/app/v3/services/createCheckpoint.server.ts
```diff
@@ -154,10 +227,33 @@ export class CreateCheckpointService extends BaseService {
...
+              id: true,
+            },
+            where: {
+              friendlyId: reason.batchFriendlyId,
+            },
+          });
+
+          if (!batchRun) {
+            logger.error("Batch not found", { friendlyId: reason.batchFriendlyId });
+            await marqs?.acknowledgeMessage(attempt.taskRunId);
```
matt-aitken: Remove ack and return success true similar to above

apps/webapp/app/v3/services/createCheckpoint.server.ts
```diff
@@ -140,10 +151,133 @@ export class CreateCheckpointService extends BaseService {
...
+              dependencyId: dependency.id,
+              childRunStatus: childRun.status,
+              childRunId: childRun.id,
+            });
+
+            return {
+              success: true,
+              checkpoint,
+              event: checkpointEvent,
+              keepRunAlive: true,
```
matt-aitken: Try this as false

PR #1265, opened 2024-08-20
v3: new build system

ericallam:
This PR overhauls the build system, as well as the `dev` and `deploy` commands.

### The major changes to the build system:

- Bundle by default (instead of external by default previously). `dependenciesToBundle` becomes `externals`.
- Outputs ESM instead of CJS
- Adds the ability to customize how building works using build extensions (including adding custom esbuild plugins)
- The `additionalFiles` config option is deprecated in favor of the `additionalFiles` build extension
- The `additionalPackages` config option is deprecated in favor of the `additionalPackages` build extension
- Added a `prisma` build extension to automatically include your schema.prisma file in deployed builds, as well as generating the client
- `resolveEnvVars` has been deprecated in favor of the `syncEnvVars` build extension
- Build extensions are also "stripped" from the built code, shrinking bundle size.
- Decrease cold-start times by dynamically importing just the task file that will be executed.
- Easily configure `jsx` settings to prevent issues with having to `import * as React from "react"`. `jsx` is set to automatic by default.

### Other major changes included in this PR:

- Add experimental support for using `bun` as a runtime, instead of just Node.js. The CLI still needs to execute with Node.js but tasks will run as a `bun` subprocess. Known issues include the lack of support for instrumentation of 3rd party SDKs.
- Support for decorators is not on by default, but can easily be added using the `emitDecoratorMetadata` build extension.
- Support for tasks located in files with the following extensions: (ts, tsx, mts, cts, js, jsx, mjs, cjs)
- Better debugging support in VSCode when running the `dev` command.
- More flexible configuration system through the adoption of [unjs c12](https://github.com/unjs/c12)
- Overhaul of the "indexing" system, which now happens at image build time, instead of at post-deployment.
- We now index task source file content to aide in future features that will help debugging

### Some big repo changes as well:

- removed all v2 exclusive packages from the repo
- moved `@trigger.dev/core-apps` and `@trigger.dev/core-backend` into `@trigger.dev/core` as subpath exports
- moved to using tshy/tsc to build packages
- packages are now type: module, but publish CJS and ESM variants
- vastly simplify the typescript setup, and support live types in the monorepo through import conditions
- upgrade `@opentelemetry-js/*` packages to the latest versions.
- add support for pkg.pr.new preview packages

### Still todo:

- [x] Ensure deploy errors during indexing are displayed correctly
- [ ] Fix version mismatch errors during deploy by downloading/executing the correct CLI version
- [x] Support for node-gyp during the install phase
- [x] Support adding additional system packages using extensions
- [ ] Improve `wasm` support
- [ ] Test, test, test


Aslemammad:
I love this! Trigger to the 🌕️

matt-aitken (approved):

gautamsi:
This breaks the config with `extraCACerts` introduced in #1214 

I see the config option and the code for injecting ca certs is still there but the option has been removed in `buildImage` params.

Also noticed that the file is not being copied from the config, potentially because of the `additionalFiles` being moved in extension.

I see the quickest option is to populate buildEnvVars which is not injecting anything yet. potentially missed from SyncEnvVars extension?

PR #1277, opened 2024-08-30
ElectricSQL run page and react-window-splitter

matt-aitken:
## ElectricSQL + run page

ElectricSQL syncs partial database replicas. This PR includes:
- The local setup and integration of it with Postgres (default to running on 3060 locally)
- The run page now live reloads the relevant spans and runs using Electric
- The inspector uses this data. For runs we fetch some additional data async from the server.

Generally this means this page is much more responsive now, requires far fewer network roundtrips and navigating between spans is way faster. The live reloading is also much more immediate.

https://github.com/user-attachments/assets/ef614586-cece-411f-9065-18f560849c3a

## react-window-splitter

This new package allows panels to be set with pixel widths and has better serialization: https://react-window-splitter-six.vercel.app/


ericallam:
What does the live reloading look like?

PR #1286, opened 2024-09-06
Frozen run fixes

matt-aitken:
Sometimes runs weren't continuing when their children were complete. This is for `triggerAndWait` and `batchTriggerAndWait`.

Changes
- Added a new service that gets given a run and figures out if there are dependencies and then schedules a resume if it makes sense
- In the finalize run service, set a final attempt status/error if passed in. Created a final attempt if there isn't one.
- If Graphile `addJob` throws an error then we return an undefined job instead of throwing.
- Don't resume the same batch multiple times

### DIFF
--- a/apps/webapp/app/v3/services/crashTaskRun.server.ts
+++ b/apps/webapp/app/v3/services/crashTaskRun.server.ts
@@ -69,6 +69,13 @@ export class CrashTaskRunService extends BaseService {
           },
         },
       },
+      attemptStatus: "FAILED",
+      error: {
+        type: "INTERNAL_ERROR",
+        code: "TASK_RUN_CRASHED",
+        message: opts.reason,
+        stackTrace: opts.logs,
+      },
     });
 
     const inProgressEvents = await eventRepository.queryIncompleteEvents(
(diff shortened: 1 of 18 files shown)


apps/webapp/app/v3/services/crashTaskRun.server.ts
```diff
@@ -146,12 +152,6 @@ export class CrashTaskRunService extends BaseService {
           }),
         },
       });
```
matt-aitken, line 154: If DEV then don't do the resuming, test in dev

apps/webapp/app/v3/services/resumeBatchRun.server.ts
```diff
@@ -98,23 +120,60 @@ export class ResumeBatchRunService extends BaseService {
...
+          dependentTaskAttempt: batchRun.dependentTaskAttempt,
+          checkpointEventId: batchRun.checkpointEventId,
+          hasCheckpointEvent: !!batchRun.checkpointEventId,
+        });
+      }
+    }
+  }
+
+  async #setBatchToCompletedOnce(batchRunId: string) {
+    const result = await this._prisma.batchTaskRun.updateMany({
```
matt-aitken, line 159: Are we missing any indexes with the new queries?

apps/webapp/app/v3/services/resumeDependentParents.server.ts
```diff
@@ -0,0 +1,267 @@
...
+          dependency,
+        }
+      );
+
+      // the child run isn't finished yet, so we can't resume the parent yet.
+      return {
+        success: true,
+        action: "not-finished",
+      };
+    }
```
matt-aitken: This can be removed now

matt-aitken:
Test cancelling from the dashboard and SDK

PR #1301, opened 2024-09-14
feat: disabling authentication for local testing

zeyaddeeb:
I just want to see if you guys are open to this. I wanted to test locally and auth is sometimes not required

## ✅ Checklist

---

## Testing

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,7 @@ APP_ORIGIN=http://localhost:3030
 ELECTRIC_ORIGIN=http://localhost:3060
 NODE_ENV=development
 V3_ENABLED=true
+AUTH_DISABLED=true
 
 # Redis is used for the v3 queuing and v2 concurrency control
 REDIS_HOST="localhost"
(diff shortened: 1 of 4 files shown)


matt-aitken:
@zeyaddeeb when you're developing locally you can sign in with any email address using magic link and it doesn't require confirmation. It instantly logs you in. This is based on NODE_ENV being "development": https://github.com/triggerdotdev/trigger.dev/blob/c1690bd1ae1df93def377265c917bddec2665551/apps/webapp/app/services/email.server.ts#L36

I think this achieves the same as what you want but also allows you to have many accounts which is really useful during development.

zeyaddeeb:
I wanted to kill auth all together, any ideas?

matt-aitken:
This approach alone won't work because you need the user to actually exist in the database for everything else to work.

I'm still not clear exactly on the use case for this given when running locally magic link just auto-logs you in without needing verification.

PR #1313, opened 2024-09-17
Add acking to RESUME_AFTER_DEPENDENCY message to the coordinator

matt-aitken:
We were just broadcasting `RESUME_AFTER_DEPENDENCY` to the coordinator before. If the coordinator crashed at the wrong time.

I've added a new message called `RESUME_AFTER_DEPENDENCY_WITH_ACK` which requires an ack to be sent back. If it's not sent back then we nack the message in the queue with a 5 second delay.

I've verified locally that this continues runs in normal situations, and if the coordinator is down.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new event handler for resuming tasks after dependencies are resolved, enhancing task coordination.
	- Added acknowledgment mechanism for message handling, improving error handling.
	- Added a new message type for resuming tasks with acknowledgment in the coordinator messages.

- **Improvements**
	- Enhanced control flow for task resumption, ensuring stricter management of dependencies and checkpoints.
	- Centralized callback structure for improved code reusability and maintainability.

### DIFF
--- a/apps/webapp/app/hooks/useSearchParam.ts
+++ b/apps/webapp/app/hooks/useSearchParam.ts
@@ -18,13 +18,13 @@ export function useSearchParams() {
         }
 
         if (typeof value === "string") {
-          search.set(param, encodeURIComponent(value));
+          search.set(param, value);
           continue;
         }
 
         search.delete(param);
         for (const v of value) {
-          search.append(param, encodeURIComponent(v));
+          search.append(param, v);
         }
       }
     },
--- a/apps/coordinator/src/index.ts
+++ b/apps/coordinator/src/index.ts
@@ -835,10 +835,16 @@ class TaskCoordinator {
             return;
           }
 
+          logger.log("WAIT_FOR_TASK checkpoint created", {
+            checkpoint,
+            socketData: socket.data,
+          });
+
           //setting this means we can only resume from a checkpoint
           socket.data.requiresCheckpointResumeWithMessage = `location:${checkpoint.location}-docker:${checkpoint.docker}`;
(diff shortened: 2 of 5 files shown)


apps/webapp/app/hooks/useSearchParam.ts
```diff
@@ -18,13 +18,13 @@ export function useSearchParams() {
...
         if (typeof value === "string") {
-          search.set(param, encodeURIComponent(value));
+          search.set(param, value);
           continue;
         }
 
         search.delete(param);
         for (const v of value) {
-          search.append(param, encodeURIComponent(v));
+          search.append(param, v);
```
matt-aitken, line 27: This fixes an unrelated issue with the run filters where they were being double uri encoded. This meant filters with weird characters in (like tasks with `/` weren't working with the filters).

apps/webapp/app/hooks/useSearchParam.ts
```diff
@@ -18,13 +18,13 @@
         }
 
         if (typeof value === "string") {
-          search.set(param, encodeURIComponent(value));
+          search.set(param, value);
```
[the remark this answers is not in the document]
matt-aitken: This isn't needed. URLSearchParams automatically does this when using `set` and then doing .toString() on the searchParams. I was doing it before and it was double encoding them which caused problems.

apps/webapp/app/hooks/useSearchParam.ts
matt-aitken: This isn't needed. URLSearchParams automatically does this when using `set` and then doing .toString() on the searchParams. I was doing it before and it was double encoding them which caused problems.

nicktrn (approved):

PR #1350, opened 2024-09-24
Puppeteer extension: set the PUPPETEER_EXECUTABLE_PATH env var

samejr:
With this change when launching puppeteer you don't need to set the executable path.


matt-aitken (approved):
Collaborated with @samejr on this

PR #1377, opened 2024-10-03
feat: Add maxDuration to tasks

ericallam:
You can now add a `maxDuration` to your tasks that will abort runs that exceed the duration, in CPU time only (e.g. it doesn't count when you are running `wait.for` or `triggerAndWait`, etc.). You can set `maxDuration` in 3 places:

- In `trigger.config.ts` to apply as a default to all tasks
- In your task definition, to apply to all runs of that task (which overrides the value in `trigger.config.ts` if there is one)
- When calling `trigger`, which overrides the previous two.

If a task run exceeds the maxDuration (only across a single attempt) then the run will fail with the following error:

CleanShot 2024-10-03 at 11 04 29@2x

## Release Notes

- **New Features**
	- Added a "Max duration" property to the run details, displaying the maximum duration of each run.
	- Introduced a new task run status, "TIMED_OUT," enhancing task run management and visibility.
	- Added new tasks with timing and retry features, improving task execution capabilities.
	- Introduced a new `TimedOutIcon` for visual representation of the "TIMED_OUT" status.

- **Improvements**
	- Enhanced handling for the "TIMED_OUT" status across various components, ensuring consistent representation and processing.
	- Improved data retrieval in the `SpanPresenter` and task run services to include maximum duration information.
	- Updated API schema to include a new optional field for maximum duration in task execution requests.

- **Bug Fixes**
	- Adjusted task run failure conditions to include the new "TIMED_OUT" status.


packages/cli-v3/src/entryPoints/deploy-index-worker.ts
```diff
@@ -115,6 +115,20 @@ if (typeof config.machine === "string") {
   });
 }
 
+// If the config has a maxDuration, we need to apply it to all tasks that don't have a maxDuration
```
matt-aitken, line 118: What if you wanted to have a default maxDuration but then have a specific task not have a maxDuration? It would be nice if you could pass `null` to say you don't want a limit. Or maybe some const we provide, like `MaxDuration.None` (that is just Math.Infinity).

matt-aitken (approved):

PR #1475, opened 2024-11-17
feat: exit if docker buildx can't be found for self-hosted builds

ghostdevv:
I ran into an issue when using trigger that I docker would fail to build as it couldn't find the `--progress` flag. This flag is only available when docker buildx is installed. This change adds in a check to see if docker buildx is installed, and if it's not it'll fail with a helpful message.

When buildx is installed the `docker buildx version` command will return a version, if it's not then the command will fail.

image

(`install-docker-buildx` isn't a real command, I just wrapped the arch linux `yay -S docker-buildx` for the demo)

We can then test the warning I added based on this

image

Additionally, I added a debug log for this. You can test that by running `trigger deploy --self-hosted --log-level debug`, and you'll get something like this at the end of the log:

image

**Edit**: added a better build message based on rabbit's suggestion

image

## ✅ Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced deployment command functionality with checks for Docker Buildx during self-hosted builds.
	- Added visual feedback with a spinner during the build process.

- **Bug Fixes**
	- Improved error handling for deployment failures and environment variable synchronization.

- **Chores**
	- Updated `tinyexec` dependency to version `^0.3.1`.
	- Updated `package.json` to version `3.2.0` and refined module export structure.

### DIFF
--- a/.changeset/spicy-turkeys-hug.md
+++ b/.changeset/spicy-turkeys-hug.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": minor
+---
+
+feat: exit if docker buildx can't be found for self-hosted builds
--- a/packages/cli-v3/package.json
+++ b/packages/cli-v3/package.json
@@ -118,7 +118,7 @@
     "std-env": "^3.7.0",
     "terminal-link": "^3.0.0",
     "tiny-invariant": "^1.2.0",
-    "tinyexec": "^0.2.0",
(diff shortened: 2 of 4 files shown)


nicktrn (approved):
Hey @ghostdevv - good spot! Thanks for this :pray: 

The tinyexec upgrade should probably be fine but I'll give it a quick spin after merging.

PR #1474, opened 2024-11-15
Full width tables

samejr:
- All pages with tables now have edge-to-edge tables
- Table.tsx styles updated to match the design
- Routes with tables on have been restructured to allow tables to scroll when needed, rather than the pages themselves.
- Added new progress action bars to the bottom of some pages to indicate progress towards e.g. number of alerts used
- Every page now has a docs button
- Runs page now has a link to the test page
- Some style updates to modals

## Release Notes

- **New Features**
	- Enhanced feedback component with a cancel option for user interaction.
	- Improved pagination controls with updated button icons for better clarity.
	- New `minimal` variant added to the InfoPanel for streamlined presentation.
	- Documentation links added across various components for easier access to resources.
	- Added a new `Spline` component for improved error display visuals.

- **UI Improvements**
	- Updated button styles and layouts across multiple components for consistency.
	- Enhanced visual feedback in tables and forms, including hover effects and spacing adjustments.
	- Improved error handling and user feedback in forms related to environment variables and schedules.
	- Refined layout and structure in the API keys and deployment management pages for better organization.
	- Adjusted layouts for better responsiveness and clarity in the task runs and schedules sections.

- **Bug Fixes**
	- Resolved layout issues in various components to ensure better responsiveness and user experience.

PR #1478, opened 2024-11-17
feat: always show alerts tab and lock email alerts when keys are missing

ghostdevv:
When I was testing my self hosted installation, I noticed that the alerts tab was missing. It was still accessible if I manually changed the url, and the webhook alerts were fully functional. This PR removes the _feature flag_ `alertsEnabled` and makes the alerts page always visible. Similarly to the Slack alert, it'll now show a notice to the user if they don't have their resend configured for alerts.

image

It has the same behaviour as before for checking whether email alerts are enabled. It uses the two environment variables. I've never used Remix before so let me know if it's not safe to check those there. I think it is as I did see there is other server code in that file, and the remix docs say that loaders run on the server.

If there is interest, I could also update the self hosting docs with how to configure email alerts in a follow up pr?

## ✅ Checklist

## Summary by CodeRabbit

- **New Features**
	- The "Alerts" menu item is now always visible in the side menu, enhancing accessibility.
	- The alert creation page now provides immediate feedback on email alert availability with a new input field or warning message.

- **Bug Fixes**
	- Simplified the control flow related to alert settings by removing unnecessary conditions.

- **Documentation**
	- Updated return values and properties in relevant functions to reflect the removal of the `alertsEnabled` feature.

### DIFF
--- a/apps/webapp/app/components/navigation/SideMenu.tsx
+++ b/apps/webapp/app/components/navigation/SideMenu.tsx
@@ -631,8 +631,6 @@ function V3ProjectSideMenu({
   project: SideMenuProject;
   organization: MatchedOrganization;
 }) {
-  const { alertsEnabled } = useFeatures();
-
   return (
     <>
       <SideMenuHeader title={"Project"} />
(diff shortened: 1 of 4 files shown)


ghostdevv:
I was worried there might be conflicts with #1474, but it looks like we're changing very different parts of the same file. So we should be ok!

PR #1488, opened 2024-11-21
Side help panel

samejr:
- Added a new side help menu that shows when you have >=1 task
- It doesn't display at all on the onboarding setup steps
- The open/close state of the panel persists in a cookie

Additional
- Button component now supports showing an icon after the shortcut
- Added an animated arrow component


apps/webapp/app/components/primitives/Buttons.tsx
```diff
@@ -258,13 +267,15 @@ export function ButtonContent(props: ButtonContentPropsType) {
...
-        {shortcut && (
-          <ShortcutKey
-            className={cn(shortcutClassName)}
-            shortcut={shortcut}
-            variant={variation.shortcutVariant ?? "medium"}
-          />
-        )}
+
+        {shortcut &&
+          (!props.shortcutPosition || props.shortcutPosition === "after-trailing-icon") && (
```
matt-aitken, line 272: It would be better to assign a default to `shortcutPosition` than do this. That way it will always have a value avoiding an undefined check here

matt-aitken (approved):

PR #1580, opened 2024-12-25
Add Support for Alternate Email Transports

### LINKED ISSUE #282: [TRI-951] Remove resend dependency from the webapp and replace with nodemailer and smtp

erin-allison:
Closes #282 
Supercedes #313 

## ✅ Checklist

- [x] I ran and tested the code works*

---

## Testing

Configured a local install for AWS SES and then SMTP, verifying each configuration properly sends a magic link email.

*I was unable to test Resend, as I do not have an account with them. The logic for Resend specifically, though, did not change, and it follows the same interface as the others, so there _shouldn't_ be any issues.

---

## Changelog

- Refactor `internal-packages/emails` to allow addition of alternative email transports (as opposed to just Resend API)
- Implement SMTP and AWS SES transports

---

## Screenshots

aws-ses
smtp

I have screen recordings of the testing saved, though I won't be sharing them publicly, as credentials (though they should be invalid by now) are visible. I am happy to provide them privately to the team for further validation, if necessary.

💯

--

## Breaking Change

This does introduce one small breaking change: currently to use Resend, all that needs to be specified is `RESEND_API_KEY`. As a result of needing to support varying config options, `EMAIL_TRANSPORT=resend` would need to be added to maintain the current behavior.

If this is not done, the application will default to the behavior of printing emails to the console, as if no email transport is configured.

This could be worked around by coding a special case for Resend due to existing deployments, but I didn't think it necessary unless was that deemed to be enough of a concern.

## Release Notes

- **New Features**
  - Enhanced email transport configuration options, including support for Resend, SMTP, and AWS SES.
  - Introduced a centralized email transport handling system for improved flexibility.
  - Added new mail transport classes for AWS SES, Resend, SMTP, and a null transport for logging.
  - New optional fields for email transport and alerting mechanisms in the environment schema.

- **Documentation**
  - Updated self-hosting guide with detailed instructions for email transport setup and configuration.
  - Enhanced clarity and usability of email configuration instructions.

- **Bug Fixes**
  - Improved clarity and usability of email configuration instructions.

### DIFF
--- a/apps/webapp/app/services/email.server.ts
+++ b/apps/webapp/app/services/email.server.ts
@@ -1,5 +1,6 @@
 import type { DeliverEmail, SendPlainTextOptions } from "emails";
 import { EmailClient } from "emails";
+import { MailTransportOptions } from "emails/transports";
 import type { SendEmailOptions } from "remix-auth-email-link";
 import { redirect } from "remix-typedjson";
 import { env } from "~/env.server";
@@ -13,7 +14,7 @@ const client = singleton(
   "email-client",
   () =>
     new EmailClient({
-      apikey: env.RESEND_API_KEY,
+      transport: buildTransportOptions(false),
       imagesBaseUrl: env.APP_ORIGIN,
       from: env.FROM_EMAIL ?? "team@email.trigger.dev",
       replyTo: env.REPLY_TO_EMAIL ?? "help@email.trigger.dev",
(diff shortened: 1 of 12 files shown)


erin-allison:
I'll be honest... I have no idea what's causing the typecheck to fail. I didn't know to run the typecheck command to validate, but now that I have run it, I get the same error locally.

```
Error: webapp:typecheck: app/services/email.server.ts(3,38): error TS2307: Cannot find module 'emails/transports' or its corresponding type declarations.
```

My IDE doesn't say anything's wrong and the code works. I'm guessing it's something small that I missed. Can someone have a look and point me in the right direction?

apps/webapp/app/services/email.server.ts
```diff
@@ -1,5 +1,6 @@
 import type { DeliverEmail, SendPlainTextOptions } from "emails";
 import { EmailClient } from "emails";
+import { MailTransportOptions } from "emails/transports";
```
matt-aitken: This causes a typecheck error because it's trying to access something not exported from the main file and it's not specified as a separate export in the package.json file.

I'd just re-export `MailTransportOptions` from the index file and change this import to be from `emails`.

matt-aitken (changes requested):
There's a TypeScript error when `pnpm run typecheck --filter webapp` is run. I've added a comment about why it's happening and an easy fix.

Having the default transport as undefined makes sense. It means people upgrading to this version will have to set some things to get alert emails to work again which is a bit annoying but I think on balance it's the right call because it's a better default for the future.

This is great, thanks!

apps/webapp/app/services/email.server.ts
erin-allison: Gah, there it is... thank you!

matt-aitken (approved):
Brilliant!

PR #1618, opened 2025-01-16
Throw an error for concurrent waits (with nice docs link)

matt-aitken:
## Problem

In the current version, you can't perform more that one "wait" in parallel.

Waits include:
- `wait.for()`
- `wait.until()`
- `task.triggerAndWait()`
- `task.batchTriggerAndWait()`
- And any of our functions with `wait` in the name.

This restriction exists because we suspend the task server after a wait, and resume it when the wait is done. At the moment, if you do more than one wait, the run will never continue when deployed, so we throw this error instead.

The most common situation this happens is if you're using `Promise.all` around some of our wait functions. Instead of doing this use our built-in functions for [triggering tasks](/triggering#triggering-from-inside-another-task). We have functions that allow you to trigger different tasks in parallel.

## Solution

Up until now we had some docs about this, but we didn't actually do anything at runtime to stop you doing this.

This PR adds an error when we detect two waits are called simultaneously. It has a link to a new section of our troubleshooting docs.

It also adds a nicer way of preventing retrying runs and hydrating errors into JSON.

## Summary by CodeRabbit

- **Documentation**
	- Added troubleshooting guidance for parallel wait operations
	- Clarified limitations of concurrent wait functions
	- Explained why multiple wait operations are not supported

- **New Features**
	- Introduced error handling for concurrent wait scenarios
	- Added specific error code for parallel wait attempts

- **Bug Fixes**
	- Prevented multiple simultaneous wait operations across runtime managers
	- Enhanced error reporting for concurrent wait issues

### DIFF
--- a/docs/troubleshooting.mdx
+++ b/docs/troubleshooting.mdx
@@ -80,6 +80,21 @@ Your code is deployed separately from the rest of your app(s) so you need to mak
 
 Prisma uses code generation to create the client from your schema file. This means you need to add a bit of config so we can generate this file before your tasks run: [Read the guide](/config/config-file#prisma).
 
+### `Parallel waits are not supported`
+
+In the current version, you can't perform more that one "wait" in parallel.
+
+Waits include:
+- `wait.for()`
+- `wait.until()`
+- `task.triggerAndWait()`
+- `task.batchTriggerAndWait()`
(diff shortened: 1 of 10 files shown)


nicktrn (approved):
:+1:

PR #1683, opened 2025-02-09
Expandable code blocks

samejr:
This adds a new "Expand" icon next to the copy button on all code blocks

CleanShot 2025-02-09 at 13 08 22

It opens the code in a large modal so you can view it more clearly

CleanShot 2025-02-09 at 13 09 27

It's default visible for all code blocks but can be turned off.

## Summary by CodeRabbit

- **New Features**
	- Introduced a modal view for code blocks, enabling users to open and interact with code in a dedicated popup with viewing and copy options.
- **Style**
	- Refined the appearance of modal titles for improved text readability and a cleaner user interface.


apps/webapp/app/components/code/CodeBlock.tsx
```diff
@@ -193,6 +200,9 @@ export const CodeBlock = forwardRef<HTMLDivElement, CodeBlockProps>(
   ) => {
     const [mouseOver, setMouseOver] = useState(false);
     const [copied, setCopied] = useState(false);
+    const [modalCopied, setModalCopied] = useState(false);
```
matt-aitken, line 203: Can't we use the existing copied state?

apps/webapp/app/components/code/CodeBlock.tsx
```diff
@@ -206,6 +216,19 @@ export const CodeBlock = forwardRef<HTMLDivElement, CodeBlockProps>(
       [code]
     );
 
+    const onModalCopied = useCallback(
```
matt-aitken, line 219: Can't we use the existing copy callback?

matt-aitken (changes requested):
It's hard to tell on my phone but are we duplicating all the syntax highlighting code? If so we should pull that out until a separate non-exported component.

apps/webapp/app/components/code/CodeBlock.tsx
samejr: I tried this at first but when you do, it triggers the copied UI to animated below the modal. You can see it through the modal's overlay and it looks like a bug.

matt-aitken (approved):

Arc-a-it:
@matt-aitken     I ran a **CLI experiment** on a large JS/TS repo.
A small, reasonable change was **BLOCKED** due to architectural blast radius.  
The decision was **arguable**, but it forced the question.

I wrote up the exact collision here :  
👉 [**[link to your arcvision-collision-proof repo]**](https://github.com/Arc-a-it/arcvision-collision-proof)

Curious if this would be unacceptable or useful in your world.

PR #1692, opened 2025-02-11
Docker compose migration

rharkor:
## ✅ Checklist

---

## Testing

- Checked the installation of the new docker compose version following the DOCKER_INSTALLATION.md
- Ran the commands `pnpm run docker` / `pnpm run docker:stop`

---

## Changelog

Replaced the old `docker-compose` version with the new `docker compose` command since the old one is deprecated, according to:
- https://docs.docker.com/compose/install/
- pnpm run build --filter webapp

---

💯

## Summary by CodeRabbit

- **Documentation**
  - Updated Docker Compose installation instructions to use the new Docker CLI plugin method with improved setup and verification steps.  
- **Chores**
  - Revised Docker-related script commands to adopt the updated `docker compose` syntax for managing services.


nicktrn:
Hey @rharkor - thanks for this! We should really get this in. Would be good if this didn't break existing dev setups.

Have you tested what happens to existing containers? Specifically:
1. The user ran docker using the old `docker-compose` command
2. The user runs it again with these changes, i.e. with `docker compose`
3. What happens to the existing resources? Are there any errors or duplicates?

I think I ran into issues with this previously. Would be great to know what happens before we get a chance to test the new instructions myself. Thanks again :pray:

rharkor:
Hey @nicktrn following the documentation it seems that you will need to stop every container before restarting theme because compose v1 use "_" and compose v2 use "-" as separator

https://docs.docker.com/compose/releases/migrate/

nicktrn:
Ah perfect, that's what I was looking for!

PR #1736, opened 2025-02-26
Remove Bun version warning

0x77dev:
## ✅ Checklist

---

## Changelog

- [Remove Bun version warning because issue was resolved](https://github.com/oven-sh/bun/issues/13799)

## Summary by CodeRabbit

- **Documentation**
  - Removed outdated warnings about specific Bun version requirements.
  - Updated guidance to streamline the setup process for Trigger.dev in Bun projects.
  - Introduced experimental support for Bun, simplifying the setup instructions.

### DIFF
--- a/docs/guides/frameworks/bun.mdx
+++ b/docs/guides/frameworks/bun.mdx
@@ -9,8 +9,6 @@ import Prerequisites from "/snippets/framework-prerequisites.mdx";
 import CliRunTestStep from "/snippets/step-run-test.mdx";
 import CliViewRunStep from "/snippets/step-view-run.mdx";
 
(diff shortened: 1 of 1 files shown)


ericallam:
@0x77dev is bun working for you?

HansKristoffer:
At this place https://github.com/triggerdotdev/trigger.dev/pull/1409/files it should also use the newest Bun version. It is right now difficult to work with Bun in Trigger.Dev because it is not using a new version of Bun.

nicktrn:
We should definitely use a more recent version and maybe make this configurable.

The reason this isn't quite as simple is that when we tested newer releases recently, there were still issues once deployed on our infra. Will have to give this another spin when we get a chance.

nicktrn:
We decided to remove our experimental Bun support entirely. Node will be the only supported runtime on v4+.

- https://github.com/triggerdotdev/trigger.dev/pull/2048

It was never well tested and there wasn't much demand in the community. Version updates became a problem due to bugs and incompatibilities, e.g. OTel and checkpoints. There were additional challenges with the node+bun image we had to use given our entry point only supports node and has to spawn bun processes.

PR #1737, opened 2025-02-26
Only allow a single dev queue consumer to dequeue at once

ericallam:


apps/webapp/app/v3/marqs/devQueueConsumer.server.ts
```diff
@@ -252,6 +262,37 @@ export class DevQueueConsumer {
...
+        logger.debug("Connection lost for more than 60 seconds, stopping the consumer", {
+          env: this.env,
+        });
+
+        await this.stop("Connection lost for more than 60 seconds");
+        return;
+      }
+
+      setTimeout(() => this.#doWork(), 1000);
+    }
```
matt-aitken, line 281: Do you need a `return` here to stop it falling through?

matt-aitken (changes requested):
Missing a `return`

matt-aitken (approved):

PR #1749, opened 2025-03-04
Fix email alert failures because of newlines in error messages (fixes #1745)

### LINKED ISSUE #1745: [TRI-4669] Run alerts don't send if the run error message has a newline character in it
We put the error.message in Run failed alert email subjects:

https://github.com/triggerdotdev/trigger.dev/blob/e0241819a401a86c7d5cfb2f55c25a58aa4e95ee/internal-packages/emails/src/index.tsx#L112

If the error.message has a newline in it email APIs reject it and we end up with this error in the queue we use to send them:

```json
{"visibilityTimeoutMs":60000,"item":{"alertId":"cmxxxxxxxxxxxxxxx"},"attempt":2,"errorMessage":"The `\\n` is not allowed in the `subject` field.","job":"v3.deliverAlert"}
```

It will try again multiple times but ultimately fail because the subject will always be invalid.

## Solution
We need to strip out investigate what characters are invalid and then strip them all out of email subjects. It would probably be best to do this at a level in the code where future developers won't have to think about this problem.

My recommendation is to do the string replacements on the `subject` here: https://github.com/triggerdotdev/trigger.dev/blob/e0241819a401a86c7d5cfb2f55c25a58aa4e95ee/internal-packages/emails/src/index.tsx#L66

<sub>[TRI-4669](https://linear.app/triggerdotdev/issue/TRI-4669/run-alerts-dont-send-if-the-run-error-message-has-a-newline-character)</sub>

ericallam:


internal-packages/emails/src/index.tsx
```diff
@@ -128,3 +130,12 @@ export class EmailClient {
...
   }
 }
+
+function formatErrorMessageForSubject(message?: string) {
+  if (!message) {
+    return "";
+  }
+
+  const singleLine = message.replace(/[\r\n]+/g, " ");
+  return singleLine.length > 30 ? singleLine.substring(0, 27) + "..." : singleLine;
```
nicktrn, line 140: @samejr will shout at you for not using ellipsis

nicktrn (approved):

PR #1764, opened 2025-03-06
feat: playwright extension

MendyLanda:
Closes #<issue>

## ✅ Checklist

---

## Testing
Made sure it works with all browsers 

---

## Screenshots
image

## Summary by CodeRabbit

- **New Features**
  - Introduced a browser automation extension supporting Playwright with Chromium, Firefox, and WebKit.
  - Added an automated task for browser testing, including webpage navigation and screenshot capture.
  - Enhanced build and configuration settings to support Playwright integration and streamline automation workflows.
  - Added Playwright as a new dependency to expand browser automation options.
  - Provided comprehensive documentation for Playwright integration, including setup and task lifecycle management.


Davidkle:
when will this merge?

nicktrn:
Thanks a lot for this @MendyLanda! Will take a look very soon :crossed_fingers: (will have to add some tests for image sizes)

@Davidkle sorry for the late reply. The great thing about [build extensions](https://trigger.dev/docs/config/extensions/overview) is you don't need to wait for them to merge. You can write your own in your local project. In this case, just save [playwright.ts](https://github.com/triggerdotdev/trigger.dev/blob/b0bf62441bdf1e334ede6c67d12c6ee4ec7c5586/packages/build/src/extensions/playwright.ts) and then import and use the extension as usual.

MendyLanda:
@nicktrn i can write these tests, but i need to know what you're looking for. 

Is there a limit on the build image size that we shouldn't surpass? I haven't seen it documented anywhere

Also, fyi the image size will be very close to what you'd get if running `npx playwright install --with-deps` would've been possible. Actually it most likely be smaller since the npx command installs all browsers by default, while we force the user to specify what browsers they need, or just install chromium. 

Or perhaps you want to alert the user when the build image is over a certain size?

nicktrn (approved):
Thanks again for this @MendyLanda!

I decided to only check image sizes locally and they're looking good as you said. Around ~600MB to add chromium. Added docs and middleware example to make this play nicely with checkpoints.

PR #1817, opened 2025-03-25
v4: New lifecycle hooks

ericallam:
This PR revamps the lifecycle hooks system, including a new way of defining global hooks, better middleware with locals support, new hooks such as `onWait`, `onResume`, and `onComplete`, and a new `init.ts` file that is automatically loaded on every run.

## Breaking changes

The params passed to the existing hooks have been consolidated into a single param object.

Before:

```ts
export const taskWithOnStart = task({
  id: "task-with-on-start",
  onStart: async (payload, { ctx }) => {
    //...
  },
  run: async (payload: any, { ctx }) => {
    //...
  },
});
```

After:

```ts
export const taskWithOnStart = task({
  id: "task-with-on-start",
  onStart: async ({ payload, ctx }) => {
    //...
  },
  run: async (payload: any, { ctx }) => {
    //...
  },
});
```

## Deprecations

These are marked as deprecated but will continue to work:

- `init` is deprecated, in favor of middleware and locals
- `cleanup` is deprecated for the same reason
- `handleError` is deprecated, use `catchError` instead

## New global hooks support

Previously, you could only define a single global lifecycle hook in the `trigger.config.ts` file, but this meant that you'd have to import any dependencies you needed in the hooks into the config file, which could cause issues because it sometimes broke the config loading.

Now you can define one or more global hooks anywhere in your codebase, like so:

```ts
import { logger, tasks } from "@trigger.dev/sdk";

tasks.onSuccess(async ({ ctx, payload, output }) => {
  logger.info("Hello, world from the success", { ctx, payload });
});
```

## Improved middleware and locals

Previously, if you wanted to initialize some resource (e.g. a database client) and provide it to your tasks, you'd have to use the `init` hook and return the client, which you could then access in your `run` function and other lifecycle hooks:

```ts
export const taskWithInitReturn = task({
  id: "task-with-init-return",
  init: async (payload, { ctx }) => {
    return { someData: "someValue" };
  },
  run: async (payload: any, { ctx, init }) => {
    console.log(init.someData); // "someValue"
  },
});
```

This works okay when passing the init option when creating a task, but the types don't really work if the `init` is defined globally. So, we've improved the middleware and added a type-safe "locals" system to be able to easily provide resources/clients, etc. to all your tasks and lifecycle hooks:

```ts
import { locals } from "@trigger.dev/sdk";
import { logger, tasks } from "@trigger.dev/sdk";

// Define the DbLocal key which is typed to the correct client type
const DbLocal = locals.create<{ connect: () => Promise<void>; disconnect: () => Promise<void> }>(
  "db"
);

// Use this getter to safely access the db client
export function getDb() {
  return locals.getOrThrow(DbLocal);
}

tasks.middleware("db", async ({ ctx, payload, next, task }) => {
  // this will run before ANY other lifecycle hooks and run function
  const db = locals.set(DbLocal, {
    connect: async () => {},
    disconnect: async () => {},
  });

  await db.connect();

  // This will call the next middleware
  await next();

  // Now you can cleanup your database connection here after everything else has finished
  await db.disconnect();
});
```

## New hooks

We've introduced three new hooks

### onWait

This hook will be called right before a wait is starting, such as `wait.for({ seconds: 5 })`, `wait.forToken(token)`, or `yourTask.triggerAndWait`.

You can use this hook to cleanup resources should you need to, before starting a possibly long wait.

### onResume

This hook is similar to onWait, but will be called after the wait has finished and before the rest of your code resumes running.

### onComplete

This hook is called when a run has completed, either as a successful or failing run.

## init.ts file

If you create an `init.ts` file in the root of your trigger directory, we'll automatically load this file before loading your task file when executing a task. This is handy as a place to put any global initialization code that you want to run before every task, without having to stick it in the `trigger.config.ts` file.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new flag to streamline test command usage.
	- Expanded lifecycle hook and middleware support for enhanced task management.
	- Added optional initialization entry point detection for improved startup flows.
	- Extended visual feedback with additional icons for varied task statuses.
	- Added new task lifecycle hooks for improved task execution management.
- **Bug Fixes**
	- Consolidated error reporting for clearer handling during task execution.
- **Tests**
	- Implemented comprehensive tests to ensure robust task execution and lifecycle management.
- **Documentation**
	- Updated testing examples to reflect the new command flag usage.


apps/webapp/app/components/runs/v3/RunIcon.tsx
```diff
@@ -28,6 +28,8 @@ type SpanNameIcons = {
 const spanNameIcons: SpanNameIcons[] = [{ matcher: /^prisma:/, iconName: "brand-prisma" }];
 
 export function RunIcon({ name, className, spanName }: TaskIconProps) {
+  console.log("spanName", spanName, name);
+
```
matt-aitken: @ericallam 👀

packages/cli-v3/src/build/bundle.ts
```diff
@@ -229,11 +230,24 @@ export async function getBundleResultFromBuild(
...
+  let initEntryPoint: string | undefined;
 
   const configEntryPoint = resolvedConfig.configFile
     ? relative(resolvedConfig.workingDir, resolvedConfig.configFile)
     : "trigger.config.ts";
 
+  // Check if the entry point is an init.ts file at the root of a trigger directory
+  function isInitEntryPoint(entryPoint: string): boolean {
+    const normalizedEntryPoint = entryPoint.replace(/\\/g, "/"); // Normalize path separators
+    const initFileName = "init.ts";
```
matt-aitken: Should we support `js` `jsx` and `tsx` files?

packages/cli-v3/src/build/bundle.ts
```diff
@@ -229,11 +230,24 @@ export async function getBundleResultFromBuild(
...
+
+    // Check if it's directly in one of the trigger directories
+    return resolvedConfig.dirs.some((dir) => {
+      const normalizedDir = dir.replace(/\\/g, "/");
+      return normalizedEntryPoint === `${normalizedDir}/${initFileName}`;
```
matt-aitken: is this safe on Windows?

PR #1854, opened 2025-03-31
cli: add dev lock file to prevent 2 dev processes running at the same time in the same dir

ericallam:
## Summary by CodeRabbit

- **New Features**
  - Introduced an automatic lock file management system for development sessions. A lock file is created at the start and reliably removed when the session stops, ensuring improved session stability and reducing potential conflicts.

### DIFF
--- a/packages/cli-v3/src/dev/lock.ts
+++ b/packages/cli-v3/src/dev/lock.ts
@@ -1,0 +1,87 @@
+import path from "node:path";
+import { readFile } from "../utilities/fileSystem.js";
+import { tryCatch } from "@trigger.dev/core/utils";
+import { logger } from "../utilities/logger.js";
+import { writeFile } from "node:fs/promises";
+import { existsSync, unlinkSync } from "node:fs";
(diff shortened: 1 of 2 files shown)


packages/cli-v3/src/dev/lock.ts
```diff
@@ -0,0 +1,87 @@
...
+import { logger } from "../utilities/logger.js";
+import { writeFile } from "node:fs/promises";
+import { existsSync, unlinkSync } from "node:fs";
+import { onExit } from "signal-exit";
+
+const LOCK_FILE_NAME = "dev.lock";
+
+export async function createLockFile(cwd: string) {
+  const currentPid = process.pid;
+  const lockFilePath = path.join(cwd, ".trigger", LOCK_FILE_NAME);
```
matt-aitken, line 13: We should check the `.trigger` folder exists first, and create it if it doesn't.

matt-aitken (changes requested):

matt-aitken (approved):

PR #1906, opened 2025-04-09
Fix for queue indexing not updating concurrencyLimit/removing it

matt-aitken:
When updating a queue the concurrencyLimit wasn't getting updated/removed. Made sure that `releaseConcurrencyOnWaitpoint` is set correctly to false if it's omitted (in an update)


nicktrn (approved):
Who doesn't like nulls?

PR #1962, opened 2025-04-22
Allow empty payloads in test runs

myftija:
Changes in this PR:

- adapted the code editor component to allow empty inputs
- exposed basic linter configs in the code editor component
- applied a couple of light touch ups to the error display elements in favor of consistency

## Screenshots

Previously:
Screenshot 2025-04-21 at 17 20 37

Now:
Screenshot 2025-04-21 at 17 21 17

## Summary by CodeRabbit

- **New Features**
  - Added configurable linting options to the JSON editor, allowing users to enable or disable linting and permit empty content without errors.

- **Style**
  - Improved the visual appearance of linting errors and diagnostic messages in the editor for better readability.

- **Bug Fixes**
  - Enhanced validation error messages for JSON fields, providing clearer feedback when invalid JSON is entered.

- **Chores**
  - Removed placeholder text from the JSON editor in the task form for a cleaner interface.

### DIFF
--- a/apps/webapp/app/v3/testTask.ts
+++ b/apps/webapp/app/v3/testTask.ts
@@ -4,46 +4,42 @@ export const TestTaskData = z
   .discriminatedUnion("triggerSource", [
     z.object({
       triggerSource: z.literal("STANDARD"),
-      payload: z.string().transform((payload, ctx) => {
-        try {
-          const data = JSON.parse(payload);
-          return data as any;
-        } catch (e) {
-          console.log("parsing error", e);
-
-          if (e instanceof Error) {
-            ctx.addIssue({
-              code: z.ZodIssueCode.custom,
-              message: e.message,
-            });
-          } else {
-            ctx.addIssue({
-              code: z.ZodIssueCode.custom,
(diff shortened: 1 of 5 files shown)


apps/webapp/app/v3/testTask.ts
```diff
@@ -4,46 +4,42 @@ export const TestTaskData = z
...
-          const data = JSON.parse(payload);
-          return data as any;
-        } catch (e) {
-          console.log("parsing error", e);
+      payload: z
+        .string()
+        .optional()
+        .transform((val, ctx) => {
+          if (!val) {
+            return {};
```
myftija, line 12: Returning `undefined` here would be more fitting, but it breaks the replay flow, which doesn't handle this well. The fix seemed a bit involved, so leaving this for a separate PR. In the meantime, the behavior stays the same, as users would still pass `{ }` as empty payload for tests.

matt-aitken: Good idea

matt-aitken (approved):

PR #1987, opened 2025-04-28
Fix managed run controller edge cases

nicktrn:
- ensure previous executions are always killed
- improve websocket disconnect logs
- protect against unexpected attempt number changes
- don't handle snapshot changes when shutting down
- prevent parallel snapshot polling intervals
- improve child process preparedness checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved issues with polling interval resets to prevent duplicate intervals and overlapping executions.
  - Fixed problems where run executions could become stuck or duplicated after warm starts, ensuring proper cleanup.
  - Added safeguards against unexpected changes in execution attempt numbers.

- **New Features**
  - Improved lifecycle management for run executions, including enhanced tracking and validation of execution attempts.
  - Enhanced debug logging for better visibility during execution and socket disconnect events.

- **Refactor**
  - Streamlined internal state management for polling and execution processes, improving reliability and idempotency.


packages/cli-v3/src/entryPoints/managed/controller.ts
```diff
@@ -486,10 +497,32 @@ export class ManagedRunController {
...
+            description: description.toString(),
+          };
+        }
+
+        return {
+          description: description.description,
+          context: description.context ? String(description.context) : undefined,
+        };
+      };
+
```
matt-aitken, line 521: Not important for this PR, but for things like this I think it would be nicer for this inline function to be a top-level util function. Then the main bit of the code is easier to understand, and you only need to dig into this if you want to go elsewhere.

matt-aitken: Unless there's a good reason I'm not a big fan of inline functions because I find it makes understanding the main logic harder.

nicktrn: Yeah, agreed. The types suck and weren't exported by socket.io

matt-aitken (approved):

PR #1995, opened 2025-04-29
Post new sign up reasons to slack, not plain

samejr:
When a new org signs up and completes the "What problem are you trying to solve?" the response is posted to Slack instead of Plain.

It looks like this: 
CleanShot 2025-04-29 at 18 45 45@2x

## Summary by CodeRabbit

- **New Features**
  - Introduced Slack integration to notify a Slack channel when a new organization is created. The notification includes the organization name, reason for joining, and creator's email.

- **Chores**
  - Updated Slack-related environment variables to be optional, allowing more flexible configuration.


apps/webapp/app/env.server.ts
```diff
@@ -714,6 +714,9 @@ const EnvironmentSchema = z.object({
 
   QUEUE_SSE_AUTORELOAD_INTERVAL_MS: z.coerce.number().int().default(5_000),
   QUEUE_SSE_AUTORELOAD_TIMEOUT_MS: z.coerce.number().int().default(60_000),
+
+  SLACK_BOT_TOKEN: z.string(),
+  SLACK_SIGNUP_REASON_CHANNEL_ID: z.string(),
```
matt-aitken: @samejr these need to be optional because of self-hosters

apps/webapp/app/services/slack.server.ts
```diff
@@ -0,0 +1,40 @@
...
+const slack = new WebClient(env.SLACK_BOT_TOKEN);
+
+type SendNewOrgMessageParams = {
+  orgName: string;
+  whyUseUs: string;
+  userEmail: string;
+};
+
+export async function sendNewOrgMessage({ orgName, whyUseUs, userEmail }: SendNewOrgMessageParams) {
+  try {
```
matt-aitken, line 17: If either of the environment variables aren't set then we would just return here.

matt-aitken (changes requested):
The env vars need to be optional

PR #2006, opened 2025-05-01
Fix controller waitpoint resolution, suspendable state, and snapshot race conditions

nicktrn:
Quite a few things in here:
- Fix testcontainer unit tests (first attempt..)
- Remove a lot of dead code
- Add snapshot queue tests
- Add debug logs to runtime manager
- Fix entitlement validation when no client exists
- Correctly resolve waitpoints that come in early
- Ensure correct state before requesting suspension
- Fix race conditions in snapshot processing
- New snapshots since engine method and webapp route
- Add run execution deprecation detection
- Update connection settings after restores
- Fix restore detection (moved to snapshot manager)

We now require two things before suspending a run:
1. The execution status needs to be either `EXECUTING_WITH_WAITPOINTS` or `QUEUED_EXECUTING`
2. The runtime manager needs to signal it's "suspendable"
3. These events can happen out of order

## Summary by CodeRabbit

- **New Features**
  - Added ability to fetch execution snapshots since a given snapshot, with new API routes and client methods in backend and webapp.
  - Introduced a `SnapshotManager` and `RunNotifier` to centralize snapshot state and notification handling.
  - Added a new `ConsoleRunLogger` and enhanced debug logging capabilities.
  - Deployment logs now always print fully in CI environments.

- **Bug Fixes**
  - Improved handling of early-arriving waitpoints and race conditions during snapshot processing.
  - Ensured proper state before suspension requests and fixed concurrency issues.

- **Refactor**
  - Replaced `ManagedRuntimeManager` with `SharedRuntimeManager` for better waitpoint management.
  - Simplified event and IPC message handling by removing obsolete wait-related events and messages.
  - Refactored snapshot polling to delegate handling externally and improved internal metrics tracking.
  - Enhanced environment variable overrides and socket event handling with improved error and disconnect management.

- **Tests**
  - Added extensive tests for `SnapshotManager` and snapshot retrieval since a given snapshot.
  - Added tests verifying `getSnapshotsSince` behavior and error handling.

- **Chores**
  - Updated TypeScript configs to exclude test files from source builds and added a test config.
  - Improved error output in publishing scripts.
  - Enhanced deployment command output formatting for CI and non-CI environments.

- **Documentation**
  - Updated change logs to reflect CI logging behavior changes.


packages/cli-v3/src/entryPoints/managed/execution.ts
```diff
@@ -144,6 +146,14 @@ export class RunExecution {
...
 
+    taskRunProcess.onSendDebugLog.attach(async (debugLog) => {
+      this.sendRuntimeDebugLog(debugLog.message, debugLog.properties);
+    });
+
+    taskRunProcess.onSetSuspendable.attach(async ({ suspendable }) => {
+      this.suspendable = suspendable;
+    });
+
     return taskRunProcess;
```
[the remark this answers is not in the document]
matt-aitken: @nicktrn is this a real concern?

packages/cli-v3/src/entryPoints/managed/execution.ts
```diff
@@ -943,14 +863,40 @@ export class RunExecution {
...
+        runId: this.runFriendlyId,
+        snapshotId: this.currentSnapshotFriendlyId,
+        executionId: this.id,
+        executionRestoreCount: this.restoreCount,
+        lastHeartbeat: this.lastHeartbeat?.toISOString(),
+      },
+    });
+  }
+
+  private set suspendable(suspendable: boolean) {
```
matt-aitken, line 954: If this is called and it does fail (the catch fires), should we return false from the function?

matt-aitken (approved):

packages/cli-v3/src/entryPoints/managed/execution.ts
nicktrn: I think it's fine to just log this for now, worst case we won't checkpoint. Should only happen during cleanup in very rare cases - we won't want to suspend then anyway.

packages/cli-v3/src/entryPoints/managed/execution.ts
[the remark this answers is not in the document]
nicktrn: rabbit has a point, have added a new detach method to taskRunProcess

internal-packages/run-engine/src/engine/systems/executionSnapshotSystem.ts
```diff
@@ -141,6 +154,72 @@ export function executionResultFromSnapshot(snapshot: TaskRunExecutionSnapshot):
...
+    include: {
+      completedWaitpoints: true,
+      checkpoint: true,
+    },
+    orderBy: { createdAt: "asc" },
+  });
+
+  return snapshots.map(enhanceExecutionSnapshot);
+}
+
```
matt-aitken: This makes me think we should have an index with the id instead of createdAt (or as well as)

internal-packages/run-engine/src/engine/systems/executionSnapshotSystem.ts
```diff
@@ -141,6 +154,72 @@ export function executionResultFromSnapshot(snapshot: TaskRunExecutionSnapshot):
...
+  };
+}
+
+export async function getExecutionSnapshotsSince(
+  prisma: PrismaClientOrTransaction,
+  runId: string,
+  sinceSnapshotId: string
+): Promise<EnhancedExecutionSnapshot[]> {
+  // Find the createdAt of the sinceSnapshotId
+  const sinceSnapshot = await prisma.taskRunExecutionSnapshot.findUnique({
```
matt-aitken: Try and avoid findUnique everywhere, it's an easier rule to follow and means if we add any joins later we don't need to remember.

internal-packages/run-engine/src/engine/systems/executionSnapshotSystem.ts
nicktrn: Should we just use IDs for everything instead of createdAt?

internal-packages/run-engine/src/engine/systems/executionSnapshotSystem.ts
nicktrn: We really need an eslint rule to help with this.. one always slips through the cracks somehow

matt-aitken (approved):

PR #2049, opened 2025-05-14
fix: zod boolean coercion should correctly parse "true" and "false"

ghostdevv:
~~Closes #<issue>~~

## ✅ Checklist

---

## Testing

I was trying to setup SMTP with my self-hosted trigger.dev instance, and I kept running into the following error:

```
Failed to send email to [...], Magic sign-in link for Trigger.dev. Error Error: 00C83842D37D0000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:354:
```

After some debugging I realised it was because `SMTP_SECURE=false` or `SMTP_SECURE="false"` both get parsed by `z.coerce.boolean()` as `true`. The only way to succesfully get this to work, without this patch, is to completely omit the `SMTP_SECURE` variable such that the result is `undefined` (presumably) thanks to the `.optional()` modifier on the field.

https://zod.dev/?id=coercion-for-primitives
image

The fix I propose for this is the `CoercedBoolean` zod util I added, which accepts either a boolean or a string `"true"` or `"false"`. We could also expand this to accept `"1"` and `"0"` but considering I couldn't see a use of this in the docs, I omitted it here.

This could be considered breaking, if someone was relying on the previous behaviour of any non-empty string equalling `true`.

[Zod Playground](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHsBGArGCAFwApgAdAOwAJqFlrMBXSgSy0pIG0rbbMcWLABsYCTgEoANLz6YYlJnG7kQRAE5MYqqdVUAzBMLDaQAXQkYN4sPqzrlJAG4TqAXgB81J%2B7du9apqm0rIWMpSyOAyYEFgw6hAwGIIiYpJUAL4SVCBSIE5GWmAoXCAUIAiqDAZGJjoBOFUBhsamGbkg5nlO8WDslCggAMwYAEwALBjjHaxgAFpYACYAsqxsKC0mGUA)

---

## Changelog

fix: zod boolean coercion should correctly parse "true" and "false"

---

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of boolean values in environment variables and filters to ensure correct interpretation of "true" and "false" string inputs.
- **New Features**
	- Introduced enhanced boolean coercion for filters and environment settings, providing more reliable validation and parsing of boolean fields.

### DIFF
--- a/apps/supervisor/src/env.ts
+++ b/apps/supervisor/src/env.ts
@@ -1,7 +1,7 @@
 import { randomUUID } from "crypto";
 import { env as stdEnv } from "std-env";
 import { z } from "zod";
-import { AdditionalEnvVars, BoolEnv } from "./envUtil.js";
+import { AdditionalEnvVars, BoolEnv, CoercedBoolean } from "./envUtil.js";
 
 const Env = z.object({
   // This will come from `spec.nodeName` in k8s
@@ -45,7 +45,7 @@ const Env = z.object({
   // Used by the workload manager, e.g docker/k8s
   DOCKER_NETWORK: z.string().default("host"),
   OTEL_EXPORTER_OTLP_ENDPOINT: z.string().url(),
-  ENFORCE_MACHINE_PRESETS: z.coerce.boolean().default(false),
+  ENFORCE_MACHINE_PRESETS: CoercedBoolean.default(false),
   KUBERNETES_IMAGE_PULL_SECRETS: z.string().optional(), // csv
 
   // Used by the resource monitor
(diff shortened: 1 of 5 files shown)


nicktrn:
Hey @ghostdevv thanks for this! Not sure how we missed this incl. your PR :eyes:

I ran into the same `SMTP_SECURE` issue yesterday while working on v4 self-hosting and fixed this here: https://github.com/triggerdotdev/trigger.dev/pull/2090

There's an existing `BoolEnv` util with tests in the supervisor, we just weren't using it everywhere: https://github.com/triggerdotdev/trigger.dev/blob/c8d252e30091c446d275754df4c1f77e8e31f934/apps/supervisor/src/envUtil.test.ts#L4

Thanks for flagging the supervisor env var, will get that fixed too. I'll close this PR as it would clash.

PR #2083, opened 2025-05-21
Add experimental externals detection

nicktrn:
If your tasks fail to build you can try the new `build.autoDetectExternal` trigger config option. It should NOT be used otherwise as it may break deploys, we're still testing this.


ericallam (approved):
Rename config property to prefix with `experimental_`

PR #2109, opened 2025-05-27
v4: Can now set project ref using the TRIGGER_PROJECT_REF env var

ericallam:


packages/cli-v3/src/utilities/envVars.ts
```diff
@@ -0,0 +1,23 @@
...
+    ...sanitizeEnvVars(processEnv),
+    ...sanitizeEnvVars(additionalVariables ?? {}),
+    ...sanitizeEnvVars(dotEnvVars),
+  };
+}
+
+function gatherProcessEnv() {
+  const $env = {
+    ...process.env,
+    NODE_ENV: "development",
```
matt-aitken: @ericallam is this correct? When deployed too?

matt-aitken (approved):

PR #2127, opened 2025-05-29
v4: dequeue performance improvements (split concurrency from dequeue)

ericallam:
This change restructures how dequeuing works in v4, increasing efficiency by an order of magnitude or more, by splitting a worker dequeuing a run and the concurrency calculations.

Previously, every single "dequeue" call from a worker (in dev, the worker is the dev CLI command, in deployed tasks, the worker is the QueueRaider), would have to search through queues until it found one with concurrency available, and then it would dequeue a single run from that queue. This mean that there was a LOT of wasted searching for eligible queues.

Now we've introduced a new data structure called a worker queue, that is a Redis list. Workers when dequeuing just do a BLPOP on their list and don't have to worry about doing any concurrency calculations.

Separately, we have internal consumers that are constantly searching for queues that have capacity and when they do, bulk dequeue messages and add them to their worker queue lists. The concurrency consumers now don't have to stop as soon as they get a single message from a single queue, they can just continuously search for runs to add to worker queues.

In addition to the consumers, we now can eagerly move runs to worker queues when runs are triggered or completed (e.g. enqueued or acked), which means even LESS searching than before since it's more targeted.

Having the worker lists also gives us the ability to know almost exactly how many runs are ready to be executed which haven't executed yet (the size of the worker queue list), giving us handy metrics and something we can eventually use for auto-scaling.

### DIFF
--- a/internal-packages/run-engine/src/engine/tests/dequeuing.test.ts
+++ b/internal-packages/run-engine/src/engine/tests/dequeuing.test.ts
@@ -1,13 +1,12 @@
 import { containerTest } from "@internal/testcontainers";
 import { trace } from "@internal/tracing";
+import { DequeuedMessage } from "@trigger.dev/core/v3";
 import { generateFriendlyId } from "@trigger.dev/core/v3/isomorphic";
 import { PrismaClientOrTransaction } from "@trigger.dev/database";
 import { expect } from "vitest";
 import { MinimalAuthenticatedEnvironment } from "../../shared/index.js";
 import { RunEngine } from "../index.js";
 import { setupAuthenticatedEnvironment, setupBackgroundWorker } from "./setup.js";
-import { DequeuedMessage } from "@trigger.dev/core/v3";
-import { setTimeout } from "timers/promises";
 
 vi.setConfig({ testTimeout: 60_000 });
 
@@ -62,12 +61,9 @@ describe("RunEngine dequeuing", () => {
       });
       expect(runs.length).toBe(10);
 
-      //check the queue length
-      const queueLength = await engine.runQueue.lengthOfEnvQueue(authenticatedEnvironment);
-      expect(queueLength).toBe(10);
-
       //dequeue
-      await setTimeout(500);
+      await engine.runQueue.processMasterQueueForEnvironment(authenticatedEnvironment.id, 5);
+
       const dequeued: DequeuedMessage[] = [];
       for (let i = 0; i < 5; i++) {
         dequeued.push(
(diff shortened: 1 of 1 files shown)


apps/webapp/app/routes/engine.v1.worker-actions.deployments.$deploymentFriendlyId.dequeue.ts
```diff
@@ -14,55 +13,7 @@ export const loader = createLoaderWorkerApiRoute(
...
-          deployment.environmentId
-        )
-      : await authenticatedWorker.dequeueFromVersion(
-          deployment.worker.id,
-          searchParams.maxRunCount
-        );
-
-    return json(dequeuedMessages);
+  async (): Promise<TypedResponse<WorkerApiDequeueResponseBody>> => {
+    return json([]);
```
matt-aitken, line 17: @ericallam does this mean when we put the platform live, we'll need to very quickly deploy the new supervisor as well otherwise no v4 runs will get dequeued?

apps/webapp/app/v3/tracer.server.ts
```diff
@@ -261,10 +283,84 @@ function getTracer() {
...
+  busyGauge.addCallback(async (res) => {
+    const { busy } = await readPoolCounters();
+    res.observe(busy);
+  });
+
+  freeGauge.addCallback(async (res) => {
+    const { free } = await readPoolCounters();
+    res.observe(free);
+  });
+}
```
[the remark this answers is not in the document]
matt-aitken: @ericallam this is interesting

internal-packages/run-engine/src/engine/systems/dequeueSystem.ts
```diff
@@ -35,519 +35,475 @@ export class DequeueSystem {
...
+                  runId,
+                  error: {
+                    type: "INTERNAL_ERROR",
+                    code: "TASK_DEQUEUED_INVALID_STATE",
+                    message: `Task was in the ${snapshot.executionStatus} state when it was dequeued for execution.`,
+                  },
+                  tx: prisma,
+                });
+                this.$.logger.error(
+                  `RunEngine.dequeueFromMasterQueue(): Run is not in a valid state to be dequeued: ${runId}\n ${snapshot.id}:${snapshot.executionStatus}`
```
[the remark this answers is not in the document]
matt-aitken: @ericallam this is a good point

internal-packages/run-engine/src/run-queue/index.ts
```diff
@@ -825,19 +946,286 @@ export class RunQueue {
...
+          processedCount,
+          lastProcessedAt,
+          service: this.name,
+          shard,
+          consumerId,
+        });
+
+        const now = performance.now();
+
+        const results = await this.#processMasterQueueShard(shard, consumerId);
```
matt-aitken: @ericallam are you very sure this function can't throw? If it does then the consumer will stop until we we restart the service. I'd use `tryCatch` here so it protects against any future changes too.

internal-packages/run-engine/src/run-queue/index.ts
```diff
@@ -825,19 +946,286 @@ export class RunQueue {
...
+                messageQueue: queue,
+                shard,
+                // TODO: make this configurable
+                maxCount,
+              })
+            );
+
+            if (error) {
+              this.logger.error(
+                `[dequeueMessageInSharedQueue][${this.name}] Failed to dequeue from queue ${queue}`,
```
matt-aitken: @ericallam let's update the log to the new fn

PR #2183, opened 2025-06-18
v4: Experimental process keep alive

ericallam:
Currently during a warm start, we still recreate the actual task run process between runs, leading to a completely fresh global environment for each run. This experimental option will keep the task run process alive between run executions, leading to even faster warm starts. This option is respected in both the `dev` CLI and in deployed tasks.

To enable this option, add this to your `trigger.config.ts`:

```ts
import { defineConfig } from "@trigger.dev/sdk";

export default defineConfig({
  project: "<project ref>",
  // This is false by default
  experimental_processKeepAlive: true,
  maxDuration: 60,
});
```

You can also pass an object to `experimental_processKeepAlive` to provide more options:

```ts
import { defineConfig } from "@trigger.dev/sdk";

export default defineConfig({
  project: "<project ref>",
  experimental_processKeepAlive: {
    enabled: true,
    // After 20 runs execute with a single process, we'll restart the process and start fresh
    maxExecutionsPerProcess: 20,
    // In dev, you can combine this option with setting a max pool size, giving you the ability to limit the number of processes created on your local dev machine. Has no effect on deployed tasks
    devMaxPoolSize: 10,
  },
  maxDuration: 60,
});
```

## Gotchas

- Be careful with memory usage and memory leaks, as this will cause memory to persist across run executions.
- It's possible different tasks get executed in the same persisted process.
- If you configure any 3rd party SDKs globally using env vars for API keys, those SDKs will not change between runs. So if you change an env var between runs, the SDK will be "stale" and continue using the old env var value. Instead, you should initialize SDKs using env vars at runtime (in any of the lifecycle hooks or inside the `run` function of a task.
- Cancelling a task will cause the task run process to be restarted. Exiting the process is the only reliable way to actually stop a running function from stopping.
- This DOES NOT effect cold starts, warm starts only will be improved. 

We recommend enabling this option and testing in a staging or preview environment before trying it out in prod, as there could be unknown issues depending on what you are doing in your tasks.

## How to use

Install the following version of all the `@trigger.dev/*` packages and the `trigger.dev` CLI: `0.0.0-process-keep-alive-20250623153513`

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #2219 
- <kbd>&nbsp;2&nbsp;</kbd> #2189 
- <kbd>&nbsp;1&nbsp;</kbd> #2183 👈

### DIFF
--- a/packages/cli-v3/src/dev/devSupervisor.ts
+++ b/packages/cli-v3/src/dev/devSupervisor.ts
@@ -358,9 +358,6 @@ class DevSupervisor implements WorkerRuntime {
           httpClient: this.options.client,
           logLevel: this.options.args.logLevel,
           taskRunProcessPool: this.taskRunProcessPool,
-          cwd: this.options.config.experimental_devProcessCwdInBuildDir
-            ? worker.build.outputPath
-            : undefined,
           onFinished: () => {
             logger.debug("[DevSupervisor] Run finished", { runId: message.run.friendlyId });
 


packages/cli-v3/src/dev/devSupervisor.ts
```diff
@@ -95,6 +98,39 @@ class DevSupervisor implements WorkerRuntime {
...
+
+    this.taskRunProcessPool = new TaskRunProcessPool({
+      env,
+      cwd: this.options.config.workingDir,
+      enableProcessReuse:
+        typeof this.options.config.experimental_processKeepAlive === "boolean"
+          ? this.options.config.experimental_processKeepAlive
+          : typeof this.options.config.experimental_processKeepAlive === "object"
+          ? this.options.config.experimental_processKeepAlive.enabled
+          : false,
```
nicktrn: The `enableProcessReuse` var and prop differ

packages/cli-v3/src/entryPoints/managed/execution.ts
```diff
@@ -111,7 +115,9 @@ export class RunExecution {
    * Kills the current execution.
    */
   public async kill({ exitExecution = true }: { exitExecution?: boolean } = {}) {
-    await this.taskRunProcess?.kill("SIGKILL");
+    if (this.taskRunProcess) {
+      await this.taskRunProcessProvider.handleProcessAbort(this.taskRunProcess);
+    }
```
nicktrn, line 120: need to check that this does the same thing as before. `handleProcessAbort` calls process.cleanup instead of process.kill

packages/cli-v3/src/entryPoints/managed/execution.ts
```diff
@@ -249,7 +232,11 @@ export class RunExecution {
     if (this.currentAttemptNumber && this.currentAttemptNumber !== run.attemptNumber) {
       this.sendDebugLog("error: attempt number mismatch", snapshotMetadata);
       // This is a rogue execution, a new one will already have been created elsewhere
-      await this.exitTaskRunProcessWithoutFailingRun({ flush: false });
+      // TODO: keep this one, kill the process even if it's a keep-alive one
```
nicktrn: todo?

packages/cli-v3/src/entryPoints/managed/execution.ts
```diff
@@ -263,7 +250,11 @@ export class RunExecution {
     if (deprecated) {
       this.sendDebugLog("run execution is deprecated", { incomingSnapshot: snapshot });
 
-      await this.exitTaskRunProcessWithoutFailingRun({ flush: false });
+      // TODO: keep this one, kill the process even if it's a keep-alive one
```
nicktrn: todo?

packages/cli-v3/src/entryPoints/managed/execution.ts
```diff
@@ -286,13 +277,13 @@ export class RunExecution {
...
+        await this.exitTaskRunProcessWithoutFailingRun({ flush: true, reason: "re-queued" });
         return;
       }
       case "FINISHED": {
         this.sendDebugLog("run is finished", snapshotMetadata);
 
-        await this.exitTaskRunProcessWithoutFailingRun({ flush: true });
+        // This can sometimes be called before the handleCompletionResult, so we don't need to do anything here
         return;
       }
```
nicktrn, line 294: I wonder if we need to exit here after all. Maybe not immediately, but as a backup. It could help catch race conditions where the poller detects a run is finished, even if handleCompletionResult is never called.

packages/cli-v3/src/executions/taskRunProcess.ts
```diff
@@ -150,6 +148,7 @@ export class TaskRunProcess {
       PATH: process.env.PATH,
       TRIGGER_PROCESS_FORK_START_TIME: String(Date.now()),
       TRIGGER_WARM_START: this.options.isWarmStart ? "true" : "false",
+      TRIGGERDOTDEV: "1",
```
nicktrn, line 151: :partying_face:

packages/cli-v3/src/entryPoints/managed/execution.ts
ericallam: This is all handled in any case when returning the process to the task run provider

PR #2206, opened 2025-06-27
v4: current concurrency sweeper

ericallam:


internal-packages/run-engine/src/run-queue/index.ts
```diff
@@ -1651,6 +1712,243 @@ export class RunQueue {
...
+
+      if (!runIds || runIds.length === 0) {
+        stream.resume();
+        return;
+      }
+
+      const deduplicatedRunIds = Array.from(new Set(runIds));
+
+      const [processError] = await tryCatch(
+        this.processCurrencyConcurrencyRunIds(concurrencyKey, deduplicatedRunIds)
```
matt-aitken: Great name 🤑🤑🤑

Think you mean processCurrenT

internal-packages/run-engine/src/run-queue/index.ts
```diff
@@ -1651,6 +1712,243 @@ export class RunQueue {
...
+      markedForAckKey,
+      count,
+    });
+  }
+
+  // Call this every 5 minutes
+  private async processMarkedRuns() {
+    if (this.abortController.signal.aborted) {
+      return;
+    }
```
matt-aitken, line 1906: We're only 10 every 5 mins, we could easily get behind here.

It would also be good to know if we're at the max (i.e. we have 10) so we can monitor and adjust accordingly if we're constantly at the max.

packages/redis-worker/src/worker.ts
```diff
@@ -513,6 +539,10 @@ class Worker<TCatalog extends WorkerCatalog> {
 
         // On success, acknowledge the item.
         await this.queue.ack(id, deduplicationKey);
+
+        if (catalogItem.cron) {
```
matt-aitken, line 543: We should use a try/finally block in here so we are very confident the next one gets scheduled. Otherwise until we deploy again nothing will happen

PR #2217, opened 2025-07-01
Eliminate race conditions in continueRunIfUnblocked to prevent stuck runs

ericallam:


internal-packages/run-engine/src/engine/systems/waitpointSystem.ts
```diff
@@ -499,190 +499,249 @@ export class WaitpointSystem {
...
-          this.$.logger.error(`continueRunIfUnblocked: run has no checkpoint`, {
-            runId: run.id,
-            snapshot,
-            blockingWaitpoints,
-          });
-          throw new Error(`continueRunIfUnblocked: run has no checkpoint: ${run.id}`);
+        default: {
+          throw new Error(
+            `continueRunIfUnblocked: invalid execution status: ${snapshot.executionStatus}`
+          );
```
matt-aitken: @ericallam we don't want an assertNever here to catch at build time?

matt-aitken (approved):

PR #2224, opened 2025-07-02
Add ECR support for remote builds

nicktrn:
Summary:
- Create ECR repos on deployment if they don't already exist
- Handles ephemeral ECR auth token creation for pushes
- Optional STS assume role support
- There are tests but mostly disabled by default due to additional config required
- This is all optional and gated by `DEPLOY_REGISTRY_HOST` detection

New (optional) env vars:
- `DEPLOY_REGISTRY_ECR_TAGS`
- `DEPLOY_REGISTRY_ECR_ASSUME_ROLE_ARN`
- `DEPLOY_REGISTRY_ECR_ASSUME_ROLE_EXTERNAL_ID`


apps/webapp/app/v3/getDeploymentImageRef.server.ts
```diff
@@ -0,0 +1,379 @@
...
+  const sts = new STSClient({ region });
+
+  // Generate a unique session name using timestamp and random string
+  // This helps with debugging but doesn't affect concurrent sessions
+  const timestamp = Date.now();
+  const randomSuffix = Math.random().toString(36).substring(2, 8);
+  const sessionName = `TriggerWebappECRAccess_${timestamp}_${randomSuffix}`;
+
+  try {
+    const response = await sts.send(
```
ericallam: Shouldn't this use tryCatch?

PR #2225, opened 2025-07-03
Expose all run options in the test run page

myftija:
Changes in this PR:

- Added a new pane in the test run page to enable setting all run options that are also exposed in the SDK.
- Added a custom duration picker UI component with nice keyboard support.
- Added a component to input run tags.
- Applied some slight improvements to the test page form submission approach, by removing the custom submit function and relying on the `onSubmit` hook exposed by conform's `useForm` instead.

Still WIP, need to move the recent runs list to another place in the test page, to avoid visual overload.


myftija:
Continuing this in another PR: https://github.com/triggerdotdev/trigger.dev/pull/2227

PR #2239, opened 2025-07-05
fix(packages/cli-v3/src/utilities/cliOutput.ts): improve contrast for chalkWorker in light mode

### LINKED ISSUE #2214: Poor contrast on the CLI in light mode
Running `npx trigger.dev@latest dev` in the terminal in light mode results in the image below where the yellow text is unreadable.

Image

The text is only readable after changing the terminal to dark mode.

Image

Please change the text colors used in light mode to improve contrast and readability, currently it seems like the same colors are used in both light and dark mode but suggest to use a different color palette for each mode.

bk7312:
Fixes #2214 

The CLI uses chalk for text color. Currently, the contrast is rather poor in light mode, especially `chalkWorker` which is unreadable. 

Unfortunately, there's no way to use different colors for light/dark mode and there's no reliable way to detect the terminal background color. 

This PR is just a quick fix and changes the color for `chalkWorker` from `#FFFF89` to `yellowBright` to make the text at least readable in light mode.

Below is a screenshot for what each color looks like in light mode for reference:

light

Compared to in dark mode:

dark

No, I did not run the code locally to test. Tried to but ran into some errors when trying to migrate the db when running `pnpm run db:migrate`, "Error: P1010: User `postgres` was denied access on the database `postgres.public`" and gave up 😅 .


nicktrn (approved):
Thanks for this @bk7312, as dark mode addicts we definitely missed this!

It also seems like light mode IDE terminals handle these bright colours more gracefully 🤔

PR #2260, opened 2025-07-11
feat: enable overriding run options when replaying

myftija:
Changes in this PR:

- Added run options and metadata to the replay run modal.
- Rearranged the layout a bit.
- Fixed an issue with the clear button in the json editor component.

image

### DIFF
--- a/apps/webapp/app/routes/resources.taskruns.$runParam.replay.ts
+++ b/apps/webapp/app/routes/resources.taskruns.$runParam.replay.ts
@@ -20,9 +20,16 @@ const ParamSchema = z.object({
   runParam: z.string(),
 });
 
+const QuerySchema = z.object({
+  environmentIdOverride: z.string().optional(),
+});
+
 export async function loader({ request, params }: LoaderFunctionArgs) {
   const userId = await requireUserId(request);
   const { runParam } = ParamSchema.parse(params);
+  const { environmentIdOverride } = QuerySchema.parse(
+    Object.fromEntries(new URL(request.url).searchParams)
+  );
 
   const run = await $replica.taskRun.findFirst({
     select: {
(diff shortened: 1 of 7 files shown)


ericallam:
Can we move the version and the machine options higher so they aren't "below the fold". Maybe just swap them with the delay and TTL options?

myftija:
@ericallam yes, sounds good, will update the test page too for consistency.

apps/webapp/app/routes/resources.taskruns.$runParam.replay.ts
```diff
@@ -71,9 +86,82 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+            queueId: true,
+          },
+          where: {
+            slug: run.taskIdentifier,
+            runtimeEnvironmentId: environment.id,
+          },
+          orderBy: {
+            createdAt: "desc",
+          },
+        });
```
nicktrn: would create a helper function for this to clean things up, mainly to remove this optional chaining:
```ts
      ? (await findCurrentWorkerDeployment({ environmentId: environment.id }))?.worker?.tasks.find(
          (t) => t.slug === run.taskIdentifier
        )
```

apps/webapp/app/v3/services/replayTaskRun.server.ts
```diff
@@ -36,57 +35,41 @@ export class ReplayTaskRunService extends BaseService {
...
-        payloadPacket = await stringifyIO(newPayload);
-      } else {
-        payloadPacket = await conditionallyImportPacket({
-          data: overrideOptions.payload,
-          dataType: existingTaskRun.payloadType,
-        });
-      }
-    } else {
-      payloadPacket = await conditionallyImportPacket({
+    const getExistingPayload = async () => {
```
nicktrn: this is probably fine, but prefer private methods instead of nested function declarations

apps/webapp/app/v3/services/replayTaskRun.server.ts
```diff
@@ -36,57 +35,41 @@ export class ReplayTaskRunService extends BaseService {
...
-        environmentId: authenticatedEnvironment.id,
-      });
+      const tags =
+        overrideOptions.tags ??
+        (
+          await getTagsForRunId({
+            friendlyId: existingTaskRun.friendlyId,
+            environmentId: authenticatedEnvironment.id,
+          })
+        )?.map((t) => t.name);
```
nicktrn: try to avoid optional chaining after await

nicktrn:
Great stuff! Just some minor code style suggestions

apps/webapp/app/routes/resources.taskruns.$runParam.replay.ts
```diff
@@ -71,9 +86,82 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+        select: {
+          friendlyId: true,
+          name: true,
+          type: true,
+          paused: true,
+        },
+      })
+    : undefined;
+
+  const backgroundWorkers = await $replica.backgroundWorker.findMany({
```
myftija: Realized this needs to be fetched again if the environment is overriden. Will adapt.

samejr:
This looks great. 👍 Could you extend the height of the JSON viewer to fill the full height of the Dialog?

CleanShot 2025-07-11 at 10 28 32

apps/webapp/app/v3/services/replayTaskRun.server.ts
myftija: yes, good point

apps/webapp/app/routes/resources.taskruns.$runParam.replay.ts
myftija: good point, it was quite unreadable too, updated 👍

apps/webapp/app/routes/resources.taskruns.$runParam.replay.ts
myftija: done

PR #2262, opened 2025-07-11
feat: link the original run from replayed runs

myftija:
This PR introduces a small improvement in the replay flow, by linking the original run.

### DIFF
--- a/apps/webapp/app/presenters/v3/SpanPresenter.server.ts
+++ b/apps/webapp/app/presenters/v3/SpanPresenter.server.ts
@@ -218,6 +218,7 @@ export class SpanPresenter extends BasePresenter {
             friendlyId: true,
           },
         },
+        replayedFromTaskRunFriendlyId: true,
       },
       where: span.originalRun
         ? {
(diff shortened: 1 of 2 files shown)


internal-packages/database/prisma/schema.prisma
```diff
@@ -644,6 +644,8 @@ model TaskRun {
 
   logsDeletedAt DateTime?
 
+  replayedFromTaskRunFriendlyId String?
```
myftija, line 647: Not happy with this name. Also it's not consistent with the other references to runs, like `rootTaskRunId`, `parentTaskRunId`, etc., where we store the internal ID as reference. But in this case it is more efficient to store the friendly ID as it avoids having to do a self-join and suffices to generate the link to the `replayedFrom` run. Other ideas?

nicktrn: In v4 the friendly IDs are `run_<internalId>`. There are helpers to "convert" from one to the other. Will hopefully simplify this 👍

myftija: Oh nice, then we can go with this for now and switch to internal IDs once we deprecate v3.

PR #2263, opened 2025-07-11
fix: Logging large objects is now much more performant and uses less memory

ericallam:


apps/webapp/app/env.server.ts
```diff
@@ -283,8 +283,8 @@ const EnvironmentSchema = z.object({
   PROD_OTEL_LOG_EXPORT_TIMEOUT_MILLIS: z.string().default("30000"),
   PROD_OTEL_LOG_MAX_QUEUE_SIZE: z.string().default("512"),
 
-  TRIGGER_OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT: z.string().default("256"),
-  TRIGGER_OTEL_LOG_ATTRIBUTE_COUNT_LIMIT: z.string().default("256"),
+  TRIGGER_OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT: z.string().default("1024"),
+  TRIGGER_OTEL_LOG_ATTRIBUTE_COUNT_LIMIT: z.string().default("1024"),
```
nicktrn, line 287: we should try to keep the self-hosting env var page updated when changing defaults, I'll push to this pr

nicktrn (approved):

PR #2311, opened 2025-07-24
sentry: remove node native integration because it breaks

ericallam:


timfish:
Sorry to hear this wasn't working for you. This integration has only just been released. What was breaking?

ericallam:
@timfish it was throwing this error:

```
Error: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /triggerdotdev/node_modules/.pnpm/@sentry-internal+node-native-stacktrace@0.2.1/node_modules/@sentry-internal/node-native-stacktrace/lib/stack-trace-linux-x64-glibc-115.node)
```

timfish:
What linux distribution are you running on? 

I think since Github dropped ubuntu-20.04 support, we compile on ubuntu-22.04 and this sets the minimum glibc version.

My guess would be that you're using an older version of Debian?

PR #2317, opened 2025-07-25
Restructured the realtime docs 

D-K-P:


samejr:
I think this issue is fixed now and can be removed: https://trigger-docs-realtime-docs-updates.mintlify.app/realtime/how-it-works#known-issues

PR #2322, opened 2025-07-29
feat: run.ctx tidying and additions

ericallam:


apps/webapp/app/routes/resources.runs.$runParam.ts
```diff
@@ -76,6 +76,30 @@ export const loader = async ({ request, params }: LoaderFunctionArgs) => {
       lockedBy: {
         select: {
           filePath: true,
+          worker: {
```
ericallam, line 79: TODO look into this because not sure if it needs updating or removing

packages/trigger-sdk/src/v3/usage.ts
```diff
@@ -65,12 +65,12 @@ export const usage = {
             durationMs: 0,
           },
           total: {
-            costInCents: run?.costInCents ?? 0,
-            durationMs: run?.durationMs ?? 0,
+            costInCents: 0,
```
ericallam: Add the cost stuff back into the execution, but still keep it out of the ctx, so this still works across attempts 2+

matt-aitken (approved):

PR #2337, opened 2025-08-02
fix(otel): prevent infinite retry loops on unicode hex escape errors

nicktrn:
## Summary

Fixed infinite retry loops in OTEL log ingestion caused by Unicode validation errors in PostgreSQL. The issue occurred when malformed Unicode data in telemetry logs triggered "lone leading surrogate in
hex escape" and "unexpected end of hex escape" errors, which were not properly handled by the DynamicFlushScheduler's retry logic.

### Unicode Error Handling
- Added specific handling for `PrismaClientKnownRequestError` with Unicode/hex escape error messages
- Implemented bisection logic for Unicode errors to isolate problematic events while preserving clean data
- Created `isUnicodeError()` helper function for reusable Unicode error detection
- Added `getPrismaErrorDetails()` helper for type-safe error detail extraction

### Observability Improvements
- Added span attributes for better error tracking: `prisma_error_type`, `prisma_error_code`
- Enhanced error logging with structured details for both known and unknown Prisma errors

### Prisma Logging Improvements
- Converted Prisma logs from unstructured stdout to structured JSON events
- Added optional `PRISMA_LOG_TO_STDOUT` environment variable to control error output format. If it equals "1" we log straight to stdout instead of using structured logs, this is helpful when running locally.
- Added `clientType` field to distinguish between writer and reader client logs

## Root Cause
The recent DynamicFlushScheduler enhancements added retry logic that re-queued failed batches, but didn't account for Unicode serialization errors that are permanent data issues rather than transient
failures. This caused problematic batches to retry indefinitely.

## Resolution
Unicode errors now trigger the same bisection strategy used for other data issues, recursively splitting batches until individual problematic events are isolated and dropped, allowing clean events to be
successfully inserted.

### DIFF
--- a/.github/workflows/unit-tests-webapp.yml
+++ b/.github/workflows/unit-tests-webapp.yml
@@ -12,8 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
-        shardTotal: [10]
+        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8]
+        shardTotal: [8]
     env:
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
       SHARD_INDEX: ${{ matrix.shardIndex }}


apps/webapp/app/db.server.ts
```diff
@@ -136,16 +145,60 @@ function getClient() {
...
   });
 
+  client.$on("info", (log) => {
+    logger.info("PrismaClient info", {
+      clientType: "writer",
+      timestamp: log.timestamp,
+      message: log.message,
+      target: log.target,
+    });
+  });
```
matt-aitken: Isn't this going to be extremely noisy?

apps/webapp/app/db.server.ts
```diff
@@ -188,16 +250,60 @@ function getReplicaClient() {
...
+    ],
   });
 
+  replicaClient.$on("info", (log) => {
+    logger.info("PrismaClient info", {
+      clientType: "reader",
+      timestamp: log.timestamp,
+      message: log.message,
+      target: log.target,
+    });
```
matt-aitken: Same for this? Too noisy?

matt-aitken (approved):

apps/webapp/app/db.server.ts
nicktrn: there's actually not many info logs, they were previously printing to stdout

PR #2379, opened 2025-08-11
Remove triggerAndPoll()

matt-aitken:
This function always had lots of warnings around it – it’s a bad idea to do this because if you have a lot of runs in the queue you’re going to waiting a long time on your backend.

So we’ve made the decision to remove it. You are welcome to poll yourself and use `runs.retrieve()` but we strongly recommend against this non-async pattern.


whchi:
Why this not mentioned in breaking changes?

PR #2393, opened 2025-08-14
fix: code editor/block theme inconsistency

myftija:
Tiny fix to make the themes across code blocks and editor components across the app consistent.


matt-aitken (approved):
🧑‍🍳💋

PR #2422, opened 2025-08-19
chore(webapp): More false errors turned to logs

matt-aitken:


apps/webapp/app/db.server.ts
```diff
@@ -201,6 +201,7 @@ function getClient() {
           message: log.message,
           target: log.target,
         },
+        ignoreError: true,
```
nicktrn, line 204: Too many of these?

PR #2436, opened 2025-08-21
fix(v3): adopt v4 dequeue strategy to increase perf

ericallam:
We will now dequeue into a redis list as many messages as we possibly 
can from actual queues, trying to dequeue up to 10 runs at a time for 
each queue in an env, in separate queue consumers. Then the 
SharedQueueConsumer can simply `lpop` off that list and process a run to
be dequeued very quickly. This should get runs through the platform side
dequeue system very quickly.

### DIFF
--- a/apps/webapp/app/v3/marqs/index.server.ts
+++ b/apps/webapp/app/v3/marqs/index.server.ts
@@ -42,6 +42,8 @@ import {
   MARQS_RETRY_PRIORITY_TIMESTAMP_OFFSET,
   MARQS_SCHEDULED_REQUEUE_AVAILABLE_AT_THRESHOLD_IN_MS,
 } from "./constants.server";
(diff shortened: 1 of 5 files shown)


apps/webapp/app/v3/marqs/index.server.ts
```diff
@@ -1862,27 +2016,38 @@ if queueCurrentConcurrency >= totalQueueConcurrencyLimit then
...
+    return nil
+end
+
+local messageIds = {}
+for i = 1, #messagesWithScores, 2 do
+    table.insert(messageIds, messagesWithScores[i])
+end
+
+-- Remove the messages from the queue and update concurrency
+redis.call('ZREM', queueKey, unpack(messageIds))
```
matt-aitken, line 2041: unpack is cool

PR #2467, opened 2025-09-02
fix(run-engine): retry non-zero exit code errors

ericallam:
We’re also now saving the retryConfig from the BackgroundWorkerTask on 
TaskRun.lockedRetryConfig when the run is first locked to the version


internal-packages/database/prisma/migrations/migration_lock.toml
```diff
@@ -1,3 +1,3 @@
 # Please do not edit this file manually
-# It should be added in your version-control system (i.e. Git)
-provider = "postgresql"
\ No newline at end of file
+# It should be added in your version-control system (e.g., Git)
```
nicktrn: What's with this?

PR #2485, opened 2025-09-09
feat: env command in the cli 

myftija:
This PR adds a couple of env commands to list and view environment variables
of a project. The `env pull` command also enables exporting the env vars into a file.

We might extend this to also support creating and deleting env vars in the future.

### DIFF
--- a/.changeset/serious-doors-call.md
+++ b/.changeset/serious-doors-call.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": minor
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/serious-doors-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"trigger.dev": minor
```
ericallam: lets do this as a patch for now

ericallam (approved):

PR #2497, opened 2025-09-11
docs(runs): add description and example to retrieve endpoint (closes …

dhinakaranst:
Checklist
I have followed every step in the contributing guide
 The PR title follows the convention
 I ran and tested that the docs page renders locally
Changelog
docs: add description and example to Runs “Retrieve run” page
image

Before: page showed no content beyond title
After: page displays description, TS example, and OpenAPI details


nicktrn:
Thanks, we fixed this since

PR #2501, opened 2025-09-15
feat(cli): attach to existing deployments in the build server

myftija:
In the build server we initialize the deployment before installing the project dependencies,
so that the status is correctly reflected in the dashboard. In this case, we need to attach
to the existing deployment and continue with the remote build process.

This is a workaround to avoid major changes in the deploy command and workflow. In the future,
we'll likely make the build server the entry point of the flow for building and deploying and also
adapt the related deployment API endpoints.

### DIFF
--- a/.changeset/seven-taxis-glow.md
+++ b/.changeset/seven-taxis-glow.md
@@ -1,0 +1,6 @@
+---
+"trigger.dev": patch
+"@trigger.dev/core": patch
+---
+
+Attach to existing deployment for deployments triggered in the build server. If `TRIGGER_EXISTING_DEPLOYMENT_ID` env var is set, the `deploy` command now skips the deployment initialization.


packages/core/src/v3/schemas/api.ts
```diff
@@ -467,6 +467,8 @@ export const GetDeploymentResponseBody = z.object({
   shortCode: z.string(),
   version: z.string(),
   imageReference: z.string().nullish(),
+  imagePlatform: z.string(),
```
myftija, line 470: this schema is not using `strict()`, so adding an additional required field won't cause compatibility issues with previous versions of the cli.

apps/webapp/app/routes/api.v1.deployments.$deploymentId.ts
```diff
@@ -52,7 +53,10 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
     shortCode: deployment.shortCode,
     version: deployment.version,
     imageReference: deployment.imageReference,
-    errorData: deployment.errorData,
+    imagePlatform: deployment.imagePlatform,
+    externalBuildData:
+      deployment.externalBuildData as GetDeploymentResponseBody["externalBuildData"],
```
myftija, line 58: Not ideal adding this to the get deployment endpoint, but it won't hurt either. The depot tokens are ephemeral and scoped, the initialize deployment endpoint already returns these.

The alternative would have been to do the matching to the existing deployment in the initialize endpoint instead, by adding a parameter for the existing deployment id. That would make its interface a bit awkward though.

We'll eventually adapt the deployment flow to use the build server as the entry point. We will then also be able to get rid of this workaround.

PR #2515, opened 2025-09-16
feat(sdk): replace onStart lifecycle hook with onStartAttempt

ericallam:
Prevent uncaught errors in the `onSuccess`, `onComplete`, and `onFailure` lifecycle hooks from failing attempts/runs.

Deprecated the `onStart` lifecycle hook (which only fires before the `run` function on the first attempt). Replaced with `onStartAttempt` that fires before the run function on every attempt:

```ts
export const taskWithOnStartAttempt = task({
  id: "task-with-on-start-attempt",
  onStartAttempt: async ({ payload, ctx }) => {
    //...
  },
  run: async (payload: any, { ctx }) => {
    //...
  },
});

// Default a global lifecycle hook using tasks
tasks.onStartAttempt(({ ctx, payload, task }) => {
  console.log(
    `Run ${ctx.run.id} started on task ${task} attempt ${ctx.run.attempt.number}`,
    ctx.run
  );
});
```

If you want to execute code before just the first attempt, you can use the `onStartAttempt` function and check `ctx.run.attempt.number === 1`:

```ts /trigger/on-start-attempt.ts
export const taskWithOnStartAttempt = task({
  id: "task-with-on-start-attempt",
  onStartAttempt: async ({ payload, ctx }) => {
    if (ctx.run.attempt.number === 1) {
      console.log("Run started on attempt 1", ctx.run);
    }
  },
});
```


packages/core/src/v3/workers/taskExecutor.ts
```diff
@@ -795,7 +797,7 @@ export class TaskExecutor {
 
     return await runTimelineMetrics.measureMetric("trigger.dev/execution", "success", async () => {
       for (const hook of globalSuccessHooks) {
-        const [hookError] = await tryCatch(
+        await tryCatch(
```
nicktrn, line 782: what happened to these? should we log them at least?

packages/core/src/v3/workers/taskExecutor.ts
```diff
@@ -817,14 +819,10 @@ export class TaskExecutor {
...
         );
-
-        if (hookError) {
-          throw hookError;
-        }
       }
 
       if (taskSuccessHook) {
-        const [hookError] = await tryCatch(
+        await tryCatch(
```
nicktrn, line 807: same

packages/core/src/v3/workers/taskExecutor.ts
```diff
@@ -892,14 +886,10 @@ export class TaskExecutor {
...
         );
-
-        if (hookError) {
-          throw hookError;
-        }
       }
 
       if (taskFailureHook) {
-        const [hookError] = await tryCatch(
+        await tryCatch(
```
nicktrn, line 874: same

packages/core/src/v3/workers/taskExecutor.ts
ericallam: They will be logged as errors in the span, we just don't rethrow them and fail the run anymore from these specific hooks

PR #2519, opened 2025-09-17
fix(engine): prevent race condition that prevents triggerAndWait runs from resuming by atomically creating associated waitpoint records

ericallam:
The tests in this PR don't actually test and verify the race condition, but they do test and verify that trying to create two runs with the same idempotency key, with one winning and one throwing a prisma p2002 error, results in the correct behavior. I originally thought this was what was causing the issue, and we didn't have tests for that behavior, so I added them. But the tests passed, leading me to find what the real cause of the issue was: a race between finding the task run with an idempotency key, and the associated waitpoint of that run being created. The fix was the create the associated waitpoint in a prisma [nested write](https://www.prisma.io/docs/orm/prisma-client/queries/relation-queries#nested-writes), which is in a transaction.

### DIFF
--- a/apps/webapp/app/runEngine/services/triggerTask.server.ts
+++ b/apps/webapp/app/runEngine/services/triggerTask.server.ts
@@ -37,10 +37,18 @@ import type {
   QueueManager,
   RunNumberIncrementer,
   TraceEventConcern,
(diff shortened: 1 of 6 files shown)


nicktrn (approved):
very nice

PR #2520, opened 2025-09-17
fix(docker): add missing email and auth environment variables to webapp | selfhoster

unckleg:
## ✅ Checklist
---
## Testing

1. Set up fresh v4-beta self-hosted deployment using docker-compose
2. Configured `.env` with `EMAIL_TRANSPORT=resend` and valid Resend API key
3. Before fix: Magic links appeared in docker logs, no emails received
4. After fix: Magic links sent successfully via email
5. Tested with both Resend and SMTP configurations
6. Verified WHITELISTED_EMAILS properly restricts login access
7. Confirmed GitHub OAuth works when AUTH_GITHUB variables are set

---
## Changelog

Added missing environment variables to webapp service in docker-compose.yml:
- Email transport configuration (EMAIL_TRANSPORT, FROM_EMAIL, REPLY_TO_EMAIL, RESEND_API_KEY)
- SMTP configuration variables (commented by default)
- Authentication variables (WHITELISTED_EMAILS, ADMIN_EMAILS, AUTH_GITHUB_CLIENT_ID/SECRET)
- Slack integration variables (ORG_SLACK_INTEGRATION_CLIENT_ID/SECRET)
- Alert email configuration (ALERT_FROM_EMAIL)

---
## Screenshots


nicktrn:
Thanks @unckleg - did you test what happens when people don't set these in their `.env`?

unckleg:
> Thanks @unckleg - did you test what happens when people don't set these in their `.env`?

Will continue to work normally since the conditions in the implementation itself.

nicktrn:
> Will continue to work normally

The reason I ask is because afaik both v1 and v2 docker compose will set those env vars to empty strings if not defined. This combined with schema parsing in env.server.ts might cause issues. The ones you added will be optional but some might require strict enums that exclude empty strings. It's also one of the reasons we didn't add all env vars to the compose file - we'd have to maintain defaults in both schema and compose file. In this case it might be fine, just needs testing.

PR #2530, opened 2025-09-19
feat(webapp): completing spans server-side no longer write-after-read, improving efficiency and perf

ericallam:
This change will prevent inefficient queries finding “in progress” span events to close server-side, instead we’ve now taken an approach that doesn’t not requiring task event reads before doing writes to close spans. This effects both v3 and v4, and is a platform-only change. This should lead to much more efficient cancellation, WAY fewer queries reading the TaskEvent/Partitioned tables, and lays the ground work for moving to ClickHouse for storing this data.


apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.$runParam.spans.$spanParam/route.tsx
```diff
@@ -772,6 +772,14 @@ function RunBody({
...
                       <Property.Value>{run.workerQueue}</Property.Value>
                     </Property.Item>
+                    <Property.Item>
+                      <Property.Label>Trace ID</Property.Label>
+                      <Property.Value>{run.traceId}</Property.Value>
+                    </Property.Item>
+                    <Property.Item>
+                      <Property.Label>Span ID</Property.Label>
+                      <Property.Value>{run.spanId}</Property.Value>
+                    </Property.Item>
```
nicktrn, line 782: :heart:

nicktrn (approved):

PR #2555, opened 2025-09-25
fix: improve Prisma version resolution for pnpm workspaces

hongkongkiwi:
## Summary
- Fixed package.json resolution bug that could select wrong package in pnpm workspaces
- Added fallback to read version directly from node_modules when primary detection fails
- Added support for workspace: protocol and helpful error messages for catalog: references

## Problem
When using pnpm workspaces without an explicit version in `trigger.config.ts`, the Prisma extension could fail to detect or select the wrong version due to:
1. Package.json resolution finding parent workspace package.json instead of @prisma/client's
2. No support for pnpm catalog: or workspace: version specifiers

## Solution
1. **Fixed root cause**: Added verification that resolved package.json belongs to the correct package
2. **Added robust fallback**: Read directly from `node_modules/@prisma/client/package.json` when primary detection fails
3. **Better error messages**: Detect catalog: references and provide clear guidance
4. **Zero new dependencies**: Solution uses only existing dependencies

## Test Plan
- [ ] Verify Prisma extension works with regular npm/yarn projects
- [ ] Test with pnpm workspace using workspace: protocol
- [ ] Test with pnpm catalog references (should get helpful error)
- [ ] Verify fallback to node_modules reading works when externals detection fails

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/build/src/extensions/prisma.ts
+++ b/packages/build/src/extensions/prisma.ts
@@ -2,8 +2,9 @@ import { BuildManifest, BuildTarget } from "@trigger.dev/core/v3";
 import { binaryForRuntime, BuildContext, BuildExtension } from "@trigger.dev/core/v3/build";
 import assert from "node:assert";
 import { existsSync } from "node:fs";
-import { cp, readdir } from "node:fs/promises";
+import { cp, readdir, readFile } from "node:fs/promises";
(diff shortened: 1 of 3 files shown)


ericallam:
This has been fixed in 4.1.1

PR #2560, opened 2025-09-25
feat(helm): support topology spread constraints

### LINKED ISSUE #2548: feat: Allow configuring the topology spread constraints in the Helm chart

It is currently not possible to specify topology spread constraints for the webapp and supervisor pods, as well as for the jobs started by the worker.

`topologySpreadConstraints` should be configurable similar to `nodeSelector`.

N/A

### Additional information

https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/

nicktrn:
fixes #2548 

all new settings incl env vars are optional. new helm chart released, but will need another bump once the new supervisor images are out to fully support spread constraints for the worker pods.

### DIFF
--- a/hosting/k8s/helm/templates/supervisor.yaml
+++ b/hosting/k8s/helm/templates/supervisor.yaml
@@ -184,10 +184,6 @@ spec:
               value: {{ default "10Gi" .Values.supervisor.config.kubernetes.ephemeralStorageSizeLimit | quote }}
             - name: KUBERNETES_EPHEMERAL_STORAGE_SIZE_REQUEST
               value: {{ default "2Gi" .Values.supervisor.config.kubernetes.ephemeralStorageSizeRequest | quote }}
-            {{- with .Values.supervisor.config.kubernetes.topologySpreadConstraints }}
-            - name: KUBERNETES_TOPOLOGY_SPREAD_CONSTRAINTS
-              value: {{ tpl (toYaml .) $ | toJson | quote }}
-            {{- end }}
             # Pod cleaner configuration
             - name: POD_CLEANER_ENABLED
               value: {{ .Values.supervisor.config.podCleaner.enabled | quote }}
@@ -275,10 +271,6 @@ spec:
       {{- with .Values.supervisor.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
-      {{- end }}
-      {{- with .Values.supervisor.topologySpreadConstraints }}
-      topologySpreadConstraints:
-        {{- tpl (toYaml .) $ | nindent 8 }}
       {{- end }}
 ---
 apiVersion: v1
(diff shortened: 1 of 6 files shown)


hosting/k8s/helm/templates/supervisor.yaml
```diff
@@ -184,6 +184,10 @@ spec:
               value: {{ default "10Gi" .Values.supervisor.config.kubernetes.ephemeralStorageSizeLimit | quote }}
             - name: KUBERNETES_EPHEMERAL_STORAGE_SIZE_REQUEST
               value: {{ default "2Gi" .Values.supervisor.config.kubernetes.ephemeralStorageSizeRequest | quote }}
+            {{- with .Values.supervisor.config.kubernetes.topologySpreadConstraints }}
+            - name: KUBERNETES_TOPOLOGY_SPREAD_CONSTRAINTS
+              value: {{ tpl (toYaml .) $ | toJson | quote }}
+            {{- end }}
```
praneetloke: I realize I requested adding support for topology spread constraints for the jobs too but I am not sure if, in practice, it would be effective, since only a single replica of a job is running at any given time, correct? If so, I suggest removing this for now.

hosting/k8s/helm/templates/supervisor.yaml
```diff
@@ -272,6 +276,10 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.supervisor.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- tpl (toYaml .) $ | nindent 8 }}
+      {{- end }}
```
praneetloke: It seems that supervisor deployment is locked to a replica count of 1. Curious what the reason for that is?

praneetloke: The reason I am asking is, although it's good to have the topology spread constraints config surfaced to users, it's likely going to be ineffectual because of the replica count being locked to 1.

hosting/k8s/helm/templates/supervisor.yaml
nicktrn: Fair point, the idea was to future-proof so spread constraints could be used across the board, but they wouldn't serve a purpose here yet. I might just remove this again for now

hosting/k8s/helm/templates/supervisor.yaml
nicktrn: We're restricted especially in the self-hosted setup where the supervisor is not spread out into multiple separate services with different configs. For example, if we spun up multiple instances with runner cleanup enabled they'd compete with each other due to using the same watches. Might remove this too until we support multiple replicas.

nicktrn:
Thanks for the comments @praneetloke - looks like this only really makes sense for the webapp right now. I think I'll remove this everywhere else to make it less confusing.

PR #2563, opened 2025-09-26
fix(run-engine): carryover batchId after PENDING_EXECUTING stalls

nicktrn:
This fixes batch resumes after stalling. Existing runs that are stuck here and are already EXECUTING again will have to be replayed.


matt-aitken (approved):

dlindb:
Interesting bug.

One thing I'm curious about: after the stalled run resumed, was there any notion of the run still being "current" relative to the state it was originally derived from, or was replay the only safe recovery path?

I'm seeing a recurring pattern in workflow systems where:

* state is observed
* state advances
* execution later resumes from the older observation

and I'm trying to understand how different systems handle that boundary in practice.

PR #2594, opened 2025-10-08
fix(otel): prevent unpaired unicode surrogate pairs from causing insert errors

ericallam:


apps/webapp/app/v3/otlpExporter.server.ts
```diff
@@ -774,24 +772,83 @@ function binaryToHex(buffer: Buffer | string | undefined): string | undefined {
...
+  return truncatedString;
+}
+
+const ASCII_ONLY_REGEX = /^[\x00-\x7F]*$/;
+
+function smartTruncateString(str: string, maximumLength: number): string {
+  if (!str) return "";
+  if (str.length <= maximumLength) return str;
+
+  if (ASCII_ONLY_REGEX.test(str)) {
```
nicktrn: I'm a bit worried about perf for long strings, maybe we could do this instead?

```suggestion
  // Check up to 2x the limit (worst case: all surrogates)
  const checkLength = Math.min(str.length, maximumLength * 2 + 2);
  if (ASCII_ONLY_REGEX.test(str.slice(0, checkLength))) {
```

apps/webapp/app/v3/otlpExporter.server.ts
```diff
@@ -774,24 +772,83 @@ function binaryToHex(buffer: Buffer | string | undefined): string | undefined {
...
+    return str.slice(0, maximumLength);
+  }
+
+  return [...str].slice(0, maximumLength).join("");
```
nicktrn: Similar thing here, maybe best to guard:

```suggestion
  return [...str.slice(0, checkLength)].slice(0, maximumLength).join("");
```

nicktrn (changes requested):
Left some code snippets, but you'd probably want to only slice by `checkLength` once

PR #2264, opened 2025-07-11
Bulk actions 2.0 (and switch all run listing to ClickHouse)

matt-aitken:
You can now cancel/replay runs based on filters that are applied to the runs list in the dashboard. You can also manually select runs for smaller operations.

We have a new "Bulk actions" page where you can view all previous actions, view the progress, abort, view associated runs, and quickly do a "Replay" of associated runs.

https://github.com/user-attachments/assets/0b9518ef-495f-4a29-84ed-9429d39585e5

This branch also removes the old Postgres `RunListPresenter`, in favour of the new ClickHouse `NextRunListPresenter`.

This switchover impacts:
- Runs page
- Waitpoint inspector
- Schedule inspector
- The `runs.list()` SDK call.

PR #2607, opened 2025-10-15
refactor: consolidate ClickHouse configuration into single CLICKHOUSE…

LeoKaynan:
## Summary

This PR consolidates redundant ClickHouse environment variables and improves the replication service initialization logic. **ClickHouse remains required** (as intended since PR #2264).

Root Cause: The dashboard always queried ClickHouse by default but did not show the runs because RUN_REPLICATION_CLICKHOUSE_URL should be defined.

### [PR #2035](https://github.com/triggerdotdev/trigger.dev/pull/2035) - Original ClickHouse Implementation (May 2024)
- Introduced PostgreSQL → ClickHouse logical replication
- Initially used a single `CLICKHOUSE_URL` for all operations

### [PR #2264](https://github.com/triggerdotdev/trigger.dev/pull/2264) - Bulk Actions 2.0 (July 2024)
- Migrated all run listings to ClickHouse (`NextRunListPresenter`)
- Made `CLICKHOUSE_URL` **required** - correct strategic decision
- Quote: "it is definitely required after this PR" - @matt-aitken

### Current State
At some point, `RUN_REPLICATION_CLICKHOUSE_URL` was introduced as a duplicate, creating unnecessary complexity.

### 1. Remove Redundant `RUN_REPLICATION_CLICKHOUSE_URL`
- Consolidate back to single `CLICKHOUSE_URL` (original #2035 design)
- Both read and write operations use the same URL
- **No architectural benefit** to having separate URLs (they pointed to same instance)

**Benefits:**
- ✅ Semantic correctness: feature flag → configuration dependency
- ✅ Avoids creating objects unnecessarily when disabled
- ✅ Clearer log messages
- ✅ Removes redundant nested conditional

## ✅ Checklist

---

## Testing
I performed several tasks

---

## Changelog

---

## Screenshots
image

💯


nicktrn (changes requested):
Thanks @LeoKaynan but let's not touch the webapp here. We try to follow the same pattern as we do for redis so we can use different URLs for different services if needed.

It's fine to fix this for self-hosters if there are any bugs though!

One thing to consider is that people have been known to use old versions of the compose files. The clickhouse env vars are present in the latest version:
https://github.com/triggerdotdev/trigger.dev/blob/63b6fc93fa694346e76b805c11f6254d6979904f/hosting/docker/webapp/docker-compose.yml#L64-L69

If there are any helm issues it would be worth looking at this block:
https://github.com/triggerdotdev/trigger.dev/blob/63b6fc93fa694346e76b805c11f6254d6979904f/hosting/k8s/helm/templates/_helpers.tpl#L420-L435

LeoKaynan:
Thanks for the feedback @nicktrn 

I understand now that the separation of URLs is intentional to follow the Redis pattern and allow flexibility for different services. I'll close this PR.

The consolidation seemed like a simplification but I see the architectural reasoning behind keeping them separate for future scalability scenarios.

Thanks for clarifying! 👍

You might have to be careful here, uses ClickHouse by default:
https://github.dev/triggerdotdev/trigger.dev/blob/a445af1b7998b91c2ebe765104ff5faeb97780d9/apps/webapp/app/services/runsRepository/runsRepository.server.ts#L133

RUN_REPLICATION_CLICKHOUSE_URL is optional.

we need to update .env.example, can I open a PR for this?

PR #2624, opened 2025-10-22
fix #2629: avoid multiple rerenders when tags is an array

### LINKED ISSUE #2629: bug: Infinite re-renders when using react hooks with tags

// not sure if applicable here

### Describe the bug

`Tag` is used in the array of dependencies of multiple `useEffect` hooks inside the custom `useRealtimeRunsWithTag` hook. This is causing infinite realtime connections to be created and eating all my browser's memory

Image

### Reproduction repo

Not applicable

### To reproduce

You just need to run `useRealtimeRunsWithTag` with an array of tags to reproduce the issue

### Additional information

_No response_

tonibardina:
## ✅ Checklist

- [x] The PR title follows the convention
- [x] I ran and verified the fix works as expected

---

## 🧪 Testing

I ran the project locally and confirmed that infinite re-renders are no longer happening.  
The subscription now initializes once and remains stable, even when using tag arrays.

---

## 🧾 Changelog

Fixed an issue causing infinite re-renders by ensuring the `tag` dependency in the `useRealtimeRunsWithTag` hook is stable.  
React was previously interpreting a new array reference each render (due to shallow comparison), which triggered unnecessary re-subscriptions.

---

## 📸 Screenshots

_No visual changes — purely internal stability improvement._

💯

fixes #2629

### DIFF
--- a/apps/supervisor/src/services/failedPodHandler.ts
+++ b/apps/supervisor/src/services/failedPodHandler.ts
@@ -25,6 +25,7 @@ export class FailedPodHandler {
 
   private readonly informer: Informer<V1Pod>;
   private readonly reconnectIntervalMs: number;
+  private reconnecting = false;
 
   // Metrics
   private readonly register: Registry;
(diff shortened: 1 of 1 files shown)


tonibardina:
NOTE: temporary fixed it by memoizing my tags. Altho it works I dont think it should be the real fix as it is forcing your users to unnecessarily memoize their tags

PR #2628, opened 2025-10-22
feat(deployments): support local builds in cloud

myftija:
This PR adds support for deploying locally built images to cloud. It uses the
same path as the self-hosting flow and adds the necessary adaptations for
authenticating and pushing to the cloud registry.

### DIFF
--- a/apps/webapp/package.json
+++ b/apps/webapp/package.json
@@ -117,7 +117,7 @@
     "@trigger.dev/core": "workspace:*",
     "@trigger.dev/database": "workspace:*",
     "@trigger.dev/otlp-importer": "workspace:*",
-    "@trigger.dev/platform": "0.0.0-prerelease-ecr-20251021203336",
+    "@trigger.dev/platform": "1.0.19",
     "@trigger.dev/redis-worker": "workspace:*",
     "@trigger.dev/sdk": "workspace:*",
     "@types/pg": "8.6.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -460,8 +460,8 @@ importers:
         specifier: workspace:*
         version: link:../../internal-packages/otlp-importer
       '@trigger.dev/platform':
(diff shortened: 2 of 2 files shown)


apps/webapp/package.json
```diff
@@ -117,7 +117,7 @@
     "@trigger.dev/core": "workspace:*",
     "@trigger.dev/database": "workspace:*",
     "@trigger.dev/otlp-importer": "workspace:*",
-    "@trigger.dev/platform": "1.0.18",
+    "@trigger.dev/platform": "0.0.0-prerelease-ecr-20251021203336",
```
myftija: to be updated to a proper release before merging

packages/cli-v3/src/deploy/buildImage.ts
```diff
@@ -414,6 +425,64 @@ async function localBuildImage(options: SelfHostedBuildImageOptions): Promise<Bu
...
+      }
+    );
+
+    loginProcess.process?.stdin?.write(credentials.password);
+    loginProcess.process?.stdin?.end();
+
+    for await (const line of loginProcess) {
+      errors.push(line);
+      logger.debug(line);
+      options.onLog?.(line);
```
nicktrn: that's a whole lotta logs!

myftija: good point, removed

packages/cli-v3/src/deploy/buildImage.ts
```diff
@@ -474,6 +542,11 @@ async function localBuildImage(options: SelfHostedBuildImageOptions): Promise<Bu
   }
 
   if (buildProcess.exitCode !== 0) {
+    if (registryHost) {
```
nicktrn: might want to make this var name more explicit to show it's only for cloud, I was worried for a sec it would log out self-hosters

nicktrn (changes requested):
smol changes

PR #2644, opened 2025-10-29
feat(react-hooks): add TriggerChatTransport for AI SDK useChat

### LINKED ISSUE #2634: feat: AI SDK ChatTransport

AI SDK v5 exposes a [custom transport](https://ai-sdk.dev/docs/ai-sdk-ui/transport) for use with the `useChat` hook

I’m wondering if it would be possible to use trigger.dev in the transport layer

The default transport of course uses HTTP and there is an example of using [WebSockets](https://github.com/vercel/ai/pull/9559/files) floating around

it seems like it would be possible to trigger the task and then subscribe to the metadata stream and leverage the realtime hooks

Not sure how else we could use the `useChat` hook with trigger

### Additional information

_No response_

jakevollkommer:
Closes #2634

## ✅ Checklist

---

## Summary

Adds a custom transport implementation for AI SDK's `useChat` hook that integrates with Trigger.dev background tasks. This enables long-running AI conversations by triggering tasks, subscribing to realtime run updates via Electric SQL, and streaming AI responses via Server-Sent Events (SSE).

## Changes

- **New Hook**: `useTriggerChat` - A drop-in replacement for AI SDK's `useChat` that works with Trigger.dev tasks
- **Transport Class**: `TriggerChatTransport` - Custom transport implementation following [AI SDK's transport pattern](https://ai-sdk.dev/docs/ai-sdk-ui/transport)
- **Dependencies**: Added `ai` (^5.0.82), `@ai-sdk/react` (^2.0.14), and `eventsource-parser` (^3.0.0)

### Architecture

- Uses [Electric SQL ShapeStream](https://electric-sql.com/docs/api/clients/typescript#shapestream) for realtime run status updates
- Uses [Server-Sent Events (SSE)](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) with `EventSourceParserStream` for streaming AI responses
- Follows the same patterns as existing `useRealtime` hook

### Developer Requirements

**Important:** Developers must create their own server action to trigger tasks. Since `useTriggerChat` is a client-side hook, it cannot directly call `tasks.trigger()` (which requires server-side execution). The `triggerTask` option expects a server action that:

1. Accepts the task identifier and payload
2. Calls `tasks.trigger()` on the server
3. Returns `{ success: true, runId, publicAccessToken }`

### Error Handling

- Gracefully handles stream disconnections and abort signals
- Warns on unparseable SSE chunks without breaking the stream
- Only closes controller when run finishes (not when individual streams end)

---

## Testing

Manually tested with a local project using `pnpm patch` to verify:

- ✅ Task triggering and run creation
- ✅ Realtime run status updates
- ✅ SSE streaming of AI responses
- ✅ Multiple concurrent streams
- ✅ Graceful handling of stream completion
- ✅ TypeScript compilation

### Usage Example

**1. Define your Trigger.dev task** (e.g. `src/trigger/chat.ts`):

```typescript
import { metadata, task } from "@trigger.dev/sdk/v3";
import { streamText } from "ai";
import { openai } from "@ai-sdk/openai";

export const chatTask = task({
  id: "chat",
  run: async ({ messages }: { messages: Array<{ role: string; content: string }> }) => {
    const result = streamText({
      model: openai("gpt-5"),
      messages,
    });

    // CRITICAL: Stream the result to the client using metadata.stream()
    // The stream key MUST match the streamKey option in useTriggerChat (default: "chat")
    await metadata.stream("chat", result.toUIMessageStream());

    const text = await result.text;
    return { text };
  },
});
```

**2. Create a server action** (e.g. `src/actions.ts`):

```typescript
"use server";

import { tasks } from "@trigger.dev/sdk/v3";

export async function triggerChatTask(task: string, payload: unknown) {
  const handle = await tasks.trigger(task, payload);
  return {
    success: true,
    runId: handle.id,
    publicAccessToken: handle.publicAccessToken,
  };
}
```

**3. Use the hook in your component** (e.g. `src/components/Chat.tsx`):

```typescript
"use client";

import { useTriggerChat } from "@trigger.dev/react-hooks";
import { chatTask } from "../trigger/chat";
import { triggerChatTask } from "../actions";

export function Chat() {
  const { messages, input, handleInputChange, handleSubmit } = useTriggerChat({
    triggerTask: triggerChatTask,
  });

  return (
    <form onSubmit={handleSubmit}>
      <div>
        {messages.map((msg) => (
          <div key={msg.id}>
            <strong>{msg.role}:</strong> {msg.content}
          </div>
        ))}
      </div>
      <input value={input} onChange={handleInputChange} />
      <button type="submit">Send</button>
    </form>
  );
}
```

---

## Changelog

Added `useTriggerChat` hook to `@trigger.dev/react-hooks` that provides AI SDK `useChat` integration with Trigger.dev background tasks. Enables long-running AI conversations with realtime streaming via custom transport implementation.

**New exports:**

- `useTriggerChat` - Hook for AI chat integration
- `TriggerChatTransport` - Custom transport class
- `TriggerChatTransportOptions` - Transport configuration type
- `TriggerChatTaskPayload` - Task payload type

**New dependencies:**

- `ai@^5.0.82`
- `@ai-sdk/react@^2.0.14`
- `eventsource-parser@^3.0.0`

---

## Resources

- [AI SDK Custom Transports Documentation](https://ai-sdk.dev/docs/ai-sdk-ui/transport)
- [AI SDK useChat Hook](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot)
- [Trigger.dev Realtime API](https://trigger.dev/docs/realtime/introduction)
- [Electric SQL ShapeStream](https://electric-sql.com/docs/api/clients/typescript#shapestream)
- [Server-Sent Events Spec](https://html.spec.whatwg.org/multipage/server-sent-events.html)

---

## Screenshots

_N/A - This is a developer-facing hook with no UI_

💯

### DIFF
--- a/packages/react-hooks/src/hooks/useTriggerChat.ts
+++ b/packages/react-hooks/src/hooks/useTriggerChat.ts
@@ -490,39 +490,6 @@ async function streamDataFromTrigger(params: {
   }
 }
 
-function checkRequiredPeerDependencies() {
-  const requiredPackages = [
-    "@ai-sdk/react",
-    "ai",
-    "@electric-sql/client",
-    "eventsource-parser",
-  ];
-
-  const missing = requiredPackages.filter((pkg) => {
-    try {
-      require.resolve(pkg);
-      return false;
-    } catch {
-      return true;
-    }
-  });
-
-  if (!missing.length) return;
-
-  const packages = missing.join(" ");
-
-  throw new Error(
-    `useTriggerChat requires the following packages:\n${missing
-      .map((pkg) => `  - ${pkg}`)
-      .join("\n")}\n\n` +
-      `Install them with:\n` +
-      `  npm install ${packages}\n` +
-      `  yarn add ${packages}\n` +
-      `  pnpm add ${packages}\n` +
-      `  bun add ${packages}`,
-  );
-}
-
 type UseTriggerChatOptions = {
   transportOptions: TriggerChatTransportOptions;
 } & Omit<ChatInit<UIMessage>, "transport">;
@@ -610,7 +577,6 @@ type UseTriggerChatOptions = {
 export function useTriggerChat(
   options: UseTriggerChatOptions,
 ): UseChatHelpers<UIMessage> {
-  checkRequiredPeerDependencies();
   const { transportOptions, ...chatOptions } = options;
 
   return useChat({


packages/react-hooks/package.json
```diff
@@ -37,7 +37,11 @@
     "check-exports": "attw --pack ."
   },
   "dependencies": {
+    "@ai-sdk/react": "^2.0.14",
```
jakevollkommer: ideally we don't have to include this dependency unless devs are using the `useTriggerChat` hook

packages/react-hooks/package.json
```diff
@@ -37,7 +37,11 @@
...
+    "@electric-sql/client": "1.0.14",
     "@trigger.dev/core": "workspace:^4.0.5",
+    "ai": "^5.0.82",
```
jakevollkommer: as above

packages/react-hooks/src/hooks/useTriggerChat.ts
```diff
@@ -0,0 +1,556 @@
...
+export interface TriggerChatTransportOptions<
+  TPayload extends TriggerChatTaskPayload = TriggerChatTaskPayload,
+> {
+  /**
+   * The Trigger.dev task identifier to trigger
+   * @default "chat"
+   *
+   * @remarks
+   * **CRITICAL:** Your Trigger.dev task MUST call `metadata.stream()` with the AI SDK stream.
+   * The stream key used in `metadata.stream()` must match the `streamKey` option (default: "chat").
```
jakevollkommer, line 42: I'm not sure if there may be a better place to put this piece of documentation

the tricky bit here is there are 2 external files required to enable this hook 
1. the trigger.dev task definition itself, which **must** call a `stream()` method from `ai` and forward the stream to `metadata.stream()`
2. the server action which invoked the trigger.dev task server-side

packages/react-hooks/src/hooks/useTriggerChat.ts
```diff
@@ -0,0 +1,556 @@
...
+   *       model: openai("gpt-4"),
+   *       messages,
+   *     });
+   *
+   *     // CRITICAL: Stream to client using metadata.stream()
+   *     await metadata.stream("chat", result.toUIMessageStream());
+   *
+   *     return { text: await result.text };
+   *   },
+   * });
```
jakevollkommer, line 63: similar to the above, it's a bit awkward adding this example here, because the `task` parameter actually just needs the `string` which is the `id` of the task (`"chat"` in the above example)

still unsure if this is the best way to document this

packages/react-hooks/src/hooks/useTriggerChat.ts
```diff
@@ -0,0 +1,556 @@
...
+      return undefined;
+    }
+  }
+
+  if (typeof metadata === "object") {
+    return metadata as ParsedMetadata;
+  }
+
+  return undefined;
+}
```
jakevollkommer, line 361: normally I would use zod for this, but I think this matches patterns in other hooks

packages/react-hooks/src/hooks/useTriggerChat.ts
```diff
@@ -0,0 +1,556 @@
...
+ *
+ *   return (
+ *     <form onSubmit={handleSubmit}>
+ *       {messages.map(m => <div key={m.id}>{m.role}: {m.content}</div>)}
+ *       <input value={input} onChange={handleInputChange} />
+ *       <button type="submit">Send</button>
+ *     </form>
+ *   );
+ * }
+ * ```
```
jakevollkommer, line 575: the above comments apply here as well

ericallam:
@jakevollkommer thanks for the PR but I'm going to be closing this as we we've made a bunch of changes around realtime streams recently and we're also planning on releasing something in this area soon, but it depends on doing a couple bits of work first.

jakevollkommer:
> @jakevollkommer thanks for the PR but I'm going to be closing this as we we've made a bunch of changes around realtime streams recently and we're also planning on releasing something in this area soon, but it depends on doing a couple bits of work first.

@ericallam no worries! I actually adapted this internally to use realtime v2 and was planning on committing them here, happy to share if it might be helpful

PR #2655, opened 2025-11-05
fix(build/prisma): support custom schema directories and Prisma 6.6+ …

### LINKED ISSUE #1926: [TRI-5172] Compatibility with Prisma 6.6+
In Prisma 6.6.0+ they require an `output` path and no longer generate the Prisma client in `node_modules`.

[Details here](https://github.com/prisma/prisma/releases/tag/6.6.0).

This is going to cause problems with our [official prisma extension](https://trigger.dev/docs/config/extensions/prismaExtension) which was built to deal with the old way.

It seems like the way to fix this would be to do this from their docs:

> pass the the --schema option to your Prisma CLI command (e.g. prisma migrate dev --schema ./prisma/schema)

We don't pass the --schema for folders right now I don't think

<sub>[TRI-5172](https://linear.app/triggerdotdev/issue/TRI-5172/prisma-with-66)</sub>

cameronmema:
…compatibility

This change makes the Prisma build extension work with any schema directory name (e.g. "schema", "models", "db", "prisma-schemas") and ensures compatibility with Prisma 6.6+.

Configuration:
Always point the `schema` option to your main schema file, not the folder:

  prismaExtension({
    version: "6.18.0",
    schema: "../../packages/prisma/src/models/schema.prisma",  // ← Point to the file
    migrate: false,
  })

The main schema file can have any name (schema.prisma, main.prisma, etc.) and can live in any directory. It does NOT have to be called "schema.prisma".

How it works:
- Auto-detects a "schema folder" by checking the parent directory of the schema file for multiple .prisma files. If 2+ .prisma files exist in that directory, we treat it as a folder-based schema setup.
- For folder-based schemas (multiple files), all .prisma files in that directory are copied to ./prisma/schema/ in the build output. For Prisma >= 6.6, we run: prisma generate --schema=./prisma/schema For older Prisma versions, we omit the --schema flag for backwards compatibility.
- For single-file schemas (only one .prisma file in the directory), the file is copied to ./prisma/schema.prisma and we run: prisma generate --schema=./prisma/schema.prisma

Examples:
  # Folder-based (multiple .prisma files): packages/prisma/src/models/
    ├── schema.prisma       ← Point to this in config
    ├── user.prisma
    └── post.prisma

  # Single file: packages/prisma/
    └── schema.prisma       ← Point to this in config

Backwards compatibility:
- Older Prisma versions (< 6.6) continue to work without the --schema flag for folder setups
- Single-file setups are unaffected
- No breaking changes

Closes #1926

Closes #<issue>

## ✅ Checklist

---

## Testing

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/packages/build/src/extensions/prisma.ts
+++ b/packages/build/src/extensions/prisma.ts
@@ -112,7 +112,12 @@ export class PrismaExtension implements BuildExtension {
 
     context.logger.debug(`PrismaExtension is generating the Prisma client for version ${version}`);
 
-    const usingSchemaFolder = dirname(this._resolvedSchemaPath).endsWith("schema");
+    // Auto-detect schema folder by checking for multiple .prisma files
+    const schemaDir = dirname(this._resolvedSchemaPath);
+    const prismaFilesInDir = await readdir(schemaDir)
+      .then((files) => files.filter((file) => file.endsWith(".prisma")))
+      .catch(() => []);
+    const usingSchemaFolder = prismaFilesInDir.length > 1;
 
     const commands: string[] = [];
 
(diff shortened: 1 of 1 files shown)


ericallam:
This should be fixed now in `4.1.1`

Issue #862, opened 2024-01-20
[TRI-1882] Python support


matt-aitken:
It would be great to offer support for Python.

Python has better support for many Machine Learning uses cases:
- Many ML libraries are only Python, or more complete in Python
- Some services have ML models that only work from Python (e.g. OpenAI's whisper model)
- Data Science and Engineering tools like pandas

It would be even better if you could mix languages. Have files alongside each other that are different languages but can declare tasks.

<sub>[TRI-1882](https://linear.app/triggerdotdev/issue/TRI-1882/python-support)</sub>

matt-aitken:
Cloudflare Workers transpiles languages into JavaScript.

- Cloudflare blog post: https://blog.cloudflare.com/cloudflare-workers-announces-broad-language-support
- Languages that can compile to JS: https://github.com/jashkenas/coffeescript/wiki/List-of-languages-that-compile-to-JS
- Transcrypt is used to compile Python to JS: https://www.transcrypt.org/docs/html/installation_use.html

listenaddress:
Hey Matt, I'd love to use trigger.dev with Python! Any idea when this might happen?

ologbonowiwi:
Having this would be great! I want to do a job to run https://github.com/joshiayush/inb automatically and trigger.dev would be a great option if it offers support for Python.

MaleicAcid:
+1 really need this.

montaign:
pls advise on timeline

LeEricCH:
+1 would really enjoy that!

chanmathew:
+1! Big use case for this is running python based crawlers and other AI inference tasks.

tgirotto:
+1 it would be a great use case for gis applications as well :)

nullmastermind:
+1

archodev:
+1

saschaseniuk:
https://feedback.trigger.dev/p/python-support-2

You can upvote for python support to make it happen 😍

BaileySimrell:
+1 this would be awesome .. just upvoted!

peter-hanzo:
Here is the list that I want to use in my project. Many of these libraries I find very useful in any project:

---

### **Asynchronous and HTTP Requests**  
- **aiohttp**: Asynchronous HTTP client and server for Python.  
- **async-timeout**: Manage timeouts in asynchronous operations.  
- **httpx**: Asynchronous HTTP requests, an alternative to `aiohttp`.  
[...]

PR #2661, opened 2025-11-09
feat(python): add Python worker support - POC

beydogan:
Related #862 and [full python support from roadmap](https://feedback.trigger.dev/p/full-python-support)

This is a POC and discussion PR for full Python worker support. I leveraged Claude Code (supervised, reviewed and tested) to create a quick POC to validate the approach before investing significant time, since I wasn't sure if you'd be okay with this direction. Sorry about the PR size, but this was the minimum version needed to make it functional.

I used gRPC for Node↔Python IPC, which opens the door for more runtimes like Go, Ruby, etc.

I'm aware it needs more polishing, cleanup, testing, and additional features to be production-ready.  I'm open to discussion and feedback. Please let me know what you think! 

Feel free to close if it doesn't fit your vision or approach, but I hope we can make it work because I believe full Python support would benefit the community. I've found current python support very limited and hard to work with in complex workflows.

## What works? (only dev mode)

- Task indexing and registering
- Running simple python only tasks with payloads and configs like max_duration etc.

## What needs to be done?

- Containerfile generation for managed and self-hosted support
- Worker log streaming
- SDK support for [triggering](https://trigger.dev/docs/triggering)
- Polishing and cleaning up python runtime related code
- Probably more stuff.

Some flows to visualize the differences:
[Build Time](https://github.com/user-attachments/assets/f0c0a3c2-fbfe-42d1-a1ab-513b7e065829):
[Index Phase](https://github.com/user-attachments/assets/963eae4c-d0c1-48e3-ba0f-854d0da48b42)
[Execution Phase](https://github.com/user-attachments/assets/7fd18a68-46f9-48ab-8951-445b3a3cc1cc)

## ✅ Checklist

---

## Testing

Tested with references/hello-world-python:
  - Task registration and indexing
  - Task execution via gRPC
  - Structured logging
  - Error handling
  - Build and dev commands

---

## Changelog

Added Python worker support with gRPC communication.

  - Implemented Python SDK with task decorator, IPC layer, execution context
  - Created index and run workers for Python task discovery and execution
  - Added Node.js stdio IPC bridge for spawning Python processes
  - Integrated Python runtime into task execution flow
  - Added Python build system with file discovery and requirements.txt handling
  - Migrated workers from stdio to gRPC for better reliability
  - Fixed max_duration conversion (seconds → ms) and PYTHONPATH handling
  - Disabled bytecode cache in dev mode for fresh reloads
---

## Screenshots

Indexed tasks:
image

Samples from running the [hello-task-v2](https://github.com/beydogan/trigger.dev/blob/python_workers/references/hello-world-python/src/trigger/example_task.py#L8)

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "clean:node_modules": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +",
     "clean:python": "find . -name '*.pyc' -delete && find . -name '__pycache__' -type d -exec rm -rf '{}' + 2>/dev/null || true",
     "clean:all": "pnpm run clean && pnpm run clean:python",
+    "rebuild:cli": "pnpm run clean:python && pnpm run build --filter trigger.dev",
     "typecheck": "turbo run typecheck",
     "test:e2e": "playwright test",
     "test:e2e:ui": "playwright test --ui",
--- a/packages/cli-v3/src/entryPoints/python/managed-index-worker.py
+++ b/packages/cli-v3/src/entryPoints/python/managed-index-worker.py
@@ -99,7 +99,9 @@ def collect_task_metadata() -> List[Dict[str, Any]]:
             if task_meta.queue:
                 task_dict["queue"] = task_meta.queue.model_dump()
             if task_meta.maxDuration is not None:
+                # Already in milliseconds from task decorator
                 task_dict["maxDuration"] = task_meta.maxDuration
+                logger.debug(f"Task {task_id} maxDuration: {task_meta.maxDuration}ms")
 
             tasks.append(task_dict)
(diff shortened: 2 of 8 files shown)


ericallam:
@beydogan this is very interesting! Thanks for the proposal and showing a working proof of concept! I think the very first thing we would need to do before introducing a full python runtime would be to produce a full python SDK (similar to our `@trigger.dev/sdk`). Additionally, when introducing something like this the consideration isn't only the code being committed, but then we would have to keep the python runtime up to date whenever working on something new. This would be too big of a burden at the moment. For that reason I'm going to close this PR for now, but hopefully one day we'll be able to resurrect it and provide a full python runtime like proposed here.

PR #2663, opened 2025-11-10
feat(deployments): stream build server logs

myftija:
This PR adds logs streaming for deployments triggered by the GitHub
integration. We use S2 and its typescript sdk to do this. A short-lived
read-only token is used to read the stream directly on the client side.

https://github.com/user-attachments/assets/0afcde78-8c88-4fb9-8446-9401378919b6

### DIFF
--- a/apps/webapp/app/components/primitives/DateTime.tsx
+++ b/apps/webapp/app/components/primitives/DateTime.tsx
@@ -15,6 +15,7 @@ type DateTimeProps = {
   showTooltip?: boolean;
   hideDate?: boolean;
   previousDate?: Date | string | null; // Add optional previous date for comparison
+  hour12?: boolean;
 };
 
 export const DateTime = ({
@@ -24,6 +25,7 @@ export const DateTime = ({
   includeTime = true,
   showTimezone = false,
   showTooltip = true,
+  hour12 = true,
 }: DateTimeProps) => {
   const locales = useLocales();
   const [localTimeZone, setLocalTimeZone] = useState<string>("UTC");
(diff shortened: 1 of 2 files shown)


apps/webapp/app/presenters/v3/DeploymentPresenter.server.ts
```diff
@@ -142,7 +145,27 @@ export class DeploymentPresenter {
       ? ExternalBuildData.safeParse(deployment.externalBuildData)
       : undefined;
 
+    const s2 = new S2({ accessToken: env.S2_ACCESS_TOKEN });
+    const projectS2AccessToken = await s2.accessTokens.issue({
```
myftija: Will cache project-scoped tokens in redis to avoid generating one for each request.

apps/webapp/app/env.server.ts
```diff
@@ -1201,6 +1201,9 @@ const EnvironmentSchema = z
     EVENT_LOOP_MONITOR_UTILIZATION_SAMPLE_RATE: z.coerce.number().default(0.05),
 
     VERY_SLOW_QUERY_THRESHOLD_MS: z.coerce.number().int().optional(),
+
+    S2_ACCESS_TOKEN: z.string(),
+    S2_DEPLOYMENT_LOGS_BASIN_NAME: z.string(),
```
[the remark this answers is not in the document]
myftija: Yeah, need to keep them optional until we have the oss s2 version.

apps/webapp/app/components/primitives/DateTime.tsx
```diff
@@ -76,6 +77,7 @@ export function formatDateTime(
     minute: includeTime ? "numeric" : undefined,
     second: includeTime && includeSeconds ? "numeric" : undefined,
     timeZone,
+    hour12: false,
```
myftija: This will affect other places where we show timestamps too.

myftija: made this a param instead to avoid changing the format in other places.

PR #2702, opened 2025-11-21
feat(deployments): `--native-build-server` support for the deploy command

myftija:
This PR adds support for CLI deployments using the native build server.

**Background**

The deployment command currently does the following:
- bundles the code
- submits the build context to our external build provider and waits for the build
- triggers deployment state transitions using the platform API

Upstream build provider outages cause issue with deployments, potentially blocking deployments entirely. We recently introduced the `--force-local-build` flag as a fallback to enable deployment without a dependency on the upstream build provider, though it requires users to have docker in their systems. This PR continues that work by providing a remote build  path which uses our own build server and does not rely on the external provider.

**Changes in this PR**

Introduced the new `--native-build-server` flag, which does the following:
- scans all files relevant for the Trigger deployment and evaluates ignore rules
- packages it up in an archive and uploads it as a deployment artifact
- queues the deployment and triggers the build
- streams logs from the build server

This no longer relies on external build services. Also deployment state transitions happen on the server-side, giving us more flexibility to evolve the flow and schemas of related deployment API endpoints. In general it gives us better control of the whole build and deployment process. This path will eventually become the default.

The `--detach` flag is also new, allowing to trigger deployments without waiting for the result.

The deployment artifacts are uploaded via pre-signed URLs to avoid unnecessary load on the platform. The new `/artifacts` endpoint generates the pre-signed URLs; size limits are enforced on s3. This endpoint is deliberately generic, we could extend it in the future to upload other artifacts client-side in a similar way, e.g., large payload packets.

### DIFF
--- a/packages/cli-v3/src/commands/deploy.ts
+++ b/packages/cli-v3/src/commands/deploy.ts
@@ -194,7 +194,7 @@ export function configureDeployCommand(program: Command) {
       .addOption(new CommandOption("--plain", "Plain output").hideHelp())
       .action(async (path, options) => {
         await handleTelemetry(async () => {
-          await printStandloneInitialBanner(true);
+          await printStandloneInitialBanner(true, options.profile);
           await deployCommand(path, options);
         });
       })
--- a/.changeset/brown-pots-beg.md
+++ b/.changeset/brown-pots-beg.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": patch
+---
+
+fix(otel): exported logs and spans will now have matching trace IDs
--- a/.changeset/healthy-cherries-joke.md
+++ b/.changeset/healthy-cherries-joke.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": patch
+---
+
+fix(cli): header will always print the correct profile
--- a/.changeset/polite-eels-divide.md
+++ b/.changeset/polite-eels-divide.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": patch
+---
+
+feat(cli): implements content-addressable store for the dev CLI build outputs, reducing disk usage
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,5 @@
 name: 🦋 Changesets Release
 
-permissions:
-  contents: write
-
 on:
   push:
     branches:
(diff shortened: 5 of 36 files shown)


apps/webapp/package.json
```diff
@@ -114,7 +118,7 @@
     "@trigger.dev/core": "workspace:*",
     "@trigger.dev/database": "workspace:*",
     "@trigger.dev/otlp-importer": "workspace:*",
-    "@trigger.dev/platform": "1.0.20",
+    "@trigger.dev/platform": "0.0.0-prerelease-20251121183112",
```
myftija: update to the actual release before merging

.changeset/config.json
```diff
@@ -1,11 +1,6 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@2.2.0/schema.json",
-  "changelog": [
```
myftija, line 3: remove

.changeset/proud-birds-change.md
```diff
@@ -0,0 +1,6 @@
+---
+"trigger.dev": patch
```
myftija, line 2: could also be a minor version bump

myftija: upgraded to minor

apps/webapp/app/v3/services/deploymentIndexFailed.server.ts
```diff
@@ -66,6 +74,21 @@ export class DeploymentIndexFailed extends BaseService {
       },
     });
 
+    const deploymentService = new DeploymentService();
```
nicktrn, line 77: this file is v3 only

packages/cli-v3/src/deploy/buildImage.ts
```diff
@@ -686,6 +686,7 @@ async function generateBunContainerfile(options: GenerateContainerfileOptions) {
     parseGenerateOptions(options);
 
   return `# syntax=docker/dockerfile:1
+# check=skip=SecretsUsedInArgOrEnv
```
nicktrn, line 689: 👌

packages/cli-v3/src/commands/deploy.ts
```diff
@@ -827,6 +893,410 @@ async function initializeOrAttachDeployment(
...
+
+  const $queuedSpinner = spinner();
+  $queuedSpinner.start("Build queued");
+
+  const abortController = new AbortController();
+
+  const s2 = new S2({ accessToken: eventStream.s2.accessToken });
+  const basin = s2.basin(eventStream.s2.basin);
+  const stream = basin.stream(eventStream.s2.stream);
+
```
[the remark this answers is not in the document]
myftija: those methods do not throw.

PR #2721, opened 2025-12-01
fix(clickhouse): ensure start_time is never older than X ms to prevent old partition merge issues

ericallam:
This is an alternate solution to the fix made in #2719 because task_events_v1 can cause issues even if hardly any events are being sent there. 

## Core problem with task_events_v1

1. **Insert arrives** with `start_time` from 3 days ago
2. **New tiny part created** in partition `20251028` (e.g., 10 rows, 3KB)
3. **That partition already has a massive merged part** (e.g., 30GB)
4. **To merge the tiny part**, ClickHouse must rewrite the entire 30GB + 10 rows
5. **That merge takes hours**, during which more tiny parts arrive
6. **Repeat forever**

The old partitions will **never stabilise** because we keep inserting late-arriving data into them. Additionally, large merges like this can overwhelm a ch replica and prevent merges on other tables, causing a cascading failure.

This change basically attacks the root problem: insert task_events_v1 records with old start times. You can now set the `EVENTS_CLICKHOUSE_START_TIME_MAX_AGE_MS` env var (defaults to 5 minutes) to clamp the start_time to close to now. Along with this we've updated the application logic that merged many events into a single span to always use the earliest span start_time, instead of just the first.


apps/webapp/app/v3/eventRepository/clickhouseEventRepositoryInstance.server.ts
```diff
@@ -34,6 +34,41 @@ function getClickhouseClient() {
     },
     maxOpenConnections: env.EVENTS_CLICKHOUSE_MAX_OPEN_CONNECTIONS,
   });
+<<<<<<< ours
```
myftija: merge conflict marker

PR #2725, opened 2025-12-02
feat(cli): implements content-addressable store for the dev CLI build outputs, reducing disk usage

ericallam:
Implements a content-addressable store for dev CLI build outputs, significantly reducing disk usage when making iterative code changes during development.

## Problem

When running `trigger.dev dev`, each code change creates a new build directory in `.trigger/tmp/build-XXXXX/`. Even minor changes resulted in ~22MB of duplicated files per build because esbuild outputs were being fully copied each time, despite most files (chunks, dependencies) being identical between builds.

## Solution

Introduced a content-addressable store at `.trigger/tmp/store/` that deduplicates build outputs using hardlinks:

1. **Content-addressable storage**: Files are stored by their content hash (using esbuild's built-in `outputFile.hash`), so identical content is only stored once
2. **Hardlinks**: Build directories contain hardlinks to store files instead of copies - same inode means zero additional disk space
3. **Hash caching**: File hashes are stored in `BuildManifest.outputHashes` to avoid recomputation when copying to worker directories

Tested with 12 iterative builds in a reference project:

| Metric | Without Dedup | With Dedup | Savings |
|--------|---------------|------------|---------|
| 12 builds | ~264 MB | 68.7 MB | **195 MB (74%)** |

- First build: 22MB (populates the store)
- Subsequent builds: ~1.7-4.3MB each (only changed files)
- Savings increase with more builds as unchanged files are simply hardlinked

## Backwards Compatibility

- `outputHashes` is optional in `BuildManifest` - falls back to computing hashes if not present
- Store is only used in dev mode, no impact on deploy builds
- Store is cleared when dev session starts fresh (`clearTmpDirs`)

### DIFF
--- a/packages/cli-v3/src/utilities/fileSystem.ts
+++ b/packages/cli-v3/src/utilities/fileSystem.ts
@@ -20,7 +20,7 @@ export async function createFile(
  * Sanitizes a hash to be safe for use as a filename.
  * esbuild's hashes are base64-encoded and may contain `/` and `+` characters.
  */
-function sanitizeHashForFilename(hash: string): string {
+export function sanitizeHashForFilename(hash: string): string {
   return hash.replace(/\//g, "_").replace(/\+/g, "-");
 }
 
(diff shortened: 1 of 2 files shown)


packages/cli-v3/src/utilities/fileSystem.ts
```diff
@@ -16,6 +16,59 @@ export async function createFile(
   return path;
 }
 
+/**
+ * Sanitizes a hash to be safe for use as a filename.
+ * esbuild's hashes are base64-encoded and may contain `/` and `+` characters.
+ */
+function sanitizeHashForFilename(hash: string): string {
```
nicktrn: dupe

PR #2729, opened 2025-12-02
feat(build): Add NeonDB branch resolution for Vercel preview environments

### LINKED ISSUE #2714: bug: syncVercelEnvVars syncs production env var for preview deployment when using NeonDB vercel integration

(Version is 4.1.2, but likely does not matter much)

### Describe the bug

When using the Vercel NeonDB integration, each preview environments gets individual env vars set just before a build starts on Vercel. Those "per preview environment" env vars are **not returned by the Vercel REST API** at `https://api.vercel.com/v8/projects/${projectId}/env?${params}`.

The problem is that syncVercelEnvVars uses the above URL to get both, env var names and their values. The synced env vars do not have the values Vercel uses during the build! Env var are not actually in sync.

To make matters worse, the Vercel NeonDB integration, automatically makes the production db env var available through the above REST API _for the staging environment_! This caused our trigger.dev staging environment to write to our prod db :( (In the Vercel UI this env var is "managed" by the Neon DB integration and cannot even be manually fixed.)

My suggestion is to "fix" this problem by modifying syncVercelEnvVars to only get the names of the env vars from the REST API but then _reading the values from the actual build environment_. Just like "syncEnvVars" does it.

(For what it's worth, follow-ups could also include:
1. To file a feature request with Vercel to ask for a way to get the "per preview environment" env vars.
2. To ask NeonDB to not set the prod url for staging [tbh wtf])

As a workaround, we switched to using "syncEnvVars" for the Neon related env vars.

### Reproduction repo

Sorry, I hope the steps are detailed enough

### To reproduce

- Make a Vercel project
- Use the Neon db Vercel integration
- Build/deploy a Vercel preview environment that also deploys to trigger.dev staging env and uses syncVercelEnvVars
- Observe that the trigger dev tasks in staging are using the production database

### Additional information

_No response_

0ski:
Vercel's NeonDB integration renders database connection environment variables at runtime, which means Trigger.dev cannot directly sync these values during the build process. This change adds support for fetching branch-specific NeonDB connection strings via the Neon API.

feat(build): Add syncNeonEnvVars extension and improve Vercel env var syncing

Add a new `syncNeonEnvVars` build extension for syncing environment variables 
from Neon database projects to Trigger.dev. The extension automatically detects 
branches and builds appropriate PostgreSQL connection strings for non-production 
environments (staging, dev, preview).

Features of `syncNeonEnvVars`:
- Fetches branch-specific database credentials from Neon API
- Generates all standard Postgres connection strings (DATABASE_URL, POSTGRES_URL, 
  POSTGRES_PRISMA_URL, etc.) with both pooled and unpooled variants
- Supports custom database name, role name, and env var prefix options
- Skips automatically in Vercel environments (Neon's Vercel integration handles this)
- Skips for production environments (designed for preview/staging/dev branches)

Improvements to `syncVercelEnvVars`:
- When running in a Vercel build environment (detected via VERCEL env var), 
  values are now read from process.env instead of the Vercel API response
- This ensures the build uses the actual runtime values Vercel provides
- Removed embedded Neon-specific logic (now handled by separate extension)
- Simplified and cleaned up the extension code

Documentation updates for both extensions with usage examples and configuration 
options.

Closes #2714

## ✅ Checklist

---

## Testing

Set up Vercel + Trigger.dev envs, used Vercel's chat-bot-ai template.

### DIFF
--- a/.changeset/funny-planes-remember.md
+++ b/.changeset/funny-planes-remember.md
@@ -2,4 +2,4 @@
 "@trigger.dev/build": patch
 ---
 
-syncVercelEnvVars to skip API and read env vars directly from env.process for Vercel build environments. New syncNeonEnvVars build extension for syncing environment variablesfrom Neon database projects to Trigger.dev. The extension automatically detects branches and builds appropriate PostgreSQL connection strings for non-production environments (staging, dev, preview).
+syncVercelEnvVars to skip API and read env vars directly from env.process for Vercel build environments. New syncNeonEnvVars build extension for syncing environment variablesfrom Neon database projects to Trigger.dev. The extension automatically detects branches and builds appropriate PostgreSQL connection strings for non-production, non-dev environments (staging, preview).


packages/build/src/extensions/core/vercelSyncEnvVars.ts
```diff
@@ -11,11 +11,192 @@ type VercelEnvVar = {
...
+  name: string;
+};
+
+type NeonEndpoint = {
+  id: string;
+  host: string;
+  type: string;
+};
+
+async function fetchNeonBranchEnvVars(options: {
```
felix-quotez: Thanks for you work on this!

I think, there is an opportunity to make this fix non-neon specific and remove the need to fetch env vars from Neon. You can get the _names_ of the env vars from the Vercel env API (it will know about the Neon env var prefix), but then get the _values_ from the current env (like what you have via syncEnvVars). Unless I am missing something, this should give us exactly the values that the Vercel build/run environment has.

In case you had local development in mind while making this change, note that the Vercel integrations are only running in the cloud. So during local development keeping today's behavior seems good.

Benefits are:
- This would work with other Vercel integrations that set env vars per preview environment (let their product deal with complexity).
- No need to maintain a list of Neon specific env var names.
- No need for fragile string processing (newHost.replace(/^([^.]+)\./, "$1-pooler.");)
- No need for VERCEL_NEON_ENV_VAR_PREFIX
- No dependency on the Neon console API

I hope this unsolicited comment is helpful. Best, Felix.

felix-quotez: Little meta thought for the fun of it: If you ship your own Vecel integration [https://feedback.trigger.dev/p/vercel-integration-3](https://feedback.trigger.dev/p/vercel-integration-3) the above suggestion would even sync the trigger.dev branch specific env vars :)

0ski: Hi @felix-quotez, thanks for your feedback.
Unfortunately NeonDB env vars are injected in Vercel's runtime. The `vercelSyncEnvVars` already pulls `env` related `env vars`, however for NeonDB they unfortunately always come from the `production` env (as you noticed).
I set up a really small project and played around with the Vercel's API, it seems that the only way to get correct connection config is thru NeonDB.
I'm thinking about creating separate build extension, to keep it tidy, called `neonSyncEnvVars` what do you think?

felix-quotez: Hi @0ski,

> Unfortunately NeonDB env vars are injected in Vercel's runtime.

On Vercel, the env vars are also available at build time. When deploying a preview environment, the Neon integration first creates a new db branch, and then sets the _correct, non-prod_ env vars _before_ the build runs--and thus before vercelSyncEnvVars runs. You can read the correct non-prod env var values during build time from process.env (not from the Vercel API). That is also why I can use syncEnvVars as a workaround right now (as briefly mentioned in the bug ticket).

I still think, you can use the Vercel API to get the _names_ of the env vars (so you know which ones to sync) but then get _values_ for these env vars from process.env and not the API. This should simplify your change a lot. Does this make sense? (There would be no need to create and maintain neonSyncEnvVars.)

For what it's worth, this is from our Vercel dashboard showing how the integrations are provisioned before the build starts.

Screenshot 2025-12-03 at 14 15 48

Best, Felix

0ski: Hi @felix-quotez,

oh I think we're not on the same page. In your case, the trigger.dev build is invoked from Vercel's CI/CD, hence Vercel handles all the `env vars` for you, there is no need for trigger.dev to pull that information from API!
What I tried, is to call the `vercelSyncEnvVars` from my local console, and to properly sync these variables, I need to pull the info from NeonDB, as there is no way to do it from Vercel API.

Now I think I get it! It would be great to handle the use-case from any place, and for Vercel, do what you described - pull just the names, and take them directly from `process.env`. That's great suggestion, thank you! ❤️ Gonna implement it.

Do you know a way for my code to check if it's being called from the Vercel build env?

Cheers,
Oskar

0ski: I figured it out, changes pushed

docs/config/extensions/syncEnvVars.mdx
```diff
@@ -114,3 +122,71 @@ export default defineConfig({
...
+<Note>
+  You need to set the `NEON_ACCESS_TOKEN` and `NEON_PROJECT_ID` environment variables, or pass them
+  as arguments to the `syncNeonEnvVars` build extension. You can generate a `NEON_ACCESS_TOKEN` in
+  your Neon [dashboard](https://console.neon.tech/app/settings/api-keys).
+</Note>
+
+<Note>
+  When running the build from a Vercel environment (determined by checking if the `VERCEL`
+  environment variable is present), this extension is skipped entirely. This is because Neon's
+  Vercel integration already handles environment variable synchronization in Vercel environments.
```
felix-quotez, line 150: _If syncVercelEnvVar is used_, under which conditions does it make sense to _also_ use syncNeonEnvVars?

docs/config/extensions/syncEnvVars.mdx
```diff
@@ -80,6 +80,14 @@ The `syncVercelEnvVars` build extension syncs environment variables from your Ve
   the project with the environment variables you want to sync.
 </Note>
 
+<Note>
+  When running the build from a Vercel build environment (e.g., during a Vercel deployment), the
+  environment variable values will be read from `process.env` instead of fetching them from the
+  Vercel API. This is determined by checking if the `VERCEL` environment variable is present. The
+  API is still used to determine which environment variables are configured for your project, but
+  the actual values come from the local environment.
```
felix-quotez: You are describing the "what" but adding the "why" would be great. Suggestion: add something like "Reading values from process.env allows to use values that Vercel integrations (such as the Neon integration) set per preview deployment in the "Provisioning Integrations" phase that happens just before the Vercel build starts".

packages/build/src/extensions/core/vercelSyncEnvVars.ts
felix-quotez: Great, thank you! Left two comment in the code.

docs/config/extensions/syncEnvVars.mdx
0ski: Good that you mention it, I forgot to add it to docs, and made a note yesterday that I need to include it here too

docs/config/extensions/syncEnvVars.mdx
0ski: That's good suggestion. Gonna include it! Thank you

docs/config/extensions/syncEnvVars.mdx
```diff
@@ -114,3 +122,71 @@ export default defineConfig({
...
+import { syncNeonEnvVars } from "@trigger.dev/build/extensions/core";
+
+export default defineConfig({
+  project: "<project ref>",
+  // Your other config settings...
+  build: {
+    extensions: [
+      syncNeonEnvVars({
+        projectId: "your-neon-project-id",
+        neonAccessToken: "your-neon-access-token",
```
nicktrn, line 193: maybe we should add a warning via jsdoc to this field to say you shouldn't hardcode secrets. I can definitely see the case where you'd pass an env var here so we should keep it.

packages/build/src/extensions/core/neonSyncEnvVars.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    const environment = environmentMap[ctx.environment as keyof typeof environmentMap];
+
+    if (!environment) {
+      throw new Error(
+        `Invalid environment '${ctx.environment}'. Expected 'prod', 'staging', 'dev', or 'preview'.`
+      );
+    }
+
+    try {
```
nicktrn, line 140: try to use our `tryCatch` helper where it makes sense (no need to change here!)

packages/build/src/extensions/core/neonSyncEnvVars.ts
```diff
@@ -0,0 +1,289 @@
...
+
+      if (!targetDatabase) {
+        throw new Error(
+          `syncNeonEnvVars: Database '${options?.databaseName}' not found in branch.`
+        );
+      }
+
+      // Step 4: Get the role (user) and password
+      const targetRoleName = options?.roleName ?? targetDatabase.owner_name;
+      const rolePasswordUrl = `https://console.neon.tech/api/v2/projects/${projectId}/branches/${neonBranchId}/roles/${targetRoleName}/reveal_password`;
```
nicktrn, line 234: they don't have a nice js client for this? 👀

0ski: They do have it: https://neon.com/docs/reference/typescript-sdk
Hmm, I was piggybacking on the `syncVercelEnvVars`, hence the direct API calls. Do you think it would be better to add Vercel (and NeonDB) SDK deps? I'm happy to do so. I don't have a strong opinion here, on one side it's more convenient and maintenance is on their SDK, on the other side we would introduce additional deps in build, which might not be needed by majority of our users.

packages/build/src/extensions/core/neonSyncEnvVars.ts
0ski: `import { tryCatch } from "@trigger.dev/core";` - it looks cool, gonna add it

docs/config/extensions/syncEnvVars.mdx
0ski: good idea!

packages/build/src/extensions/core/neonSyncEnvVars.ts
nicktrn: let's not add, was just curious. keep deps minimal

packages/build/src/extensions/core/neonSyncEnvVars.ts
0ski: Ah actually I would have to wrap all the calls separately and rethrow errors. I think it's better to just leave it as it is

packages/build/src/extensions/core/neonSyncEnvVars.ts
0ski: Yeah, I also lean towards not adding these deps now

docs/config/extensions/syncEnvVars.mdx
0ski: Ah, crap I forgot to update, need to create a separate PR

PR #2732, opened 2025-12-03
chore(releases): automatically update the lockfile in changeset PRs

myftija:
This is a step which we currently need to do manually and it's rather painful. The lockfile update is necessary due to cross references in our packages.

Added it as a separate job instead of a step to start from fresh workspace, as the state that the `changeset` step leaves the workdir is not explicitly clear to the reader.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,6 +82,14 @@ jobs:
           package_version=$(echo '${{ steps.changesets.outputs.publishedPackages }}' | jq -r '.[0].version')
           echo "package_version=${package_version}" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -94,3 +94,41 @@ jobs:
...
+        with:
+          version: 10.23.0
+
+      - name: Setup node
+        uses: buildjet/setup-node@v4
+        with:
+          node-version: 20.19.0
+
+      - name: Install and update lockfile
+        run: pnpm install
```
myftija, line 130: this should respect the root level `minimumReleaseAge: 4320` config set in `pnpm-workspace.yaml`

PR #2763, opened 2025-12-07
fix(webapp): remove Spline 3D animation to reduce bundle size by ~47%

beydogan:
I noticed the dashboard takes a while(20-30seconds) to load sometimes and that led me investigate the reason. I found out one of the JS chunks is 2035kb, which is almost half of the whole JS chunks. 

[That chunk](https://cloud.trigger.dev/build/_shared/chunk-TNKKYXTE.js) is three.js library. Which is a dependency for `@splinetool/react-spline`, which is only being used to render a spline scene on the error page. 

As a user, I prefer to have a fast loading dashboard instead of a fancy 3D animation on an error page. Though, same/similar animation can be achieved via lottie or as css animation probably, which will have much smaller size.   

This change basically **reduces JS bundle size from 4283kb to 2248kb**. Even though the assets are being cached I get this slow loading dashboard several times per day. I find that a bit poor UX. 

Further optimization is possible by implementing tree shaking to [companyicons](https://github.com/triggerdotdev/companyicons) to bundle only used icons to the build instead of all the icons. That would probably cut another ~600kb since [webapp is using only 5 icons from there](https://github.com/search?q=repo%3Atriggerdotdev%2Ftrigger.dev+companyicons+path%3A%2F%5Eapps%5C%2Fwebapp%5C%2F%2F&type=code) but loading whole 748kb.

Also noticed the assets are being served without gzip on production, which would improve the situation if you want to keep the animation.

```bash
 > curl -I -H "Accept-Encoding: gzip, br" https://cloud.trigger.dev/build/_shared/chunk-TNKKYXTE.js 2>&1

content-type: application/javascript; charset=UTF-8
```

I'm noticing this issue since currently I'm in Türkiye and have a high latency to US-East. Probably a CDN would help a lot too!

## ✅ Checklist

---

## Testing

Ran locally and went to a 404 page.

---

## Changelog

fix(webapp): remove Spline 3D animation to reduce bundle size

---

### Before

image

### After 

image

### Resources

image

💯

### DIFF
--- a/apps/webapp/app/components/ErrorDisplay.tsx
+++ b/apps/webapp/app/components/ErrorDisplay.tsx
@@ -1,11 +1,9 @@
 import { HomeIcon } from "@heroicons/react/20/solid";
 import { isRouteErrorResponse, useRouteError } from "@remix-run/react";
-import { motion } from "framer-motion";
 import { friendlyErrorDisplay } from "~/utils/httpErrors";
 import { LinkButton } from "./primitives/Buttons";
 import { Header1 } from "./primitives/Headers";
 import { Paragraph } from "./primitives/Paragraph";
-import Spline from "@splinetool/react-spline";
 import { type ReactNode } from "react";
 
 type ErrorDisplayOptions = {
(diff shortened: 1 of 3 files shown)


beydogan:
If you are fine with this change but don't like the empty state of the page, if you provide the logo asset, I can try to do something with lottie or css that looks something similar visually.

samejr:
Hey @beydogan thanks for flagging this - it's a great spot. We have an alternative PR to solve this where Spline is dynamically loaded instead so we retain the animation: https://github.com/triggerdotdev/trigger.dev/pull/2834

beydogan:
@samejr perfect, thank you!

PR #2834, opened 2026-01-05
Fix(webapp): Dynamically load spline on 404 on page

samejr:
Remove the spline dependency and load it dynamically on the 404 page.

Uses a web component with a workaround for React.

### DIFF
--- a/apps/webapp/app/components/TriggerRotatingLogo.tsx
+++ b/apps/webapp/app/components/TriggerRotatingLogo.tsx
@@ -13,6 +13,10 @@ declare global {
       >;
     }
   }
+
+  interface Window {
+    __splineLoader?: Promise<void>;
+  }
 }
 
 export function TriggerRotatingLogo() {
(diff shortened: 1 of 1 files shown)


apps/webapp/app/components/ErrorDisplay.tsx
```diff
@@ -5,8 +5,21 @@ import { friendlyErrorDisplay } from "~/utils/httpErrors";
 import { LinkButton } from "./primitives/Buttons";
 import { Header1 } from "./primitives/Headers";
 import { Paragraph } from "./primitives/Paragraph";
-import Spline from "@splinetool/react-spline";
-import { type ReactNode } from "react";
+import { type ReactNode, useEffect } from "react";
+
```
matt-aitken: Let's pull this out into a separate TriggerRotatingLogo component so it doesn't contaminate this file.

Also pump the issue below into Claude and see what it thinks about the race condition

ericallam (approved):

PR #2841, opened 2026-01-06
fix(batch): optimize processing batch trigger v2

ericallam:
This PR fixes some issues with the new BatchQueue by implementing the full two-phase dequeue process in the FairQueue, and moving the responsibility of consuming the worker queue to the BatchQueue and independently enabling it via the `BATCH_QUEUE_WORKER_QUEUE_ENABLED` env var. We've also introduced the `BATCH_QUEUE_SHARD_COUNT` env var to control the count of master queue shards in the FairQueue. We can also control how many queues are considered in each iteration of the master queue consumer via the `BATCH_QUEUE_MASTER_QUEUE_LIMIT` env var.

This PR will also now skip trying to dequeue from tenants that are at concurrency capacity, which should lead to fewer issues with low concurrency tenants blocking higher concurrency tenants from processing.

### DIFF
--- a/internal-packages/run-engine/src/batch-queue/index.ts
+++ b/internal-packages/run-engine/src/batch-queue/index.ts
@@ -13,6 +13,7 @@ import {
   CallbackFairQueueKeyProducer,
   WorkerQueueManager,
   BatchedSpanManager,
+  isAbortError,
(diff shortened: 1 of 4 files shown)


ericallam:
@claude review this PR please

PR #2860, opened 2026-01-09
fix(helm): support secrets.existingSecret for core secrets

### LINKED ISSUE #2859: bug: Using Kubernetes Secrets instead of hardcoded values is not working

Not applicable

### Describe the bug

According to the [Kubernetes Documentation](https://trigger.dev/docs/self-hosting/kubernetes#custom-values) there are two ways to provide secrets.

1. Hardcoding them like so:
```yaml
secrets:
  enabled: true
  sessionSecret: "your-32-char-hex-secret-1"
  magicLinkSecret: "your-32-char-hex-secret-2"
  # ...
```

2. Using a Kubernetes Secret like so:
```yaml
# Recommended: existingSecret, must contain at least the following keys:
# - SESSION_SECRET
# - MAGIC_LINK_SECRET
# - ENCRYPTION_KEY
# - MANAGED_WORKER_SECRET
# - OBJECT_STORE_ACCESS_KEY_ID
# - OBJECT_STORE_SECRET_ACCESS_KEY
secrets:
  enabled: false
  existingSecret: "your-existing-secret"
```

Sadly the second version is simply not working.
If you set the secret this way they are not recognized leaving you with a bunch of errors like this one (when starting the web app):
```
+ NODE_PATH=/triggerdotdev/node_modules/.pnpm/node_modules exec dumb-init node --max-old-space-size=8192 ./build/server.js
/triggerdotdev/node_modules/.pnpm/zod@3.25.76/node_modules/zod/v3/types.cjs:120
        throw result.error;
        ^
ZodError: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "SESSION_SECRET"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "MAGIC_LINK_SECRET"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "ENCRYPTION_KEY"
    ],
    "message": "Required"
  }
]
```

### Reproduction repo

https://trigger.dev/docs/self-hosting/kubernetes#custom-values

### To reproduce

1. Get yourself the example values.yaml from the docs
2. set `secrets.enable` to `false` to not use the hardcoded secrets
3. Create a secret and reference it in `secrets.existingSecret`
4. Deploy the helm chart using your values file
5. Encounter deployment issues due to seemingly missing secrets

### Additional information

_No response_

nicktrn:
When users set `secrets.enabled=false` to use an external secret via `secrets.existingSecret`, the environment variables `SESSION_SECRET`, `MAGIC_LINK_SECRET`, `ENCRYPTION_KEY`, and `MANAGED_WORKER_SECRET` were not being populated from the secret.

Fixes #2859

Also adds automatic helm prereleases for PRs

### DIFF
--- a/.github/workflows/helm-pr-prerelease.yml
+++ b/.github/workflows/helm-pr-prerelease.yml
@@ -1,0 +1,138 @@
+name: 🧭 Helm Chart PR Prerelease
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - "hosting/k8s/helm/**"
+
+concurrency:
+  group: helm-prerelease-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+env:
+  REGISTRY: ghcr.io
+  CHART_NAME: trigger
+
+jobs:
+  lint-and-test:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: "3.18.3"
+
+      - name: Build dependencies
+        run: helm dependency build ./hosting/k8s/helm/
+
+      - name: Extract dependency charts
+        run: |
+          cd ./hosting/k8s/helm/
+          for file in ./charts/*.tgz; do echo "Extracting $file"; tar -xzf "$file" -C ./charts; done
+
+      - name: Lint Helm Chart
+        run: |
+          helm lint ./hosting/k8s/helm/
+
+      - name: Render templates
+        run: |
+          helm template test-release ./hosting/k8s/helm/ \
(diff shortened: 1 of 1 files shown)


nicktrn:
@claude check if we could easily add helm chart prereleases for prs like this one. i want to avoid having to manually trigger releases. desired outcome: pr that touches helm chart files -> helm prerelease. note that semver app version selectors like ~4.0.0 should not match prereleases

nicktrn:
@claude looks great, can you add that workflow to this pr to test? also, make sure this is based on our existing helm workflow as we know it works. something we'll have to figure out: when we keep pushing to the same branch / pr we shouldn't keep adding comments each time but instead just update the existing comment. don't create a gh release btw, no need

PR #2874, opened 2026-01-13
Fix/sdk stream root fallback

### LINKED ISSUE #2852: bug: getRunIdForOptions function is duplicated, contains inconsistent fallback behavior

System:
    OS: macOS 15.7.3
    CPU: (16) arm64 Apple M4 Max
    Memory: 1.99 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh

  Binaries:
    Node: 22.16.0 - 
    Yarn: 1.22.22 
    npm: 10.9.2 
    pnpm: 10.26.2 
    Watchman: 2026.01.05.00 

### Describe the bug

The `getRunIdForOptions` function is duplicated in two locations with inconsistent fallback behavior. The SDK version (`packages/trigger-sdk/src/v3/streams.ts`) lacks fallback logic that exists in the core implementation (`packages/core/src/v3/realtimeStreams/manager.ts`).

When calling `streams.pipe("key", stream, { target: "root" })` from a root task (a task not triggered by another task), the SDK throws an error instead of falling back to the current run ID. The semantic meaning of target: "root" is "send to the root of this task tree" - if the current task IS the root, it should behave the same as target: "self".

### Reproduction repo

This can be reproduced in any Trigger.dev project using the SDK's streams.pipe() API with target: "root" from a directly-triggered task.

### To reproduce

```typescript
import { streams, task } from "@trigger.dev/sdk";

  // This task is triggered directly (not a child of another task)
  export const rootTask = task({
    id: "root-task",
    run: async (payload) => {
      const myStream = createSomeAsyncGenerator();

      // This will throw an error even though it should work
      const { waitUntilComplete } = streams.pipe("output", myStream, {
        target: "root"  // Fails because rootTaskRunId is undefined for root tasks
      });

      await waitUntilComplete();
    },
  });
```

  **Steps:**
  1. Create a task that is triggered directly (not as a child of another task)
  2. Call streams.pipe() with { target: "root" }
  3. Observe the error is thrown

### Additional information

**Root Cause:** Two `getRunIdForOptions` functions exist with different behavior:

  | File | Fallback for `"root"`/`"parent"`? |
  |------|----------------------------------|
  | `packages/trigger-sdk/src/v3/streams.ts:665-683` | ❌ No |
  | `packages/core/src/v3/realtimeStreams/manager.ts:141-159` | ✅ Yes |

  **Why the core fallback is unreachable:** The SDK resolves targets to string IDs FIRST, then passes them to the core. By the time `StandardRealtimeStreamsManager.pipe()` is called, it receives a string ID (or an error was already thrown), so its fallback logic is dead code for SDK callers.

  **Evidence the fallback is intended:**
  - Documentation (`docs/tasks/streams.mdx:374,399`) shows `target: "root"` as valid usage
  - Internal API (`packages/core/src/v3/runMetadata/manager.ts:330-333`) passes semantic targets directly and relies on the fallback

  **Origin:** Both functions introduced in same commit (`536d9fa21` - "feat(realtime): Realtime streams v2 #2632"). Inconsistency was present from creation.

  **Affected functions:** `streams.pipe()`, `streams.append()`, `streams.writer()` - all use the same `getRunIdForOptions` in `streams.ts`

bharathkumar39293:
This PR fixes the SDK stream helpers (streams.pipe, streams.append, streams.writer) to correctly handle semantic targets root and parent in root tasks. Previously, the getRunIdForOptions function in the SDK lacked fallback logic when the current task was a root task, causing errors. This change aligns SDK behavior with Core semantics, ensuring that the current run ID is used as a fallback.

Testing

Automated:

Added regression tests in packages/trigger-sdk/src/v3/streams.test.ts

Verified that:

streams.pipe(..., { target: "root" }) does not throw in a root task

streams.pipe(..., { target: "parent" }) does not throw in a root task

Tests pass locally

Manual:

Built the SDK successfully using pnpm run build --filter @trigger.dev/sdk

Changelog:
Fixed SDK stream helpers to fallback to current run ID when target: "root" or target: "parent" is used in a root task, aligning SDK behavior with Core semantics.

Changeset included:

Added a patch changeset for @trigger.dev/sdk

Issue Link:
Closes #2852

---

<a href="https://app.devin.ai/review/triggerdotdev/trigger.dev/pull/2874">
  <picture>
    Open with Devin
  </picture>
</a>


bharathkumar39293:
This change aligns SDK fallback behavior with the existing Core implementation.
Longer-term, semantic target resolution could be centralized to avoid duplication, but this PR intentionally keeps scope minimal.

ericallam:
@bharathkumar39293 This looks great but can you remove the change to the throttle function in the webapp or explain the reason for the change?

bharathkumar39293:
Hi @ericallam ,

I’ve removed the unrelated throttle changes from this PR so it now focuses purely on the SDK stream root/parent fallback fix. The throttle/webapp fix for dashboard span refresh is still present in PR #2875, where it belongs.
This should make the review easier and keep the changes scoped correctly.
Thanks!

bharathkumar39293:
@ericallam
Thanks again for the earlier review.
The PR is now fully scoped to the SDK fix only, tests + changeset included, and no further changes are pending from my side.
Appreciate an approval when you get a chance.

PR #2878, opened 2026-01-13
fix(webapp): cancel button in API key regeneration modal

nicktrn:
The Cancel button was missing an onClick handler to close the modal dialog. This caused confusing behavior where clicking Cancel would not dismiss the dialog. Also added type="button" to prevent form submission since the button is inside a form.


0ski (approved):
LGTM!

PR #2885, opened 2026-01-14
feat(webapp): New limits page

samejr:
CleanShot 2026-01-14 at 13 41 02

### DIFF
--- a/apps/webapp/app/presenters/v3/ScheduleListPresenter.server.ts
+++ b/apps/webapp/app/presenters/v3/ScheduleListPresenter.server.ts
@@ -96,7 +96,7 @@ export class ScheduleListPresenter extends BasePresenter {
 
     const schedulesCount = await CheckScheduleService.getUsedSchedulesCount({
       prisma: this._replica,
-      environments: project.environments,
+      projectId,
     });
 
     const limit = await getLimit(project.organizationId, "schedules", 100_000_000);


apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
```diff
@@ -0,0 +1,431 @@
...
+    organizationId,
+    environmentApiKey,
+  }: {
+    userId: string;
+    projectId: string;
+    organizationId: string;
+    environmentApiKey: string;
+  }): Promise<LimitsResult> {
+    // Get organization with all limit-related fields
+    const organization = await this._replica.organization.findUniqueOrThrow({
```
ericallam: use `findFirstOrThrow`

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
```diff
@@ -0,0 +1,431 @@
...
+    // Resolve batch concurrency config
+    const batchConcurrencyConfig = resolveBatchConcurrencyConfig(
+      organization.batchQueueConcurrencyConfig
+    );
+    const batchConcurrencySource = organization.batchQueueConcurrencyConfig
+      ? "override"
+      : "default";
+
+    // Get schedule count for this org
+    const scheduleCount = await this._replica.taskSchedule.count({
```
ericallam: this worries me, really not sure if this will be fast

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
```diff
@@ -0,0 +1,431 @@
...
+            environment: {
+              organizationId,
+            },
+          },
+        },
+      },
+    });
+
+    // Get alert channel count for this org
+    const alertChannelCount = await this._replica.projectAlertChannel.count({
```
ericallam, line 139: same, performance could be an issue

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
```diff
@@ -0,0 +1,431 @@
...
+      where: {
+        project: {
+          organizationId,
+        },
+      },
+    });
+
+    // Get active branches count for this org
+    const activeBranchCount = await this._replica.runtimeEnvironment.count({
```
ericallam, line 146: same, possible perf issue

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
samejr: Done

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
samejr: AI found a more performant way to do this query which i've committed: 

The current query goes through instances -> environment -> organizationId which requires multiple joins with no direct indexes.
Better approach: TaskSchedule has a projectId, and Project has organizationId. Query through project instead:

```
const scheduleCount = await this._replica.taskSchedule.count({  
   where: {   
       project: {    
           organizationId,    
       },
    },
});
```

This is more efficient because:
`TaskSchedule has @@index([projectId])`
Avoids the extra join through `TaskScheduleInstance` and `RuntimeEnvironment`
Direct path: TaskSchedule → Project (by projectId) → filter by organizationId
The current query requires: TaskSchedule → TaskScheduleInstance → RuntimeEnvironment → Organization (3 joins, no helpful indexes on the instance table).

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
samejr: AI response to this: 

This query is actually okay, though not perfectly optimized:
Current state:
- ProjectAlertChannel has `@@unique([projectId, deduplicationKey])` - composite index with `projectId` as leading column ✓
- `Project`has NO @@index([organizationId]) ✗

How the query executes:
Join `ProjectAlertChannel` → `Project` via `projectId` (uses composite unique index) ✓
Filter Project by organizationId (no index, requires scan) ✗

However, this is probably fine in practice because:
- The Project table is small (most orgs have 1-10 projects)
- The join to ProjectAlertChannel uses the index efficiently

If you wanted to optimize further, you'd add an index to the Project model:

```
model Project {
   // ... fields  
   @@index([organizationId])
}
```

This would help all queries that filter projects by organization. But for a low-volume page like limits, the current query is acceptable.

apps/webapp/app/presenters/v3/LimitsPresenter.server.ts
samejr: This one can be improved. I've update it:


RuntimeEnvironment has its own organizationId field AND an index on it (@@index([organizationId])). No need to go through the project relation:

```ts
// Current: goes through project (no index on project.organizationId)
project: { organizationId }

// Better: direct field with index
organizationId  // @@index([organizationId]) exists
```

Changed to:

```ts
const activeBranchCount = await this._replica.runtimeEnvironment.count({
  where: {
    organizationId,
    branchName: {
      not: null,
    },
    archivedAt: null,
  },
});
```

PR #2889, opened 2026-01-15
fix(webapp): logs page

mpcgrid:
## Changelog
* Removed EVENT_REPOSITORY_CLICKHOUSE_ROLLOUT_PERCENT
* Added hasLogsPageAccess featureFlag for logs page
* Replaced attributes with attributes_text for logs to reduce memory usage and improve query performance
* Added support for event_v1 for logs, now depending on the settings the logs are fetched either from  `task_events_v1` or `task_events_v2`
* Show an error in the interface in cast the repository store is `postgres`
Screenshot 2026-01-15 at 12 49 22

## Testing

To change between where the query looks for logs there is the following logic

1. organization.featureFlags (highest priority) 
2. global feature flags (via flags() function)
3. env.EVENT_REPOSITORY_DEFAULT_STORE (fallback)

The feature flags are:
`{"hasLogsPageAccess": true}`
`{"taskEventRepository": "clickhouse_v2"}`

The accepted values are for `taskEventRepository`: `postgres` | `clickhouse` | `clickhouse_v2`
Changing the taskEventRepository value in DB is reflected instantly in the application.


mpcgrid:
@coderabbitai review

PR #2898, opened 2026-01-16
fix(core): harden zodSocket protocol detection against payload field ambiguity

ShivaReddyVanja:
## Problem
Fixed a critical bug where user messages containing a `payload` property were incorrectly identified as protocol-wrapped messages, causing data corruption.
The original implementation only checked for the presence of `payload` to determine if a message was wrapped, leading to false positives when user data legitimately contained a `payload` field.

## Solution
Enhanced protocol detection to require **both** `payload` AND `version` properties, plus validation that `payload` is a plain object (not array/null/primitive).

Changes:
- Enhanced hasValidPayload check to require both 'payload' AND 'version'
- Added isObject() helper to exclude arrays and null values
- Refactored message normalization into dedicated normalizeMessage() method
- Added comprehensive test suite with 11 test cases covering all edge cases

The fix ensures that only messages with the full protocol signature (version + payload as object) are treated as wrapped messages.

## ✅ Checklist

---

## Testing

✅ Added a new test file covering 11 cases
✅ All 11 new tests passing
✅ All 392 existing core tests passing
---

## Changelog
Fixes: Protocol ambiguity where user data with 'payload' field was misinterpreted
Added Tests: 11/11 passing, including edge cases for non-object payloads and missing version

## Screenshots

Screenshot 2026-01-16 at 2 22 36 PM
Screenshot 2026-01-16 at 2 24 49 PM

💯

### DIFF
--- a/packages/core/src/v3/zodSocket.ts
+++ b/packages/core/src/v3/zodSocket.ts
@@ -79,6 +79,12 @@ type MessageFromSocketSchema<
   payload: z.input<GetSocketMessageSchema<TMessageCatalog, K>>;
 };
 
+type NormalizedMessage = {
+  type: string;
+  version?: string;
+  payload: Record<string, any>;
+};
+
 export type MessagesFromSocketCatalog<TMessageCatalog extends ZodSocketMessageCatalogSchema> = {
   [K in keyof TMessageCatalog]: MessageFromSocketSchema<K, TMessageCatalog>;
 }[keyof TMessageCatalog];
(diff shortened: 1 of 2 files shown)


ericallam:
@ShivaReddyVanja appreciate the PR but this is the type of change that we can't really accept. The amount of risk changing something like this is too high, and would require extensive testing

ShivaReddyVanja:
ok,  i completely understand . Is there any other  area that i can help, happy to fix !

PR #2926, opened 2026-01-23
fix(engine): store costInCents and usageDurationMs on the TaskRun table via existing run engine updates

ericallam:
Moving usage updates into the run engine to prevent inefficient &
additional incremental updates to the TaskRun table. Read/Modify/Write
pattern is safe inside of the run engine because of the run lock. We can
also now cap the usageDurationMs value from overflowing and causing an
error.

## Why?

This is preventing at least one update per TaskRun and instead updating these values piggybacking on other updates.

### TL;DR
Aurora readers share the same storage as the writer, but maintain separate in-memory page caches. This means:
- **Storage is always consistent** - writes are synchronously committed to shared storage
- **Page cache can lag** - typically <100ms, but can cause stale reads if data is cached

### How It Works
1. Writer commits to shared storage (synchronous 4/6 quorum)
2. Writer sends cache invalidation messages to readers (asynchronous)
3. If reader has data in cache → returns cached (potentially stale) value
4. If reader has cache miss → fetches from shared storage (always current)

### Monitoring
```sql
SELECT server_id,
       CASE WHEN session_id = 'MASTER_SESSION_ID' THEN 'Writer' ELSE 'Reader' END AS role,
       replica_lag_in_msec
FROM aurora_replica_status();
```

### DIFF
--- a/internal-packages/run-engine/src/engine/systems/runAttemptSystem.ts
+++ b/internal-packages/run-engine/src/engine/systems/runAttemptSystem.ts
@@ -899,7 +899,7 @@ export class RunAttemptSystem {
 
           const failedAt = new Date();
 
-          const retryResult = await retryOutcomeFromCompletion(prisma, {
+          const retryResult = await retryOutcomeFromCompletion(this.$.readOnlyPrisma, {
             runId,
             error: completion.error,
             retryUsingQueue: forceRequeue ?? false,


internal-packages/run-engine/src/engine/retrying.ts
```diff
@@ -37,6 +37,10 @@ export type RetryOutcome =
       settings: TaskRunExecutionRetry;
       machine?: string;
       wasOOMError?: boolean;
+      // Current usage values for calculating updated totals
+      usageDurationMs: number;
```
mpcgrid, line 41: usageDurationMs might overflow for long running tasks (more than 26 days)
There is a cap for this,  just double checking if it also influences the metrics.

internal-packages/run-engine/src/engine/retrying.ts
```diff
@@ -37,6 +37,10 @@ export type RetryOutcome =
...
+      costInCents: number;
```
mpcgrid, line 42: Considering we have a large numbers of tasks the rounding might count as an overcharge or loss.
In this case, do we want the smallest unit of measuring to be cents?

internal-packages/run-engine/src/engine/retrying.ts
ericallam: Yea we handle the overflow now in this PR

internal-packages/run-engine/src/engine/retrying.ts
ericallam: Yea something to investigate

PR #2929, opened 2026-01-23
fix(react-hooks): prevent onComplete from firing prematurely when stream disconnects

### LINKED ISSUE #2856: bug: useRealtimeRun onComplete fires prematurely in self-hosted environment (Traefik + Docker Compose)
#### System:
    OS: macOS 15.7
    CPU: (36) x64 Intel(R) Xeon(R) W-2191B CPU @ 2.30GHz
    Memory: 71.57 GB / 128.00 GB
    Shell: 5.9 - /bin/zsh
#### Binaries:
    Node: 23.6.1 - /Users/<home>/.nvm/versions/node/v23.6.1/bin/node
    npm: 10.9.2 - /Users/<home>/.nvm/versions/node/v23.6.1/bin/npm
    pnpm: 10.19.0 - /Users/<home>/Library/pnpm/pnpm

### Reproduction Environment
- Hosting: Self-hosted using the standard Trigger.dev `docker-compose.yml`.
- Reverse Proxy: Traefik.
- Hook Config: Using `baseURL` pointing to the Traefik entrypoint for the Trigger API.
- SDK Versions: `@trigger.dev/sdk@4.3.2`, `@trigger.dev/react-hooks@4.3.2`.
- Framework: Next.js 16.10.0 / React 19.2.0

### Infrastructure Details
The instance is running behind Traefik using the default Docker Compose setup. While the `onComplete` callback fires early, the `run` state returned by the hook _does_ continue to update correctly in the UI as the task progresses, eventually reaching `COMPLETED`. This suggests the realtime connection itself is working, but the logic governing the callback trigger is misfiring in this network environment.

### Describe the bug

When using the `useRealtimeRun` hook with a self-hosted Trigger.dev instance, the `onComplete` callback fires immediately upon the first successful data fetch, even if the run is still in an active state (e.g., `EXECUTING`).

### Reproduction repo

None

### Reproduction Steps
1. Trigger a task that includes a significant delay (e.g., 5-10 seconds).
2. Use `useRealtimeRun` in a client component with a `baseURL` and an `onComplete` callback.

```ts
const { run } = useRealtimeRun(runId, {
     accessToken: publicToken,
     baseURL: "https://your-self-hosted-trigger.com",
     onComplete: (run) => {
       console.log("onComplete called with status:", run.status);
     }
   });
```

3. Observe that onComplete is called almost immediately after the component mounts.
4. Logging the `run` object inside `onComplete` shows `status: "EXECUTING"`.### Expected Behavior
`onComplete` should only be called once the run reaches a terminal state (`COMPLETED`, `FAILED`, `CANCELED`, `CRASHED`).
### Actual Behavior
`onComplete` is called immediately with the non-terminal state.

### Additional information

I also checked traefik logs but there were no issues.

nicktrn:
## Summary

Fixes #2856 - The `onComplete` callback in `useRealtimeRun` was firing prematurely in self-hosted environments with reverse proxies like Traefik.

## Root Cause

The callback was triggered when the long-poll stream ended, regardless of whether the run had actually completed. Reverse proxies often close idle connections, causing the stream to end prematurely. In this case it was caused by fetch abort due to React strict mode.

## Fix

Changed the condition from checking if `run` exists to checking if `run?.finishedAt` exists, ensuring `onComplete` only fires when the run has reached a terminal state.

---

<a href="https://app.devin.ai/review/triggerdotdev/trigger.dev/pull/2929">
  <picture>
    Open with Devin
  </picture>
</a>


bbreunig:
@nicktrn this PR only fixes the onComplete function. However this would still not return a Server-sent Event since you haven't set `liveSsl: true` as I mentioned here: https://github.com/triggerdotdev/trigger.dev/issues/2856#issuecomment-3733023579
Is there a PR planned for this as well? Or am I missing something obvious?

nicktrn:
> @nicktrn this PR only fixes the onComplete function. However this would still not return a Server-sent Event since you haven't set `liveSsl: true` as I mentioned here: [#2856 (comment)](https://github.com/triggerdotdev/trigger.dev/issues/2856#issuecomment-3733023579) Is there a PR planned for this as well? Or am I missing something obvious?

hey @bbreunig I've commented on the issue with some additional details re the most likely cause: react strict mode.
This causes the long poll to restart (we intentionally don't use sse here) and the previous fetch request to be aborted, which triggers `onComplete`.

I thought about adding a more specific fix here but this seems cleaner and covers more potential edge cases.

PR #2980, opened 2026-01-30
feat(server): New TTL system, enforce max queue length limits, lazy waitpoint creation

ericallam:
This PR implements a new run TTL system and queue size limits to prevent unbounded queue growth which should help prevent situations where queues enter a "death spiral" where the queue will never be able to catch up. 

The main/correct way to battle this situation is to enforce a maximum TTL on all runs (e.g. up to 14 days) where runs that have been queued for that maximum TTL will get auto-expired, making room for newer runs to execute. This required creating a new TTL system that can handle higher workloads and is now deeply integrated into the RunQueue. When runs are enqueued with a TTL, they are added to their normal queue as well as to the TTL queue. When runs are dequeued, they are removed from both their normal queue and the TTL queue. If runs are dequeued by the TTL system, they are removed from their normal queue. Both these dequeues happen automatically so there is no race condition.

The TTL expiration system is also made reliable by expiring runs via a Redis worker, which is enqueued to atomically inside the TTL dequeue lua script.

### Optional associated waitpoints

Additionally, this PR implements an optimization where runs that aren't triggered with a dependent parent run will no longer create an associated waitpoint. Associated waitpoints are then lazily created if a dependent run wants to wait for the child run post-facto (via debounce or idempotency), which is a rare situation but is possible. This means fewer waitpoint creations but also fewer waitpoint completions for runs with no dependencies.

### Environment Queue Limits

Prevents any single queue growing too large by enforcing queue size limits at trigger time.

- Queue size checks happen at trigger time - runs are rejected if queue would exceed limit
- Dashboard UI shows queue limits on both the Queues page and a new Limits page
- In-memory caching for queue size checks to reduce Redis load

### Batch trigger fixes

Currently when a batch item cannot be created for whatever reason (e.g. queue limits) the run will never get created, which means a stalled run if using `batchTriggerAndWait`. We've updated the system to handle this differently: now when a batch item cannot be triggered and converted into a run, we will eventually (after retrying 8 times up to 30s) we will create a "pre-failed" run with the error details, correctly resolving the batchTriggerAndWait.


apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.env.$envParam.queues/route.tsx
```diff
@@ -509,7 +512,10 @@ export default function Page() {
                   {queues.length > 0 ? (
                     queues.map((queue) => {
                       const limit = queue.concurrencyLimit ?? environment.concurrencyLimit;
-                      const isAtLimit = queue.running >= limit;
+                      const isAtConcurrencyLimit = queue.running >= limit;
+                      const isAtQueueLimit =
+                        environment.queueSizeLimit !== null &&
+                        queue.queued >= environment.queueSizeLimit;
```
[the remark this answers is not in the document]
ericallam: The limit is on the environment but it does apply to the individual queues

apps/webapp/app/runEngine/concerns/queues.server.ts
```diff
@@ -208,14 +221,126 @@ export class DefaultQueueManager implements QueueManager {
     return task.queue.name ?? defaultQueueName;
   }
 
+  /**
+   * Resolves queue names for batch items and groups them by queue.
+   * Returns a map of queue name -> count of items going to that queue.
+   */
+  async resolveQueueNamesForBatchItems(
```
ericallam: We might not need this anymore

apps/webapp/app/runEngine/concerns/queues.server.ts
```diff
@@ -208,14 +221,126 @@ export class DefaultQueueManager implements QueueManager {
...
+    }
+
+    return queueCounts;
+  }
+
+  /**
+   * Validates queue limits for multiple queues at once.
+   * Returns the first queue that exceeds limits, or null if all are within limits.
+   */
+  async validateMultipleQueueLimits(
```
ericallam: same thing, might not be needed

apps/webapp/app/runEngine/services/batchTrigger.server.ts
```diff
@@ -264,6 +264,16 @@ export class RunEngineBatchTriggerService extends WithRunEngine {
 
           return batch;
         }
+        case "ABORTED": {
```
ericallam: Clean this up, no longer doing it this way in the new streaming batch stuff

packages/core/src/v3/schemas/api.ts
```diff
@@ -409,6 +409,8 @@ export const StreamBatchItemsResponse = z.object({
   itemsAccepted: z.number(),
   /** Number of items that were deduplicated (already enqueued) */
   itemsDeduplicated: z.number(),
+  /** Number of items skipped due to queue size limits */
+  itemsSkipped: z.number().optional(),
```
ericallam: this needs removed also

packages/trigger-sdk/src/v3/shared.ts
```diff
@@ -1571,10 +1571,15 @@ async function executeBatchTwoPhase(
...
 
   // If the batch was cached (idempotent replay), skip streaming items
+  let actualRunCount = batch.runCount;
   if (!batch.isCached) {
     try {
       // Phase 2: Stream items
-      await apiClient.streamBatchItems(batch.id, items, requestOptions);
+      const streamResult = await apiClient.streamBatchItems(batch.id, items, requestOptions);
+      // Use the runCount from Phase 2 if provided (may differ if items were skipped)
+      if (streamResult.runCount !== undefined) {
```
ericallam: this needs to go

queue-metrics-design.md
```diff
@@ -0,0 +1,1448 @@
+# Queue Metrics System Design
```
ericallam: remove this file

PR #2998, opened 2026-02-04
fix(webapp): synchronize run page highlight when hovering spans

mpcgrid:
Closes #<issue>

## ✅ Checklist

---

## Testing

Checked in the interface 

---

## Changelog

Synchronized row between spans and timeline  on hover

---

## Screenshots

Screenshot 2026-02-04 at 17 36 09

---

<a href="https://app.devin.ai/review/triggerdotdev/trigger.dev/pull/2998" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


nicktrn (commented):
## PR Review: Synchronize run page highlight on hover

**Overall**: The feature goal is good — syncing hover highlights between the tree panel and timeline panel is a clear UX improvement. However, the current implementation has a high-severity bug, significant performance concerns, and several code quality issues that should be addressed before merging.

---

### HIGH — Stale hover after mousewheel scroll

When scrolling via mousewheel while the cursor hovers over a row, the rows move *under* the mouse but `onMouseEnter`/`onMouseLeave` events **do not fire** for the new rows appearing beneath the cursor. The `pendingHoverIdRef` retains the ID of the row that was under the mouse when scrolling *started*. When the 150ms debounce expires, that stale ID is applied — highlighting a row the user isn't pointing at.

Similarly, when virtualized rows are recycled during scroll (TanStack Virtual unmounts/mounts row components), no `onMouseLeave` fires for the unmounted row, leaving `pendingHoverIdRef` stale.

**Suggestion**: When scroll ends, set `hoveredNodeId` to `null` instead of applying the ref value. Let the next natural mouse event re-establish hover:

```js
scrollTimeoutRef.current = setTimeout(() => {
  setIsScrolling(false);
  setHoveredNodeId(null);
  pendingHoverIdRef.current = null;
}, 150);
```

---

### HIGH — Performance: every hover re-renders the entire tree + timeline

`hoveredNodeId` is React state in `TasksTreeView`. Every `setHoveredNodeId` call re-renders the parent, which re-renders **both** `TreeView` components (tree + timeline). Key factors:

- `TreeView` is **not** wrapped in `React.memo`
- `renderNode` callbacks are **inline arrow functions**, recreated every render
- Virtualizer uses `overscan: 50`, so ~70-80 rows re-render per panel per hover event (~140-160 total)
- Timeline rows are expensive (multiple `Timeline.Point`, `motion.div`, Framer Motion animations)

**Suggestion**: Consider a CSS/DOM-based approach that avoids React re-renders entirely. Add `data-node-id` attributes to rows and toggle a `data-hovered` attribute via refs + direct DOM manipulation. The hover highlight is purely visual — it doesn't need to participate in React's render cycle. If that feels too imperative, an intermediate option is to add `pointer-events: none` to the scroll container during scroll (one state toggle on the container, not per-node), and keep CSS `:hover` for the actual highlight.

---

### MEDIUM — CSS hover replaced with JS hover adds latency

The PR replaces native CSS `:hover` pseudo-classes (`hover:bg-grid-dimmed`) with React state-driven styling. CSS `:hover` is handled on the compositor thread and is effectively instant. The JS approach adds at minimum one frame of latency (~16ms) through the React event → setState → reconciliation → paint pipeline, and more under load. Users will notice the highlight trailing the mouse when scanning quickly across rows.

---

### LOW — `setHoveredNodeId` prop passed but unused in TimelineView

`TimelineViewProps` includes `setHoveredNodeId`, but `TimelineView` only uses `handleHoverChange` (which wraps `setHoveredNodeId` with scroll-aware logic). The raw setter shouldn't be exposed — it bypasses scroll suppression and is dead code.

---

### LOW — `handleHoverChange` dependency on `isScrolling` state destabilizes callback identity

```js
const handleHoverChange = useCallback((nodeId) => {
  pendingHoverIdRef.current = nodeId;
  if (!isScrolling) {
    setHoveredNodeId(nodeId);
  }
}, [isScrolling]); // changes every time isScrolling toggles
```

When `isScrolling` toggles, `handleHoverChange` gets a new identity. This would bust any `React.memo` on child components receiving it. If `isScrolling` were tracked as a ref instead of state, the callback would be stable (`[]` deps).

---

### LOW — Duplicated hover styling logic

The hover class logic is copy-pasted in both the tree `renderNode` and timeline `renderNode`:

```js
state.selected
  ? isHovered ? "bg-grid-bright" : "bg-grid-dimmed"
  : isHovered ? "bg-grid-dimmed" : "bg-transparent"
```

Could be extracted into a helper like `getRowBgClass(selected, hovered)`.

---

### LOW — Indentation bug in `handleScroll`

```js
scrollTimeoutRef.current = setTimeout(() => {
  setIsScrolling(false);
    setHoveredNodeId(pendingHoverIdRef.current); // 8 spaces, should be 6
}, 150);
```

---

### LOW — Unused parameter

`handleScroll` takes `scrollTop: number` but never reads it. Should be `(_scrollTop: number) => void` or `() => void`.

---

### LOW — `NodeJS.Timeout` type in browser context

`useRef<NodeJS.Timeout | null>` — in a browser environment this should be `ReturnType<typeof setTimeout>` for correctness.

apps/webapp/app/routes/_app.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.$runParam/route.tsx
```diff
@@ -694,6 +694,20 @@ function TasksTreeView({
...
   const [showQueueTime, setShowQueueTime] = useState(false);
   const [scale, setScale] = useState(0);
+  const [hoveredNodeId, setHoveredNodeId] = useState<string | null>(null);
+  const [isScrolling, setIsScrolling] = useState(false);
+  const scrollTimeoutRef = useRef<NodeJS.Timeout | null>(null);
+  const pendingHoverIdRef = useRef<string | null>(null);
+
+  useEffect(() => {
+    return () => {
+      if (scrollTimeoutRef.current) {
```
matt-aitken, line 704: I am pretty sure this is going to cause a re-render of the entire tree every time a hover changes. Which will crush the rendering performance in the browser for complex trees.

We have to very careful to avoid the inputs to the tree view changing rapidly.

PR #3049, opened 2026-02-13
fix: handle invalid source map columns in Bun runtime

### LINKED ISSUE #3045: Bun runtime: source-map crash during indexing (Column must be >= 0, got -1)
## Description

When using `runtime: 'bun'` in trigger.config.ts, the indexing phase crashes with:

```
TypeError: Column must be greater than or equal to 0, got -1
    at SourceMapConsumer_findMapping (source-map/lib/source-map-consumer.js:588:13)
    at SourceMapConsumer_originalPositionFor (source-map/lib/source-map-consumer.js:653:17)
    at mapSourcePosition (source-map-support/source-map-support.js:244:38)
    at wrapCallSite (source-map-support/source-map-support.js:397:9)
    at prepareStackTrace (source-map-support/source-map-support.js:446:25)
    at registerResources (trigger.dev/src/indexing/registerResources.ts:32:11)
    at bootstrap (trigger.dev/src/entryPoints/managed-index-worker.ts:91:37)
```

This happens because Bun's runtime produces `-1` as a column value when resolving stack traces, and `source-map@0.6.1` (transitive dep via `source-map-support@0.5.21`) throws on negative columns.

This is a known Bun issue: https://github.com/oven-sh/bun/issues/8087

## Reproduction

- trigger.dev v4.3.3
- `runtime: 'bun'` in trigger.config.ts
- A monorepo project with shared modules imported across packages

The error started appearing after a refactoring that introduced new cross-package imports, likely producing source maps that trigger the Bun bug.

## Workaround

We patched `source-map@0.6.1` via pnpm to clamp `-1` columns to `0` instead of throwing:

```diff
-      throw new TypeError('Column must be greater than or equal to 0, got '
-                          + aNeedle[aColumnName]);
+      aNeedle[aColumnName] = 0;
```

## Suggested fix

Since bun runtime is already marked as experimental, trigger.dev could either:
1. Catch and gracefully handle this error in `registerResources` during indexing
2. Pin or patch `source-map-support` to handle invalid column values
3. Document this as a known bun runtime limitation

capaj:
## Summary

- Wraps `source-map-support`'s `prepareStackTrace` hook to catch errors from invalid source maps instead of crashing
- Extracted into a shared `installSourceMapSupport` utility used by all 4 entry points
- When `source-map` processing fails (e.g. column value of `-1`), falls back to default stack trace formatting

## Context

Bun's runtime can produce source maps with column values of `-1`, which causes `source-map@0.6.1` (transitive dep via `source-map-support@0.5.21`) to throw during the indexing phase:

```
TypeError: Column must be greater than or equal to 0, got -1
    at SourceMapConsumer_findMapping (source-map/lib/source-map-consumer.js:588:13)
    at registerResources (trigger.dev/src/indexing/registerResources.ts:32:11)
```

This is a known Bun issue: https://github.com/oven-sh/bun/issues/8087

Fixes #3045

## Test plan

- [ ] Deploy a project with `runtime: 'bun'` that previously failed with the column error
- [ ] Verify indexing completes successfully
- [ ] Verify stack traces still work (just with column 0 fallback when source map is invalid)

### DIFF
--- a/packages/cli-v3/src/entryPoints/dev-index-worker.ts
+++ b/packages/cli-v3/src/entryPoints/dev-index-worker.ts
@@ -13,18 +13,14 @@ import {
 } from "@trigger.dev/core/v3/workers";
 import { sendMessageInCatalog, ZodSchemaParsedError } from "@trigger.dev/core/v3/zodMessageHandler";
 import { readFile } from "node:fs/promises";
-import sourceMapSupport from "source-map-support";
 import { registerResources } from "../indexing/registerResources.js";
+import { installSourceMapSupport } from "../utilities/installSourceMapSupport.js";
 import { env } from "std-env";
 import { normalizeImportPath } from "../utilities/normalizeImportPath.js";
 import { detectRuntimeVersion } from "@trigger.dev/core/v3/build";
 import { schemaToJsonSchema } from "@trigger.dev/schema-to-json";
 
-sourceMapSupport.install({
-  handleUncaughtExceptions: false,
-  environment: "node",
-  hookRequire: false,
-});
+installSourceMapSupport();
 
 process.on("uncaughtException", function (error, origin) {
   if (error instanceof Error) {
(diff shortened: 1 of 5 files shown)


ericallam:
@claude isn't there an issue in this repo that is related to installing source maps and causing OOM errors, how would this work effect that? I think maybe it had to do with indexing issues.

capaj:
@claude I think you failed to get back

PR #3084, opened 2026-02-18
Fix for stale data when hovering

matt-aitken:
- Ensure hover displays up-to-date information rather than cached/stale values


matt-aitken:
Going to combine this into another PR with Query fixes

PR #3112, opened 2026-02-23
fix: ensure .env.local and .env.development.local override .env in CLI

### LINKED ISSUE #2999: bug: .env.local doesn't override .env values in dev command

v4.3.2

### Describe the bug

When running trigger dev, .env.local should override values from .env, but it doesn't. The .env values are used instead

### Reproduction repo

n/a

### To reproduce

To reproduce:

1. Create .env with TEST_VAR=from-env
2. Create .env.local with TEST_VAR=from-local
3. Run npx trigger.dev dev
4. In your task, check process.env.TEST_VAR
5. It shows from-env instead of from-local

Expected: `.env.local` should override `.env`, so TEST_VAR should be `from-local`.

### Additional information

The issue is most probably in `packages/cli-v3/src/utilities/dotEnv.ts`. The code passes an array to `dotenv.config({ path: envFilePath })`, but `dotenv.config()` only accepts a string path, not an array. This means only the first file (`.env`) is loaded.

Potential fix:
Loop through each file in ENVVAR_FILES and call dotenv.config() for each one in order, so later files override earlier ones

deepshekhardas:
This PR fixes the environment variable loading order in the CLI to follow standard precedence conventions.

Previously, .env was being loaded in a way that might not allow local overrides to consistently take precedence depending on how dotenv.config handles duplicates in the path array. By moving .env to the end of the ENVVAR_FILES list, we ensure that local and environment-specific files are preferred.

Fixes #2999


ArronKing:
Having this issue also.

I think this PR just needs a slight change, it's not the order of the .env files. Line 20:

`dotenv.config`

It can be given override: true, by default that is set to false. When false, the first file wins.

https://www.npmjs.com/package/dotenv

PR #3232, opened 2026-03-17
fix(engine): lockless waitpoint insert for batch items to eliminate lock contention

ericallam:
When processing batchTriggerAndWait items, each batch item was acquiring a
Redis lock on the parent run to insert a TaskRunWaitpoint row. With high
concurrency (processingConcurrency=50), this caused LockAcquisitionTimeoutError
(880 errors/24h in prod), orphaned runs, and stuck parent runs.

Since blockRunWithCreatedBatch already transitions the parent to
EXECUTING_WITH_WAITPOINTS before items are processed, the per-item lock is
unnecessary. The new blockRunWithWaitpointLockless method performs only the
idempotent CTE insert and timeout scheduling without acquiring the lock.

### DIFF
--- a/.server-changes/fix-batch-waitpoint-lock-contention.md
+++ b/.server-changes/fix-batch-waitpoint-lock-contention.md
@@ -1,0 +1,6 @@
+---
+area: webapp
+type: fix
+---
+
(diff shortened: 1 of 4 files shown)


ericallam:
Needs approval

nicktrn (approved):

PR #3239, opened 2026-03-19
fix: auto-detect Bun runtime when not explicitly set in trigger.config.ts

### LINKED ISSUE #3105: bug: Deployment UI shows Node.js runtime when using Bun

n/a

### Describe the bug

The deployment UI shows "Node.js" as the runtime even when the project is built and run with Bun.

- Build works with a pre-build command like: `bun run build --filter="./packages/*"`
- Logs show Bun in use (e.g. `bun run ...`, Bun version)
- UI still displays "Node.js" as the runtime

Expected: UI should show "Bun" (or the actual runtime) when Bun is used. This is a display bug only; behavior is correct.

### Reproduction repo

n/a

### To reproduce

1. Use Bun for build/run (e.g. add pre-build: `bun run build --filter="./packages/*"`).
2. Deploy successfully.
3. Open the deployment in the dashboard.
4. See "Node.js" shown as runtime instead of Bun.

### Additional information

Purely a UI/label issue

abhay0132:
Fixes #3105

## Problem
When a user runs their project with Bun (e.g. via a pre-build command like 
`bun run build`) but does not explicitly set `runtime: "bun"` in their 
trigger.config.ts, the deployment UI incorrectly shows "Node.js" as the runtime.

This happened because the default runtime was hardcoded to "node" regardless 
of what runtime was actually executing the process.

## Fix
Auto-detect the runtime by checking `process.versions.bun` at config resolution 
time. If Bun is detected and the user has not explicitly set a runtime in their 
config, we now correctly default to "bun" instead of "node".

The explicit user config always takes priority — this only affects the fallback.

## Changes
- `packages/cli-v3/src/config.ts`: replaced hardcoded `"node"` default with 
  runtime auto-detection using `process.versions.bun`

### DIFF
--- a/packages/cli-v3/src/config.ts
+++ b/packages/cli-v3/src/config.ts
@@ -191,7 +191,8 @@ async function resolveConfig(
     ["run_engine_v2" as const].concat(config.compatibilityFlags ?? [])
   );
 
-  const defaultRuntime: BuildRuntime = features.run_engine_v2 ? "node" : DEFAULT_RUNTIME;
+  const detectedRuntime: BuildRuntime = typeof process.versions.bun === "string" ? "bun" : "node";
+  const defaultRuntime: BuildRuntime = config.runtime ?? detectedRuntime;
 
   const mergedConfig = defu(
     {


abhay0132:
Closing this, misread the bug as a config issue when it's display-only. Will open a corrected PR targeting the dashboard rendering layer.

Issue #3259, opened 2026-03-24
self-hosted managed indexer treats env var lookup failure as fatal instead of degrading gracefully


flowq-C:
## Summary
The build-stage indexer aborts the entire deployment if environment variables cannot be fetched. This creates a full deployment outage from a secondary metadata/indexing dependency.

## Proven behavior
- Indexer failure prevents image build completion
- No fallback path
- No warning-only mode

## Why this matters
- This turns a metadata/bootstrap problem into a full deployment outage
- In self-hosted environments this is especially brittle because auth/storage/config drift is more likely than in cloud

## Repro
1. Cause envvar fetch to fail during local build indexing
2. Run deployment
3. Observe full build failure

## Expected
- Indexer should continue with:
  - `env = {}`
- or mark metadata incomplete but still allow deployment image build

## Actual
- Full Docker build fails
- Deployment status ends as `FAILED`

## Evidence
- `managed-index-controller` throws immediately on failed envvar fetch:
  - `throw new Error(\`Failed to fetch environment variables: ${$env.error}\`)`

## Suggested fix
- Downgrade envvar fetch failure to warning
- Continue indexing with empty env variables
- Only fail hard if task graph itself cannot be resolved

flowq-C:
Additional confirmation from production-like self-hosted validation:

We were able to get the deployment through, but only after working around the build-stage API/auth path.

That reinforced the main point of this issue:
- the indexer is currently too eager to make secondary metadata/bootstrap calls fatal
- in self-hosted setups, auth, routing, object storage, and local registry topology are all more fragile than in cloud
- turning those failures into full deployment outages makes the platform operationally brittle

Even after the broader deployment path was repaired, the original recommendation still stands:
- env/bootstrap/indexing failures should degrade gracefully where possible
- the deployable image should not be blocked by non-essential metadata steps

flowq-C:
## Confirmed and extended — root cause of both build-stage failures documented in #3258

Full working fix with reproduction, root cause, and patch code posted on the related issue:

→ **#3258** covers both the `/envvars` 401 and the `/background-workers` connectivity failure with complete working code

The summary of what we found:

1. **`/envvars`** → runtime key gets 401 → indexer aborts → graceful degradation (empty vars, log warning) is the right upstream fix
[...]

Issue #3258, opened 2026-03-24
self-hosted build indexer cannot fetch project env vars with runtime environment key and aborts local deployment build


flowq-C:
## Summary
In self-hosted Trigger.dev local deployments, the build image indexer (`managed-index-controller`) fetches project environment variables during the Docker build stage. In our setup this call fails and aborts the entire deployment build.

## Proven behavior
- Docker build fails inside:
  - `managed-index-controller.mjs`
- Error:
  - `Failed to fetch environment variables: 403 Forbidden`
- Direct API probe against:
  - `GET /api/v1/projects/:projectRef/envvars`
- Response with production runtime key (`tr_prod_...`):
  - `401 Invalid or Missing API key`

## Why this is a Trigger.dev self-hosted bug
- The build pipeline passes `TRIGGER_SECRET_KEY` into the image build.
- The indexer depends on fetching env vars during the build.
- The supplied key path does not authenticate successfully against the envvars endpoint in self-hosted mode.
- Result: deployment image build hard-fails before publish.

## Repro
1. Self-host Trigger.dev v4 with local builds enabled
2. Run `trigger.dev deploy`
3. Let build reach the `managed-index-controller` step
4. Observe failure on env var fetch

## Expected
- Build-stage env var fetch succeeds with the provided deployment/build auth
- or the indexer degrades gracefully and continues with an empty env set

## Actual
- Build aborts with:
  - `Failed to fetch environment variables: 403 Forbidden`

## Evidence
- Build log excerpt:
  - `Failed to index deployment`
  - `message: 'Failed to fetch environment variables: 403 Forbidden'`
- Direct endpoint probe:
  - `GET /api/v1/projects/proj_ncohokyumnepswndlhei/envvars`
  - with `tr_prod_...`
  - returns `401 Invalid or Missing API key`

## Suggested fix
- Use the correct auth token type for build-time envvar resolution
- or make `managed-index-controller` non-fatal when envvar fetch fails

flowq-C:
Validated follow-up from the same live self-hosted environment:

The envvar/auth failure was real, but there was an additional important detail:
- once we routed build-time API traffic directly to the internal webapp endpoint in the host network, background worker creation succeeded during indexing
- before that, the build-stage indexer path was returning `403 Forbidden` on the background-worker creation call as well

So the observed self-hosted behavior was:
- envvar lookup path was not reliable with the deployment/build auth path
[...]

Issue #3256, opened 2026-03-24
self-hosted v4.4.3 native deployments can stay stuck on PENDING because build enqueue is skipped


flowq-C:
## Summary

On self-hosted Trigger.dev v4.4.3, native deployments can get stuck on `PENDING` and later time out because the build is never actually enqueued.

## Environment

- Trigger.dev self-hosted
- Version: `v4.4.3`
- API origin: custom self-hosted origin, not `cloud.trigger.dev`
- Native build / self-hosted deployment flow

## Expected behavior

Creating a native deployment should move it from `PENDING` to `INSTALLING` / `BUILDING` automatically.

## Actual behavior

The deployment row is created, but the build is never enqueued. The deployment stays on `PENDING` and later times out.

## Reproduction

1. Run self-hosted Trigger.dev v4.4.3
2. Create a native deployment
3. Artifact upload succeeds
4. Deployment remains `PENDING`
5. Eventually deployment becomes `TIMED_OUT`

## What we found

In the webapp bundle, the path that should enqueue the build depends on a client initialized by `initializeClient()`.

That client is only initialized when `isCloud()` is true and billing env vars are present.

In self-hosted with a custom origin, `isCloud()` is false, so the client is never initialized.

Then `enqueueBuild()` effectively becomes a no-op because it immediately returns when `client` is missing.

## Evidence

Relevant bundle logic in our running container:

- `initializeClient()` only initializes when `isCloud()` and billing env vars are set
- `enqueueBuild()` returns early when `client` is falsy
- deployment creation path still calls `deploymentService.enqueueBuild(...)`

Observed DB state for affected deployments:

- status: `PENDING`
- no `startedAt`
- no `builtAt`
- no `deployedAt`
- no worker attached

## Impact

Self-hosted native deploys are not reliable without patching the bundle or bypassing this code path.

## Question

Is this an intended limitation of self-hosted native builds in v4.4.3, or is this a bug in the self-hosted code path?

matt-aitken:
Native (Git-based) deployments depend on a build service that is only available on Trigger.dev Cloud. Self-hosted instances must deploy with the CLI (`npx trigger.dev@latest deploy`) or the GitHub Action. We will hide this option in self-hosted instances so that deployments cannot enter the stuck `PENDING` state. Closing this issue in favour of that change. Follow-up: https://github.com/triggerdotdev/trigger.dev/issues/4835

PR #3260, opened 2026-03-24
fix(cli): push localhost registry images in self-hosted local builds

### LINKED ISSUE #3257: self-hosted local registry deployments can succeed without pushing the image, leading to No such image at runtime
## Summary

On self-hosted local builds targeting a local registry like `localhost:5000`, deployments can be marked as successful while the runner later fails with `No such image`, because the built image was never pushed to the registry.

## Environment

- Trigger.dev self-hosted
- Version: `v4.4.3`
- Local build flow
- Local registry image tags like `localhost:5000/...`

## Expected behavior

If self-hosted local builds use a local registry image tag such as `localhost:5000/...`, the built image should still be pushed so supervisors/runners can pull it.

## Actual behavior

The deployment can complete successfully, but the supervisor later fails to pull the image:

`No such image: localhost:5000/...`

In our case, older image tags existed, but the new deployment tag did not exist after a successful deploy.

## Reproduction

1. Use self-hosted Trigger.dev v4.4.3 with a local registry tag like `localhost:5000/...`
2. Perform a local deploy
3. Deployment reports success
4. Trigger a run
5. Supervisor attempts to pull the deployment image and fails with `404 / no such image`

## What we found

In the local build code path, `shouldPush()` returns `false` by default for image tags starting with:

- `localhost`
- `127.0.0.1`
- `0.0.0.0`

That seems wrong for self-hosted local-registry setups, because the built image then never reaches the registry that supervisors/runners are expected to use.

## Evidence

Relevant logic in `buildImage.ts`:

```ts
case undefined: {
  return imageTag.startsWith("localhost") ||
    imageTag.startsWith("127.0.0.1") ||
    imageTag.startsWith("0.0.0.0")
    ? false
    : true;
}
```

Observed runtime failure:

- deployment marked successful
- supervisor log: `No such image: localhost:5000/...`
- previous tags existed, new tag for the new deployment did not

## Impact

Self-hosted local-registry deployments can appear successful but fail at runtime when the runner tries to pull the image.

## Question

Should local-registry targets in self-hosted mode default to `push=true`, or is there another supported path that supervisors are expected to use here?

flowq-C:
## Summary
Self-hosted Trigger.dev installations commonly use a local registry such as `localhost:5000`.

The current `shouldPush()` behavior treats localhost-style image tags as non-push by default. In practice this can produce a deployment that appears complete while the supervisor later fails with `No such image`, because the image was never pushed to the local registry.

## What changed
- changed `shouldPush()` so local builds do not silently skip pushing localhost-tagged images

## Why
- self-hosted local registry setups still require a push step
- skipping the push creates a broken deployment/runtime split

## Real-world failure mode
- deployment version is created
- runtime later tries to start the container
- supervisor fails with:
  - `No such image: localhost:5000/...`

## Notes
- this is specifically important for self-hosted local-registry setups
- it avoids a misleading deployment success state when the runtime image is unavailable

Closes #3257

### DIFF
--- a/packages/cli-v3/src/deploy/buildImage.ts
+++ b/packages/cli-v3/src/deploy/buildImage.ts
@@ -1076,7 +1076,9 @@ const BuildKitMetadata = z.object({
   "image.name": z.string().optional(),
 });
 
-// Don't push if the image tag is a local address, unless the user explicitly wants to push
+// Self-hosted installations commonly use a local registry such as localhost:5000.
+// In that case we still need to push, otherwise the deployment can complete while the
+// supervisor later fails with "No such image".
 function shouldPush(imageTag: string, push?: boolean) {
   switch (push) {
     case true: {
@@ -1086,11 +1088,7 @@ function shouldPush(imageTag: string, push?: boolean) {
       return false;
     }
     case undefined: {
-      return imageTag.startsWith("localhost") ||
-        imageTag.startsWith("127.0.0.1") ||
-        imageTag.startsWith("0.0.0.0")
-        ? false
-        : true;
+      return true;
     }
     default: {
       assertExhaustive(push);


flowq-C:
Thanks for the quick signal.

We opened this PR while debugging multiple reproducible self-hosted deployment failures in a real Trigger.dev self-hosted setup.

Relevant linked issues:
- #3256
- #3257
- #3258
- #3259

This PR is meant as a minimal proposed fix, not a drive-by feature change.

If external PRs require a vouch first, the practical ask is:
- either vouch the contributor so we can iterate normally
- or cherry-pick / re-apply the patch internally if you agree with the fix direction

We are happy to adjust the patch, split it further, or add a changeset if needed once there is a path for contribution.

flowq-C:
Follow-up from live validation on our self-hosted stack:

We have since validated the broader deployment path end-to-end against real self-hosted deployments.

This PR is still directionally correct, but the full local-registry problem turned out to have 3 layers:
- localhost-style tags defaulting to `push=false`
- buildkit not reliably reaching a local registry via `localhost`
- insecure HTTP/auth handling when the effective registry endpoint differs from the original tag

So I would still consider this patch useful, but not sufficient by itself for every self-hosted local-registry topology.

For our stack, we only reached stable local deploys after fixing the full chain.

PR #3261, opened 2026-03-24
fix(webapp): progress native self-hosted deployments instead of dead enqueue path

### LINKED ISSUE #3256: self-hosted v4.4.3 native deployments can stay stuck on PENDING because build enqueue is skipped

flowq-C:
## Summary
In self-hosted native deployments, deployments could remain stuck on `PENDING` or later `TIMED_OUT` because the expected build transition was not happening through the current enqueue path.

In our self-hosted reproduction, progressing the deployment was the action that actually moved the deployment forward and allowed the build/runtime flow to continue.

## What changed
- replace the native self-hosted build transition call from `enqueueBuild(...)` to `progressDeployment(...)`

## Why
- native self-hosted deployments already have the necessary deployment metadata
- the current enqueue behavior can leave deployments stuck without a real build transition

## Real-world failure mode
- deployment initializes successfully
- deployment remains `PENDING`
- eventually deployment becomes `TIMED_OUT`

## Notes
- this change is intentionally narrow
- it only targets the native self-hosted build transition path

Closes #3256

### DIFF
--- a/apps/webapp/app/v3/services/initializeDeployment.server.ts
+++ b/apps/webapp/app/v3/services/initializeDeployment.server.ts
@@ -249,12 +249,16 @@ export class InitializeDeploymentService extends BaseService {
         new Date(Date.now() + timeoutMs)
       );
 
-      // For github integration there is no artifactKey, hence we skip it here
+      // For self-hosted native builds, progressing the deployment is the action that actually
+      // transitions the deployment out of PENDING and starts the local build path.
+      // For github integration there is no artifactKey, hence we skip it here.
       if (payload.isNativeBuild && payload.artifactKey && !payload.skipEnqueue) {
         const result = await deploymentService
-          .enqueueBuild(environment, deployment, payload.artifactKey, {
-            skipPromotion: payload.skipPromotion,
-            configFilePath: payload.configFilePath,
+          .progressDeployment(environment, deployment.friendlyId, {
+            contentHash: payload.contentHash,
+            git: payload.gitMeta,
+            runtime: payload.runtime,
+            buildServerMetadata,
           })
           .orElse((error) => {
             logger.error("Failed to enqueue build", {


flowq-C:
Thanks for the quick signal.

We opened this PR while debugging multiple reproducible self-hosted deployment failures in a real Trigger.dev self-hosted setup.

Relevant linked issues:
- #3256
- #3257
- #3258
- #3259

This PR is meant as a minimal proposed fix, not a drive-by feature change.

If external PRs require a vouch first, the practical ask is:
- either vouch the contributor so we can iterate normally
- or cherry-pick / re-apply the patch internally if you agree with the fix direction

We are happy to adjust the patch, split it further, or add a changeset if needed once there is a path for contribution.

flowq-C:
Follow-up from live validation:

We continued debugging this against the same self-hosted installation and got the deploy path working end-to-end.

This PR still matches one of the real failure modes we saw early in the process:
- self-hosted native deployments could sit in `PENDING` / later `TIMED_OUT`

After repairing the full path, we successfully deployed newer versions and created background workers in the real self-hosted environment.

So even though this PR was auto-closed by contribution policy, the underlying failure mode was real in our repro and not just a local misread of deployment state.

PR #3303, opened 2026-03-31
fix: add build step to @internal/compute package

nicktrn:
The @internal/compute package had its main/types pointing to ./src/index.ts with no build step. This works in dev (tsc resolves .ts at compile time) but fails at runtime in Docker because Node.js can't load .ts files directly.

Added tsconfig.build.json and build/clean/dev scripts matching the pattern used by schedule-engine and other internal packages. Exports now point to dist/.

### DIFF
--- a/internal-packages/compute/package.json
+++ b/internal-packages/compute/package.json
@@ -2,13 +2,27 @@
   "name": "@internal/compute",
   "private": true,
   "version": "0.0.1",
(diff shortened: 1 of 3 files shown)


nicktrn:
ready

ericallam (approved):

myftija (approved):

PR #3348, opened 2026-04-08
feat(webapp): set application_name on prisma connections

nicktrn:
Sets `application_name` on the Prisma writer and replica connection strings using the existing `SERVICE_NAME` env var, so DB load can be attributed by service.


nicktrn:
ready

matt-aitken (approved):

Issue #3292, opened 2026-03-29
OTLP nanosecond timestamp overflow in webapp event repository


nicktrn:
Several places in the webapp multiply epoch milliseconds by 1,000,000 **before** converting to BigInt, which causes IEEE 754 precision loss (~256ns errors in ~0.2% of cases). The result exceeds `Number.MAX_SAFE_INTEGER` (~9e15) since epoch-ms * 1e6 is ~1.7e18.

The correct pattern is `BigInt(ms) * BigInt(1_000_000)` (multiplication after BigInt conversion). This already exists in `convertDateToNanoseconds()` in the same file - the buggy functions just didn't use it.

**Affected locations:**

1. `apps/webapp/app/v3/eventRepository/common.server.ts:24` - `getNowInNanoseconds()`
   ```typescript
   // Bug: multiplication in float-land
   return BigInt(new Date().getTime() * 1_000_000);
   // Fix:
   return BigInt(new Date().getTime()) * BigInt(1_000_000);
   ```

2. `apps/webapp/app/v3/eventRepository/common.server.ts:38` - `calculateDurationFromStart()`
   ```typescript
   // Bug:
   const duration = Number(BigInt($endtime.getTime() * 1_000_000) - startTime);
   // Fix:
   const duration = Number(BigInt($endtime.getTime()) * BigInt(1_000_000) - startTime);
   ```

3. `apps/webapp/app/v3/eventRepository/index.server.ts:217` - `recordRunDebugLog()`
4. `apps/webapp/app/v3/runEngineHandlers.server.ts:431` - retry event recording

**Impact:** Low. The errors are +-256 nanoseconds, which is unlikely to cause visible issues. But it's an easy fix and prevents any edge cases with span ordering.

**Context:** Found during compute workload manager PR review (#3114). The supervisor's `msToNano()` was fixed in that PR using a split approach that also preserves sub-ms precision from `performance.now()` arithmetic.

Linear: TRI-8269

aayushbaluni:
I've submitted PR #3381 fixing all instances of the BigInt overflow pattern, including the one at `runEngineHandlers.server.ts:432` that was missed in the previous PR #3378. However, the PR was auto-closed because I need to be vouched as a contributor. Could a maintainer please help with that?

AkhilTrivediX:
[...]

PR #3381, opened 2026-04-15
fix: convert to BigInt before nanosecond multiplication (all instances)

aayushbaluni:
## Summary

Fixes #3292

This is a resubmission of #3378 which was closed because it missed one instance of the bug. This PR fixes ALL instances.

## Root Cause

Multiple functions compute nanosecond timestamps as `BigInt(milliseconds * 1_000_000)`. When milliseconds exceed `Number.MAX_SAFE_INTEGER / 1_000_000`, the multiplication overflows IEEE 754 precision before the BigInt conversion.

## Fix

Convert to BigInt **before** multiplication in all instances:
```typescript
BigInt(milliseconds) * BigInt(1_000_000)
```

### Files changed
- `apps/webapp/app/v3/eventRepository/common.server.ts` - `getNowInNanoseconds`, `calculateDurationFromStart`, `calculateDurationFromStartJsDate`
- `apps/webapp/app/v3/eventRepository/index.server.ts` - `recordRunDebugLog`
- `apps/webapp/app/v3/runEngineHandlers.server.ts` - startTime calculation (previously missed)

### DIFF
--- a/apps/webapp/app/v3/eventRepository/common.server.ts
+++ b/apps/webapp/app/v3/eventRepository/common.server.ts
@@ -21,7 +21,7 @@ export function extractContextFromCarrier(carrier: Record<string, unknown>) {
 }
 
 export function getNowInNanoseconds(): bigint {
-  return BigInt(new Date().getTime() * 1_000_000);
+  return BigInt(new Date().getTime()) * BigInt(1_000_000);
 }
 
 export function getDateFromNanoseconds(nanoseconds: bigint): Date {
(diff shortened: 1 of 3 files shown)


aayushbaluni:
Hi! This PR was auto-closed because I'm not in the vouched users list. I understand the policy from CONTRIBUTING.md.

This is a resubmission of #3378 which was closed because it missed one instance of the BigInt overflow bug. This PR fixes ALL instances including the previously missed `runEngineHandlers.server.ts:432`.

The bug causes IEEE 754 precision loss when computing nanosecond timestamps — `BigInt(milliseconds * 1_000_000)` overflows before BigInt conversion. The fix converts to BigInt first: `BigInt(milliseconds) * BigInt(1_000_000)`.

Could a maintainer please vouch me so this can be reviewed? Happy to answer any questions about the fix.

PR #3439, opened 2026-04-23
ci: automate helm chart release alongside package release

nicktrn:
Wires up automatic Helm chart releases to ride along with the existing changeset-driven package release flow.

Today `Chart.yaml` is bumped by hand and `release-helm.yml` fires only when a human pushes a `helm-v*` tag. With this, the changeset release PR also carries a `Chart.yaml` bump so main always matches the published version, and `release.yml` invokes `release-helm.yml` via `workflow_call` after Docker images are published.

`helm-v${VERSION}` tag is pushed as a marker (same GITHUB_TOKEN trick as `v.docker.*`). Manual `helm-v*` tag flow still works. Chart.yaml consistency check in `release-helm.yml` is the safety net if the bump job ever drifts.

First rollout: the open `changeset-release/main` PR has stale Chart.yaml. Bump it manually on that branch before merging, otherwise the first automated helm release fails at the consistency check.

### DIFF
--- a/.github/workflows/changesets-pr.yml
+++ b/.github/workflows/changesets-pr.yml
@@ -121,3 +121,35 @@ jobs:
           else
             echo "No changes to commit"
           fi
+
+  bump-chart-version:
+    name: Bump Helm chart version on release PR
+    runs-on: ubuntu-latest
+    needs: update-lockfile
+    permissions:
+      contents: write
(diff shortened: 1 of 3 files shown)


nicktrn:
ready

myftija (approved):

Issue #3443, opened 2026-04-24
Vouch request: fix Helm chart CLICKHOUSE_URL literal-placeholder bug for self-hosters with external ClickHouse


ThullyoCunha:
### Why do you want to contribute?

I'm Thullyo Cunha, platform engineer at Meistrari (https://tela.com). We self-host Trigger.dev v4 on EKS for our internal task platform. While bringing it up against an external ClickHouse + `clickhouse.external.existingSecret` (the officially recommended path for not committing secrets to values.yaml), we hit a bug that blocks the webapp on first start:

```
goose run: parse "http://default:${CLICKHOUSE_PASSWORD}@<host>:8123?secure=false": net/url: invalid userinfo
```

**Root cause:** `hosting/k8s/helm/templates/_helpers.tpl` (v4.0.5, also on `main`) renders `CLICKHOUSE_URL` with a shell-style literal `${CLICKHOUSE_PASSWORD}` expecting bash expansion. But `docker/scripts/entrypoint.sh` does `export GOOSE_DBSTRING="$CLICKHOUSE_URL"` — single-pass POSIX sh substitution, so the inner `${...}` survives as literal text and goose rejects the URL.

**Fix:** switch the two helpers (`trigger-v4.clickhouse.url` and `trigger-v4.clickhouse.replication.url`, external+existingSecret branch) from shell-style `${CLICKHOUSE_PASSWORD}` to Kubernetes' `$(CLICKHOUSE_PASSWORD)`. Kubelet substitutes `$(VAR)` at container-creation time from earlier env entries, and the chart already declares `CLICKHOUSE_PASSWORD` from the secret right before `CLICKHOUSE_URL` — so the URL reaches the entrypoint with the real password already inlined. No entrypoint change, no image change, minimal blast radius, plain-password branch untouched.

I've verified the fix end-to-end in our test EKS cluster (webapp boots, goose applies all 6 migrations, runtime ClickHouse client connects). Branch ready at https://github.com/ThullyoCunha/trigger.dev/tree/fix/clickhouse-external-url-kubelet-expansion — would appreciate a vouch so I can open the PR.

I'd also be glad to add a short README/values.yaml note about the URL-userinfo-safety caveat on `CLICKHOUSE_PASSWORD` (kubelet substitutes verbatim without percent-encoding) as part of the same PR if that's useful.

### Prior contributions or relevant experience

- GitHub: https://github.com/ThullyoCunha
[...]

PR #3445, opened 2026-04-24
fix(helm): expand CLICKHOUSE_PASSWORD in webapp CLICKHOUSE_URL via kubelet

ThullyoCunha:
## Summary

When the official Helm chart is deployed with an external ClickHouse and `clickhouse.external.existingSecret` set — the documented path for not committing secrets to `values.yaml` — the webapp pod crash-loops on startup:

```
goose run: parse "http://default:${CLICKHOUSE_PASSWORD}@<host>:8123?secure=false": net/url: invalid userinfo
```

Context in vouch request #3443.

## Root cause

Two pieces interact:

1. `hosting/k8s/helm/templates/_helpers.tpl` renders `CLICKHOUSE_URL` (and `RUN_REPLICATION_CLICKHOUSE_URL`) with a shell-style literal `${CLICKHOUSE_PASSWORD}` expecting bash expansion at container start.
2. `docker/scripts/entrypoint.sh` does `export GOOSE_DBSTRING="$CLICKHOUSE_URL"` — single-pass POSIX sh substitution, so the inner `${...}` survives as literal text and goose rejects it.

Reproduces against the latest published chart (`oci://ghcr.io/triggerdotdev/charts/trigger:4.0.5`) and `main`.

## Fix

Switch the two helpers (external + `existingSecret` branch) from shell-style `${CLICKHOUSE_PASSWORD}` to Kubernetes' `$(CLICKHOUSE_PASSWORD)`. Kubelet substitutes `$(VAR)` at pod-creation time from earlier env entries, and the chart already declares `CLICKHOUSE_PASSWORD` from the Secret immediately before `CLICKHOUSE_URL`, so the URL reaches the entrypoint with the real password already inlined. No entrypoint change, no image change. The plain-password branch (no `existingSecret`) is unchanged.

Operator caveat added as template comments: `CLICKHOUSE_PASSWORD` must be URL-userinfo-safe since kubelet substitutes verbatim without percent-encoding. Hex-encoded passwords (e.g. `openssl rand -hex 32`) are safe by construction.

## Verification

- `helm template` against `external.existingSecret` now renders `value: "http://default:$(CLICKHOUSE_PASSWORD)@<host>:8123?secure=false"` (was `${CLICKHOUSE_PASSWORD}`).
- `helm template` against the plain-password branch is byte-identical to before.
- Deployed end-to-end on a staging EKS cluster (Meistrari platform): webapp container reaches `goose: successfully migrated database to version: 6`, Node.js ClickHouse client connects at runtime.

## Alternatives considered

- **Change `entrypoint.sh`** to `eval` / `envsubst` the URL — larger surface, touches every deployment mode (Docker Compose + k8s) and every container image.
- **Mirror the Postgres pattern** (chart reads the full URL via `valueFrom.secretKeyRef`, as in `trigger-v4.postgres.useSecretUrl`) — cleaner long-term but requires a new `values.yaml` field and a migration path for existing users. Happy to follow up with that as a separate PR if the minimal fix here isn't the preferred direction.

Closes the issue raised in #3443.

### DIFF
--- a/hosting/k8s/helm/templates/_helpers.tpl
+++ b/hosting/k8s/helm/templates/_helpers.tpl
@@ -400,6 +400,19 @@ ClickHouse hostname
 
 {{/*
 ClickHouse URL for application (with secure parameter)
+
+Note on the external+existingSecret branch: the password is expanded via
+Kubernetes' `$(VAR)` syntax, not shell `${VAR}`. Kubelet substitutes
+`$(CLICKHOUSE_PASSWORD)` at container-creation time from the
+CLICKHOUSE_PASSWORD env var declared just before CLICKHOUSE_URL in
+webapp.yaml. Shell-style `${...}` does not work here because
+`docker/scripts/entrypoint.sh` assigns CLICKHOUSE_URL to GOOSE_DBSTRING
+with a single-pass expansion (`export GOOSE_DBSTRING="$CLICKHOUSE_URL"`),
+so any inner `${...}` reaches goose verbatim and fails URL parsing.
+
+CLICKHOUSE_PASSWORD must contain only URL-userinfo-safe characters — the
+value is substituted verbatim, so `@ : / ? # [ ] %` break the URL. Use a
+hex-encoded password or percent-encode before storing in the Secret.
 */}}
 {{- define "trigger-v4.clickhouse.url" -}}
 {{- if .Values.clickhouse.deploy -}}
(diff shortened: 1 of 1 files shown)


ericallam:
@ThullyoCunha make sure to open the PR in draft status or it gets auto-closed

ThullyoCunha:
Thanks @ericallam — my bad, missed the draft requirement. Reopened as a draft in #3449 (same patch).

PR #3446, opened 2026-04-24
Fix debounce forwarding in batchTriggerAndWait arrays

GautamBytes:
## Summary

Fixes the array input path for `batchTriggerAndWait` so per-item `options.debounce` is forwarded instead of being dropped.

The async iterable path already forwarded debounce correctly; this brings the array serializer in line with that behavior for both `task.batchTriggerAndWait([...])` and `tasks.batchTriggerAndWait(id, [...])`.

Closes TRI-8310.


GautamBytes:
@ericallam can you check this , not sure why its closed even after claude not added me to vouch.yml

PR #2671, opened 2025-11-12
docs(kubernetes): use v4 helm chart instead of beta

mstruebing:
## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/triggerdotdev/trigger.dev/blob/main/CONTRIBUTING.md)
- [x] The PR title follows the convention.
- [x] I ran and tested the code works

---

## Testing

_[Describe the steps you took to test this change]_

I've executed the commands on my own Kubernetes cluster and it works.

---

## Changelog

_[Short description of what has changed]_

Updated the documentation to use the latest helm chart (v4) instead of beta.

---

## Screenshots

_[Screenshots]_

Screenshot 2025-11-12 at 14 26 30

💯

PR #3449, opened 2026-04-24
fix(helm): expand CLICKHOUSE_PASSWORD in webapp CLICKHOUSE_URL via kubelet

### LINKED ISSUE #3443: Vouch request: fix Helm chart CLICKHOUSE_URL literal-placeholder bug for self-hosters with external ClickHouse

ThullyoCunha:
## Summary

When the official Helm chart is deployed with an external ClickHouse and `clickhouse.external.existingSecret` set — the documented path for not committing secrets to `values.yaml` — the webapp pod crash-loops on startup:

```
goose run: parse "http://default:${CLICKHOUSE_PASSWORD}@<host>:8123?secure=false": net/url: invalid userinfo
```

Context in vouch request #3443. Re-opening in draft status per bot policy (previous attempt was #3445, closed by automation because it wasn't draft; no changes to the patch).

## Root cause

Two pieces interact:

1. `hosting/k8s/helm/templates/_helpers.tpl` renders `CLICKHOUSE_URL` (and `RUN_REPLICATION_CLICKHOUSE_URL`) with a shell-style literal `${CLICKHOUSE_PASSWORD}` expecting bash expansion at container start.
2. `docker/scripts/entrypoint.sh` does `export GOOSE_DBSTRING="$CLICKHOUSE_URL"` — single-pass POSIX sh substitution, so the inner `${...}` survives as literal text and goose rejects it.

Reproduces against the latest published chart (`oci://ghcr.io/triggerdotdev/charts/trigger:4.0.5`) and `main`.

## Fix

Switch the two helpers (external + `existingSecret` branch) from shell-style `${CLICKHOUSE_PASSWORD}` to Kubernetes' `$(CLICKHOUSE_PASSWORD)`. Kubelet substitutes `$(VAR)` at pod-creation time from earlier env entries, and the chart already declares `CLICKHOUSE_PASSWORD` from the Secret immediately before `CLICKHOUSE_URL`, so the URL reaches the entrypoint with the real password already inlined. No entrypoint change, no image change. The plain-password branch (no `existingSecret`) is unchanged.

Operator caveat added as template comments: `CLICKHOUSE_PASSWORD` must be URL-userinfo-safe since kubelet substitutes verbatim without percent-encoding. Hex-encoded passwords (e.g. `openssl rand -hex 32`) are safe by construction.

## Verification

- `helm template` against `external.existingSecret` now renders `value: "http://default:$(CLICKHOUSE_PASSWORD)@<host>:8123?secure=false"` (was `${CLICKHOUSE_PASSWORD}`).
- `helm template` against the plain-password branch is byte-identical to before.
- Deployed end-to-end on a staging EKS cluster (Meistrari platform): webapp container reaches `goose: successfully migrated database to version: 6`, Node.js ClickHouse client connects at runtime.

## Alternatives considered

- **Change `entrypoint.sh`** to `eval` / `envsubst` the URL — larger surface, touches every deployment mode (Docker Compose + k8s) and every container image.
- **Mirror the Postgres pattern** (chart reads the full URL via `valueFrom.secretKeyRef`, as in `trigger-v4.postgres.useSecretUrl`) — cleaner long-term but requires a new `values.yaml` field and a migration path for existing users. Happy to follow up with that as a separate PR if the minimal fix here isn't the preferred direction.

## Changeset

None added — the Helm chart isn't versioned through `@changesets/cli` (docs/chart-only PRs historically merge without a changeset, e.g. #2671). Happy to add one if the policy changed.

Closes #3443.

### DIFF
--- a/hosting/k8s/helm/templates/_helpers.tpl
+++ b/hosting/k8s/helm/templates/_helpers.tpl
@@ -400,6 +400,19 @@ ClickHouse hostname
 
 {{/*
 ClickHouse URL for application (with secure parameter)
+
+Note on the external+existingSecret branch: the password is expanded via
+Kubernetes' `$(VAR)` syntax, not shell `${VAR}`. Kubelet substitutes
+`$(CLICKHOUSE_PASSWORD)` at container-creation time from the
+CLICKHOUSE_PASSWORD env var declared just before CLICKHOUSE_URL in
+webapp.yaml. Shell-style `${...}` does not work here because
+`docker/scripts/entrypoint.sh` assigns CLICKHOUSE_URL to GOOSE_DBSTRING
+with a single-pass expansion (`export GOOSE_DBSTRING="$CLICKHOUSE_URL"`),
+so any inner `${...}` reaches goose verbatim and fails URL parsing.
+
+CLICKHOUSE_PASSWORD must contain only URL-userinfo-safe characters — the
+value is substituted verbatim, so `@ : / ? # [ ] %` break the URL. Use a
+hex-encoded password or percent-encode before storing in the Secret.
 */}}
 {{- define "trigger-v4.clickhouse.url" -}}
 {{- if .Values.clickhouse.deploy -}}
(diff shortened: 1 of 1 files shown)


nicktrn (approved):
thanks for this @ThullyoCunha 🙏

PR #3461, opened 2026-04-28
ci(helm): roll prereleases on main pushes + manual trigger

nicktrn:
Today the helm prerelease workflow only fires on PRs that touch `hosting/k8s/helm/**`. Two consequences we ran into:

1. The `changeset-release/main` PR's prerelease comment goes stale once the release branch gets force-pushed without a helm-touching commit (the bot's `Chart.yaml` bump alone doesn't seem to refire the trigger reliably).
2. The release PR's chart references an `appVersion` (e.g. `v4.4.5`) whose Docker images don't exist until *after* merge + tag. So that prerelease chart can't actually be installed end-to-end.

Renames the workflow to `helm-prerelease.yml` and adds two new triggers:

- **`push: main`** with `paths: hosting/k8s/helm/**` -> rolling prereleases versioned `<base>-main.<sha>`. `appVersion` stays at whatever `Chart.yaml` has (i.e. last released), so installs pull real images. Tests that chart structure is deployable, even if the app code is one release behind.
- **`workflow_dispatch`** with optional `app_version` input -> manually trigger a prerelease and optionally override `appVersion` (e.g. pin to `main` or a specific tag). Useful for testing chart + app-version combinations on demand.

PR behavior unchanged: same `<base>-pr<N>.<sha>` versioning, same posted/updated comment.

Why not also bypass paths for `changeset-release/main`? The release PR's chart references not-yet-built `v4.4.5` images, so those prereleases aren't actually installable. The rolling main prerelease covers the testable case better.

Why not SHA-pin `appVersion` to a built image like `main-<sha>`? Bigger change - the docker publish workflows currently only push `:main` (no SHA-suffixed tag). Worth doing later if we want first-class "install one chart, get exactly that commit's app code" testing, but out of scope here.

Diff is mostly a rename. Substantive changes:
- new `push` and `workflow_dispatch` triggers
- `prerelease` job `if:` extended for the new event types
- version logic branches per event
- new "Override appVersion" step (workflow_dispatch only)
- new "Write run summary" step so non-PR runs surface the install instructions
- PR comment steps gated on `github.event_name == 'pull_request'`
- concurrency group falls back to `github.ref` for non-PR runs


nicktrn:
ready

myftija (approved):

PR #3471, opened 2026-04-29
fix(webapp): preserve filters on queues page action redirects

nicktrn:
Queues page action handler was rebuilding the redirect URL with only `?page=`, so any pause/resume/override modal confirmation wiped the user's search query. With hundreds of queues filtered down to a handful, every confirmation dropped you back to the unfiltered list - and pagination still pointed at the previous numeric page, so you'd land on a different slice than you came from.

Swap the manual rebuild for `url.search` so the full querystring (including any future filter params) flows through. Drops the now-unused `SearchParamsSchema.parse` call inside `action`; the loader still validates on the way back.

### DIFF
--- a/.server-changes/queues-preserve-filters-on-action.md
+++ b/.server-changes/queues-preserve-filters-on-action.md
@@ -1,0 +1,6 @@
+---
+area: webapp
+type: fix
+---
+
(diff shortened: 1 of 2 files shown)


nicktrn:
ready

ericallam (approved):

PR #3483, opened 2026-04-30
feat(cli-v3): add --no-browser flag and examples to init/login --help

nicktrn:
Closes the most common friction point hit while setting up a fresh project from an agent harness: the CLI auto-opens the user's default browser during auth and there is no supported way to skip it (the existing `isLinuxServer()` path only triggers when `xdg-open` is missing entirely).

`--no-browser` on `login` and `init` prints the URL and waits to be visited from any browser. The flag threads through the embedded `login()` call inside `init`.

While here:
- `init` now errors loudly when stdin is non-TTY without `--yes` instead of default-and-exiting silently at the first prompt (which left the project half-initialized: deps installed, no config or example file).
- Both commands gain an `Examples` block in `--help` rendered between the description and the arguments/options list, so `--help | head` surfaces the common invocations.

Other commands also call `login()` embedded and would benefit from `--no-browser` too, but kept this PR scoped to the cases the friction log called out.

### DIFF
--- a/.changeset/cli-no-browser-headless-init.md
+++ b/.changeset/cli-no-browser-headless-init.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": patch
+---
+
+Add `--no-browser` flag to `init` and `login` to skip auto-opening the browser during authentication. Also error loudly when `init` is run without `--yes` under non-TTY stdin (previously default-and-exited silently, leaving the project half-initialized). Both commands now show an `Examples` section in `--help`.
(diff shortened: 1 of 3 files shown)


nicktrn:
ready

ericallam:
@nicktrn ci failures

nicktrn:
> @nicktrn ci failures

@ericallam pre-existing, flaky run engine tests

ericallam (approved):

PR #3488, opened 2026-05-01
fix(webapp): invalid HTML nesting in errors Activity tooltip

D-K-P:
The Activity peak count tooltip in the errors list rendered a `<button>` (from `SimpleTooltip`'s default `TooltipTrigger`) inside the row's `<a>` link (`TableCell to={errorPath}`). Interactive content nested inside other interactive content is invalid HTML and triggers accessibility warnings. Adding `asChild` to `SimpleTooltip` makes the existing `<span>` the trigger directly, removing the nested `<button>`.


D-K-P:
Ready to review

nicktrn (approved):

PR #3494, opened 2026-05-01
ci: pin actions to SHAs and add dependabot config

nicktrn:
Most actions in this repo were several major versions behind, which is why every CI run has been emitting Node 20 deprecation warnings.

Pinning every action to a commit SHA (with the version as a trailing comment) means each CI run uses the exact code that was reviewed when the bump landed, instead of whatever a maintainer last pointed the major tag at. Dependabot is configured to group all action bumps into one weekly PR with a 7-day cooldown.

Worth flagging:

- The Claude Code action ships ~daily but the model is set separately via `--model` in `claude_args`, so SHA-pinning the action gives reproducibility without locking the model.
- The kubeconform container is digest-pinned (`docker://image:tag@sha256:...`). Dependabot's github-actions ecosystem doesn't track `docker://` references ([explicit TODO in dependabot-core](https://github.com/dependabot/dependabot-core/blob/main/github_actions/lib/dependabot/github_actions/file_parser.rb)), so it needs manual bumps either way - but the digest pin protects against tag repointing for free.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 7
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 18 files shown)


nicktrn:
ready

ericallam (approved):

PR #3504, opened 2026-05-01
fix(helm): supervisor OTLP endpoint resolves cross-namespace

nicktrn:
Reported by external contributor. The supervisor template hardcoded a short DNS name for `OTEL_EXPORTER_OTLP_ENDPOINT`, which the supervisor then propagates verbatim into runner pods (`apps/supervisor/src/workloadManager/kubernetes.ts:196`). When runners are spawned in a different namespace via `supervisor.config.kubernetes.namespace`, the short name doesn't resolve and span/log export silently fails - runs complete fine but the dashboard shows nothing.

Same FQDN pattern the chart already uses for `TRIGGER_WORKLOAD_API_DOMAIN` (line 203). Verified with `helm template trigger . --namespace my-ns` - renders `http://trigger-webapp.my-ns.svc.cluster.local:3030/otel`.

Cheers Niels

### DIFF
--- a/hosting/k8s/helm/templates/supervisor.yaml
+++ b/hosting/k8s/helm/templates/supervisor.yaml
@@ -237,7 +237,7 @@ spec:
               value: {{ .Values.supervisor.config.debug | quote }}
             # OTEL
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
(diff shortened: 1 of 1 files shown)


nicktrn:
ready

ericallam (approved):

PR #3503, opened 2026-05-01
ci: add actionlint workflow

nicktrn:
Adds an `actionlint` job that runs on changes to `.github/workflows/**` and `.github/actions/**`. Catches workflow bugs at PR time — expression typos, deprecated runner labels, broken matrices, and shellcheck issues in `run:` blocks.

Run from the official `docker://rhysd/actionlint` image, digest-pinned alongside everything else.

Existing workflows had 6 shellcheck findings — fixed inline:

- 3× SC2086 (`>> $GITHUB_OUTPUT` → `>> "$GITHUB_OUTPUT"`) in `helm-prerelease.yml`, `release-helm.yml`, `release.yml`
- 1× SC2129 in `publish-webapp.yml` (grouped repeated redirects under one `{ ... } >> "$GITHUB_OUTPUT"`)
- 2× SC2193 false-positive in `publish-worker.yml` and `publish-worker-v4.yml` (the `[[ "${{ matrix.package }}" == *-provider ]]` pattern). Refactored to pass `matrix.package` via `env:` and use bash parameter expansion - cleaner and dodges the warning.

PR #3506, opened 2026-05-01
ci: add zizmor workflow security scanner

nicktrn:
Adds zizmor alongside the actionlint job from #3503. Both now run as parallel jobs in a single `.github/workflows/workflow-checks.yml`, triggered on `.github/workflows/**` and `.github/actions/**` changes.

Zizmor is configured with `unpinned-uses: hash-pin` policy via `.github/zizmor.yml`, so any future unpinned action will fail CI. Findings upload SARIF to the Security tab alongside CodeQL.

Bulk of the diff is cleanup of the findings zizmor surfaced on first run. `zizmor --fix=all` handled most of them mechanically; the rest were judgment calls.


nicktrn:
ready :lock:

ericallam (approved):

PR #3560, opened 2026-05-12
fix(core): cap idempotencyKey length at the API boundary

d-cs:
`tasks.trigger`, `tasks.batchTrigger`, `batch.create`, `wait.createToken`, `wait.forDuration`, and the input/session stream waitpoint endpoints all accept a caller-supplied `idempotencyKey` and store it verbatim against a composite-unique index on `TaskRun`, `BatchTaskRun`, or `Waitpoint`. The schemas had no length cap, so a sufficiently long high-entropy key produced an index row larger than the underlying storage layer can hold. The insert failed at the database, and the caller saw a generic 500 from `RunEngineTriggerTaskService.call()` / `CreateBatchService` / waitpoint creation, depending on the endpoint.

Keys produced by `idempotencyKeys.create()` are 64-character SHA-256 hashes and never trip this — it only manifests for direct REST callers (or SDK callers passing a raw string they generated themselves). Low-entropy keys also sail through, because the storage layer compresses repeated bytes before they reach the index, which is why the failure mode is intermittent and tied to caller-side key shape.

## Fix

Add `.max(2048, "<field> must be 2048 characters or less")` to the seven schemas that feed an indexed `idempotencyKey` column:

- `TriggerTaskRequestBody.options.idempotencyKey`
- `BatchTriggerTaskItem.options.idempotencyKey`
- `CreateBatchRequestBody.idempotencyKey`
- `CreateWaitpointTokenRequestBody.idempotencyKey`
- `CreateInputStreamWaitpointRequestBody.idempotencyKey`
- `CreateSessionStreamWaitpointRequestBody.idempotencyKey`
- `WaitForDurationRequestBody.idempotencyKey`

Plus the `idempotency-key` HTTP header on the trigger route (and the three batch routes that re-export `HeadersSchema`). The header schema is lifted out of `api.v1.tasks.$taskId.trigger.ts` into `apps/webapp/app/v3/triggerHeaders.server.ts` so it can be exercised in tests without dragging the route's import-time side effects.

The 2048 character ceiling is chosen to sit safely under the per-row index limit while staying generous against existing callers — keys that fit before still fit. Oversized keys now return a structured Zod 400 instead of a generic 500.

Limit is documented under `Idempotency key` in `docs/limits.mdx` and as a `<Note>` on `docs/idempotency.mdx`.

## Test plan

- [x] 15 schema unit tests added (`packages/core/src/v3/schemas/idempotencyKey.test.ts`, `apps/webapp/test/routes/triggerHeaders.test.ts`) — rejection-with-message + boundary acceptance for each capped schema. The webapp test exercises the extracted `TriggerHeadersSchema` directly with no mocks.
- [x] `pnpm run build --filter @trigger.dev/core`
- [x] `pnpm run typecheck --filter webapp`
- [x] End-to-end verified locally: baseline (small key) → 200; 3000-char high-entropy header → 400 with the expected Zod error; same key at the 2048 boundary → 200; same key with the cap reverted → the database rejected the insert and the route returned 500 to the caller. Cap restored.


d-cs:
Verified — keeping as is.

The `.optional()` on `traceparent` and `tracestate` predates this PR; the original `HeadersSchema` on `main` had the same mix (lines 30–43 of `apps/webapp/app/routes/api.v1.tasks.$taskId.trigger.ts` before the extraction). I copied the schema verbatim when lifting it into `triggerHeaders.server.ts` so the test could import it without route side effects, so this is unchanged behavior, not a new inconsistency.

Functionally `.optional()` and `.nullish()` are equivalent for these fields — HTTP headers reach the schema as `string | undefined`, never literal `null` — so the diff would be cosmetic only. Out of scope for an idempotency-key cap PR; happy to harmonize in a separate cleanup if anyone wants to take it.

ericallam:
@d-cs Usually we split out docs updates into their own PRs (cause merging docs changes into main causes the docs site to be updated immediately, and there is usually a decent time between merging a change into main and releasing to users), but in this case its A-OK because the change is pretty minor and fine to go live "early".

apps/webapp/app/routes/api.v1.tasks.$taskId.trigger.ts
```diff
@@ -28,7 +28,10 @@ const ParamsSchema = z.object({
 });
 
 export const HeadersSchema = z.object({
-  "idempotency-key": z.string().nullish(),
+  "idempotency-key": z
+    .string()
+    .max(2048, "idempotency-key must be 2048 characters or less")
```
ericallam, line 33: What if instead of rejecting, we just truncated to 2048 characters instead?

d-cs: I don't think that would be desirable, might be a user is using a defective script to generate keys (think abnormally long prefixes) and the truncation would allow accidentally bypassing the protection. Failing at this point prevents a potential hidden issue later down the line

ericallam: Makes sense!

ericallam (approved):

PR #3562, opened 2026-05-12
fix(core): use Promise.allSettled in TracingSDK flush/shutdown

### LINKED ISSUE #3556: bug(v4): TracingSDK.flush() uses Promise.all, drops trace + log exports on shutdown when one provider rejects
## Problem

Image

On self-hosted Trigger.dev v4.4.5, the user task's `logger.info(...)` calls and child-trigger spans are silently dropped from a run's dashboard trace when `processKeepAliveEnabled` is `false` (the default for self-host). Engine state events (`QUEUED` / `PENDING_EXECUTING` / `EXECUTING` / `run:subscribe`) still appear because they're written server-side; only the user-emitted telemetry vanishes.

## Environment

- Trigger.dev v4.4.5 (Helm chart `oci://ghcr.io/triggerdotdev/charts/trigger:4.4.5`, image `ghcr.io/triggerdotdev/trigger.dev:v4.4.5`)
- Self-hosted on Kubernetes (EKS), single-shard ClickHouse from the chart
- Node 22.16.0 task runtime, `processKeepAliveEnabled: false`
- `OTEL_EXPORTER_OTLP_ENDPOINT=http://trigger-webapp:3030/otel`

## Root cause

`TracingSDK.flush()` at [`packages/core/src/v3/otel/tracingSDK.ts#L371-L377`](https://github.com/triggerdotdev/trigger.dev/blob/v4.4.5/packages/core/src/v3/otel/tracingSDK.ts#L371-L377):

```ts
public async flush() {
  await Promise.all([
    this._traceProvider.forceFlush(),
    this._logProvider.forceFlush(),
    this._meterProvider.forceFlush(),
  ]);
}
```

`Promise.all` semantics: the first rejection rejects the combined promise immediately, abandoning the other two providers' in-flight HTTP exports. The runner-side flush is invoked from the worker handler at [`packages/cli-v3/src/entryPoints/managed-run-worker.ts#L716-L731`](https://github.com/triggerdotdev/trigger.dev/blob/v4.4.5/packages/cli-v3/src/entryPoints/managed-run-worker.ts#L716-L731), which races a 5 s timeout — but a rejection short-circuits the race in milliseconds.

`TracingSDK.shutdown()` at L379-L385 has the same bug.

## Evidence

Runner pod stdout at task completion (from CloudWatch, redacted run ID):

```
00:25:54.571  [execution] executing task run process    ← task fork spawned
       ── 13.2 s of task code (with multiple logger.info calls + batchTrigger spans) ──
00:26:07.791  [execution] completed run attempt
00:26:07.792  [taskRunProcessProvider] Keep-alive disabled, cleaning up process immediately
00:26:07.795  Flushed usage in 1.032 ms
00:26:07.797  Flushed runMetadata successfully
00:26:07.797  Failed to flush tracingSDK              ← here
00:26:07.797  Flushed all in 4.039 ms
00:26:07.878  [controller] Exiting process
```

Total tracingSDK flush attempt: **4 ms** — well under the 5000 ms timeout. So one of `_traceProvider.forceFlush()` / `_logProvider.forceFlush()` / `_meterProvider.forceFlush()` rejected effectively synchronously, taking the entire flush down with it via \`Promise.all\`.

We verified separately that the backing infra was fully healthy throughout the window:
- ClickHouse pod: 0 restarts, ~34% memory utilization, container Running, PVC bound.
- Webapp's `/otel/v1/logs` endpoint was returning steady 200s.
- Webapp's `ClickhouseEventRepository.flushBatch` was logging `Inserted batch into clickhouse` with non-zero `written_rows` every few seconds.

So the data simply never leaves the runner.

Most likely rejecter: `_meterProvider.forceFlush()`. The metrics pipeline is the newest moving part in this path — wraps the OTLP metric exporter in a custom `BufferingMetricExporter` (tracingSDK.ts:311), introduced in #3061 with follow-up #3106. The trace and log paths are older and use stock OTel batch processors. But the `Promise.all` bug applies regardless of which provider rejects.

## Suggested fix

Swap `Promise.all` for `Promise.allSettled` in both `TracingSDK.flush()` and `TracingSDK.shutdown()`, and log per-provider rejections without bubbling them up:

```ts
public async flush() {
  const results = await Promise.allSettled([
    this._traceProvider.forceFlush(),
    this._logProvider.forceFlush(),
    this._meterProvider.forceFlush(),
  ]);
  results.forEach((r, i) => {
    if (r.status === 'rejected') {
      const name = ['trace', 'log', 'meter'][i];
      console.error(\`Failed to flush \${name} provider:\`, r.reason);
    }
  });
}
```

That preserves visibility into which provider failed (currently lost—the worker just logs "Failed to flush tracingSDK" with no provider attribution) while letting the other two providers' flushes complete rather than being abandoned.

## Impact

Any task running on a worker with `processKeepAliveEnabled: false` is at risk of losing all user-emitted trace data on shutdown if any of the three providers rejects `forceFlush()`. For short tasks (under ~5–10 s), this manifests as a run with engine events only and no user spans/logs.

## Reproduction (qualitative)

1. Self-host v4.4.5 with the default Helm values.
2. Deploy a task that calls \`logger.info(...)\` a few times and does \`tasks.batchTriggerAndWait\` on a small batch. Keep total task duration under 30 s.
3. Trigger the task in prod.
4. Open the run trace in the dashboard 1 min later.
5. Observe: engine events present; \`logger.info\` lines and \`batchTriggerAndWait\` spans absent. Runner pod stdout (if accessible) shows \`Failed to flush tracingSDK\`.

ericallam:
## Summary

- Switch `TracingSDK.flush()` and `shutdown()` from `Promise.all` to `Promise.allSettled`
- Prevents one provider's rejection from abandoning other providers' in-flight exports
- Adds per-provider error logging for better debugging

Fixes #3556

## Test plan

- [ ] Deploy to self-hosted environment with `processKeepAliveEnabled: false`
- [ ] Run task with `logger.info()` calls and verify they appear in trace
- [ ] Verify no "Failed to flush tracingSDK" errors in runner logs

### DIFF
--- a/.changeset/fix-tracing-sdk-flush.md
+++ b/.changeset/fix-tracing-sdk-flush.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/core": patch
+---
+
+Fixed TracingSDK.flush() and shutdown() to use Promise.allSettled instead of Promise.all, preventing one provider's rejection from abandoning the other providers' in-flight exports. This fixes an issue where user-emitted trace data (logger.info calls, child spans) could be silently dropped on shutdown when any provider fails to flush.
--- a/packages/core/src/v3/otel/tracingSDK.ts
+++ b/packages/core/src/v3/otel/tracingSDK.ts
@@ -369,19 +369,31 @@ export class TracingSDK {
   }
 
   public async flush() {
-    await Promise.all([
+    const results = await Promise.allSettled([
       this._traceProvider.forceFlush(),
       this._logProvider.forceFlush(),
       this._meterProvider.forceFlush(),
     ]);
+    const providerNames = ["trace", "log", "meter"] as const;
+    results.forEach((result, index) => {
+      if (result.status === "rejected") {
+        console.error(`Failed to flush ${providerNames[index]} provider:`, result.reason);
+      }
+    });
   }
 
   public async shutdown() {
-    await Promise.all([
+    const results = await Promise.allSettled([
       this._traceProvider.shutdown(),
       this._logProvider.shutdown(),
       this._meterProvider.shutdown(),
     ]);
+    const providerNames = ["trace", "log", "meter"] as const;
+    results.forEach((result, index) => {
+      if (result.status === "rejected") {
+        console.error(`Failed to shutdown ${providerNames[index]} provider:`, result.reason);
+      }
+    });
   }
 }
 
--- a/packages/core/src/v3/otel/tracingSDK.ts
+++ b/packages/core/src/v3/otel/tracingSDK.ts
@@ -376,14 +376,15 @@ export class TracingSDK {
     ]);
     const providerNames = ["trace", "log", "meter"] as const;
     const errors: Error[] = [];
+    const failedNames: string[] = [];
     results.forEach((result, index) => {
       if (result.status === "rejected") {
-        console.error(`Failed to flush ${providerNames[index]} provider:`, result.reason);
+        failedNames.push(providerNames[index]);
         errors.push(result.reason instanceof Error ? result.reason : new Error(String(result.reason)));
       }
     });
     if (errors.length > 0) {
-      throw new AggregateError(errors, "One or more providers failed to flush");
+      throw new AggregateError(errors, `One or more providers failed to flush: ${failedNames.join(", ")}`);
     }
   }
 
@@ -395,14 +396,15 @@ export class TracingSDK {
     ]);
     const providerNames = ["trace", "log", "meter"] as const;
     const errors: Error[] = [];
+    const failedNames: string[] = [];
     results.forEach((result, index) => {
       if (result.status === "rejected") {
-        console.error(`Failed to shutdown ${providerNames[index]} provider:`, result.reason);
+        failedNames.push(providerNames[index]);
         errors.push(result.reason instanceof Error ? result.reason : new Error(String(result.reason)));
       }
     });
     if (errors.length > 0) {
-      throw new AggregateError(errors, "One or more providers failed to shutdown");
+      throw new AggregateError(errors, `One or more providers failed to shutdown: ${failedNames.join(", ")}`);
     }
   }
 }
--- a/packages/core/src/v3/otel/tracingSDK.ts
+++ b/packages/core/src/v3/otel/tracingSDK.ts
@@ -375,11 +375,16 @@ export class TracingSDK {
       this._meterProvider.forceFlush(),
     ]);
     const providerNames = ["trace", "log", "meter"] as const;
+    const errors: Error[] = [];
     results.forEach((result, index) => {
       if (result.status === "rejected") {
         console.error(`Failed to flush ${providerNames[index]} provider:`, result.reason);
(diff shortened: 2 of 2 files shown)


ericallam:
@claude can you address this review:

With Promise.allSettled, flush() and shutdown() now always resolve (never reject), even when individual providers fail. This changes the public API contract. The callers in managed-run-worker.ts:725-738 and dev-run-worker.ts:724-737 (flushTracingSDK) wrap _tracingSDK?.flush() in Promise.race inside an outer Promise.allSettled (managed-run-worker.ts:673, dev-run-worker.ts:687), which tracks success/failure via successfulFlushes/failedFlushes. After this PR, flushTracingSDK will always resolve successfully, so "tracingSDK" will always appear in successfulFlushes even if underlying providers failed. The failure tracking at the flushAll level becomes less precise for tracing SDK failures specifically — though the individual provider errors are still logged to console by the new code. This is arguably the right tradeoff (partial success is better than total failure), but reviewers should be aware the caller-level failure detection for tracing is now effectively bypassed.

ericallam:
@claude improved this pr based on this review:

Code review — PR #3562 (fix(core): use Promise.allSettled in TracingSDK flush/shutdown)

  Verdict

  ACCEPT-WITH-NITS — small, well-targeted fix. Rebase before merge.

  Important findings

  None.

  Nits

  1. 🟡 Branch is behind main — diff includes an accidental revert of PR #3552 (SIGSEGV retry).
  → .changeset/retry-sigsegv.md, packages/core/src/v3/errors.ts:361,398, packages/core/test/errors.test.ts
  → Branch was cut before 8e675a4 landed on main; merging as-is would silently revert SIGSEGV retry behavior, its changeset, and its tests.
  → Rebase onto current main before merging.
  2. 🟡 Duplicate failure reporting. Each rejected provider is console.error'd inside flush()/shutdown() and then thrown as part of the AggregateError. Existing callers (e.g. managed-run-worker.ts:690-696) already
  log rejected reasons, so each failure prints twice.
  → packages/core/src/v3/otel/tracingSDK.ts:381,400
  → Drop the per-provider console.error and rely on the AggregateError carrying the cause, or accept the duplication.
  3. 🟡 AggregateError.message shadows per-provider detail. Existing caller at managed-run-worker.ts:692 logs result.reason (the AggregateError), which surfaces only the generic prefix unless .errors is inspected.
  → packages/core/src/v3/otel/tracingSDK.ts:386-388,405-407
  → Include provider names in the aggregate message (e.g. `One or more providers failed to flush: ${failedNames.join(", ")}`) so the existing log line stays informative.

  Strengths

  - Success-path behavior is identical to before — no extra logs, no spurious throw, same return.
  - Provider-name array order correctly matches Promise.allSettled input order, so error logs identify the right exporter.
  - Failure semantics preserved via AggregateError — callers that previously detected failure via a thrown rejection still do.
  - AggregateError is safe for this package (target: es2022, engines.node >=18.20.0).
  - Changeset present, correctly scoped to @trigger.dev/core as patch.

PR #3650, opened 2026-05-17
fix(sdk): cache realtime-stream create response per (runId, key)

ericallam:
## Summary

Stops `chat.response.write` / `streams.writer` hot-loops from issuing a fresh `PUT /realtime/v1/streams/:runId/self/:key` (and the `realtimeStreams || $1` array push on `TaskRun` that goes with it) for every chunk.

### Root cause

Each call into `streams.pipe()` constructs a brand-new `StreamInstance`, which calls `apiClient.createStream(runId, "self", key)`. The server route handler unconditionally runs `prisma.taskRun.update({ data: { realtimeStreams: { push: streamId } } })` to record the stream in the run's tracking array, then mints an S2 access token and returns it. For S2/v2 streams the access token is scoped to the org basin (1-day server-side TTL) and the stream name is deterministic from `(runId, streamId)`, so every PUT for the same `(runId, streamId)` returns equivalent credentials. The only thing differing per call is the array push.

`chat.response.write(part)` opens a one-shot stream per chunk via `streams.writer(CHAT_STREAM_KEY, { execute: ({ write }) => write(part) })`. Customers running per-chunk writers inside a `chat.agent` turn end up doing hundreds-to-thousands of PUTs against the same TaskRun row, with each PUT rewriting an ever-growing TOAST'd array. Under concurrency this monopolised the writer pool with `Lock:tuple` contention on the writer DB.

### Fix

`StandardRealtimeStreamsManager` now keeps a `Map<string, Promise<CreateStreamResponseLike>>` keyed by `${runId}:${key}`. First call PUTs as before; subsequent calls reuse the cached promise and construct a fresh `StreamsWriterV2` straight from the cached access token / basin / stream name. Net effect for hot-loop writers is **one PUT per `(run, stream-key)`** for the lifetime of the SDK process.

`StreamInstance` grows an optional `createStream` callback so the manager can inject the cached resolver while preserving the standalone-construction path used by tests.

### Risk & invalidation

- S2 v2 access tokens default to a 1-day server-side TTL (`accessTokenExpirationInMs`). The server SWR-caches them per `(basin, streamPrefix)` so the value the SDK gets is always a live token; reusing it within a single run is safe.
- Cache evicts on `createStream` failure (poisoned-promise guard).
- Cache clears on `manager.reset()` for parity with the existing `activeStreams` behaviour.
- v1 streams also benefit incidentally: subsequent `pipe()` calls for the same `(runId, key)` skip the array push, so the run's `realtimeStreams` no longer accumulates duplicates of the same id.

## Test plan

- [ ] Vitest suite for `realtimeStreams` still passes (`pnpm run test --filter @trigger.dev/core`).
- [ ] Cut a chat-prerelease snapshot from this branch and run a `chat.agent` reference task that emits N `chat.response.write` calls in a tight loop; verify exactly **one** `PUT /realtime/v1/streams/...` lands on the server and the run's `TaskRun.realtimeStreams` ends with exactly one entry per stream key.
- [ ] Confirm the run's S2 stream still receives all N chunks (writer reuse, no data loss).
- [ ] Confirm `streams.pipe()` for multiple distinct keys in the same run each gets its own PUT (cache scoped per `(runId, key)`).

### DIFF
--- a/.changeset/realtime-streams-creds-cache.md
+++ b/.changeset/realtime-streams-creds-cache.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/core": patch
+---
+
+Cache the `PUT /realtime/v1/streams/:runId/self/:key` response per `(runId, key)` so repeated `streams.pipe()` / `chat.response.write` / `chat.stream.writer` calls reuse the same S2 credentials instead of issuing a fresh PUT (and the `realtimeStreams || $1` array push it triggers on `TaskRun`) for every chunk. Hot-loop writers, most notably `chat.response.write` called per chunk inside a `chat.agent` turn, now do one PUT per `(run, stream-key)` instead of one per write, eliminating the writer-pool lock contention that scaled with the customer's chunk rate. S2 v2 access tokens are scoped to the org basin with a 1-day server-side TTL so reusing them across calls within a single run is safe; the cache evicts on `createStream` failure and on `manager.reset()`.
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "vite-tsconfig-paths": "^4.0.5",
     "vitest": "3.1.4"
   },
-  "packageManager": "pnpm@10.23.0",
+  "packageManager": "pnpm@10.33.2",
   "dependencies": {
     "@changesets/cli": "2.26.2",
     "@remix-run/changelog-github": "^0.0.5",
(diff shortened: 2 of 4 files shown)


ericallam:
Closing — the credential-cache fix is already published as `@trigger.dev/core@0.0.0-chat-prerelease-20260517214402` (built from this branch). Mainline integration for the 4.5 release is tracked in our internal issue tracker.

PR #3664, opened 2026-05-18
fix(webapp): catch loader/action throws before Remix serializes them

d-cs:
## Summary

Companion to #3536, which patched routes that already had a leaking `catch (e) { return json({error: e.message}, 500) }`. That pattern can't reach routes which have no catch in the first place — when those throw, Remix's default error path serializes `error.message` into the response body, and the SDK then wraps the leaked string as `TriggerApiError`.

Across 28 raw api.v1 loaders/actions plus one dashboard polling endpoint, each handler now:

- Wraps its body in `try { ... } catch (error) { ... }`.
- Re-throws `Response` instances so auth helpers' `throw json(...)` / `throw redirect(...)` pass through unchanged.
- Logs non-Response errors via `logger.error` so server-side visibility is preserved.
- Returns a generic body — `{"error": "Internal Server Error"}` 500 for raw API routes, or `{ changelogs: [] }` 200 for the polling widget (degrade silently across transient blips; the consumer hook already coped with empty payloads).

For six routes where #3536 left an inner try/catch covering only a service call (`alertChannels`, `batches.results`, `deployments.finalize`, `deployments.background-workers`, `deployments.promote`, `projects.background-workers`): an outer try/catch is added so auth/parsing failures are also sanitized. Inner typed-error handling (`ServiceValidationError` → 422 with message, etc.) is preserved exactly.

For two routes whose existing catch returned 400 + `error.message` (`api.v1.authorization-code`, `api.v1.orgs.\$orgParam.projects` action): the body is sanitized to a generic per-route string. **Status code stays 400** — clients that key on the 4xx/5xx distinction (and the SDK's no-retry-on-4xx behavior) are unaffected.

## Test plan

- [x] \`pnpm run typecheck --filter webapp\`
- [x] Per-route synthetic-throw probe: inject \`throw new Error("SYNTHETIC ...")\` at the top of each catch'd try, curl the route with a dummy bearer, confirm the response body is the generic shape and that the synthetic message lands server-side via \`logger.error\`. 29 routes verified.
- [x] Real-P1001 probe on the envvars loader: \`docker stop database\` mid-flight, confirm response is generic 500 (not the leaked Prisma message).
- [x] Sampled legitimate 4xx/2xx paths across each pattern variant (naked-wrap, partial-expanded, 400-preserved) to confirm the wraps don't interfere with normal control flow.


apps/webapp/app/routes/api.v1.projects.$projectRef.envvars.$slug.$name.ts
```diff
@@ -22,73 +23,82 @@ export async function action({ params, request }: ActionFunctionArgs) {
...
+          keepEmptyValues: true,
+        });
+
+        if (result.success) {
+          return json({ success: true });
+        } else {
+          return json({ error: result.error }, { status: 400 });
+        }
       }
     }
```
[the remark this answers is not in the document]
d-cs: Out of scope for this PR — the switch's missing default is pre-existing behavior, not introduced by the leak-sanitization changes. Worth a follow-up if we want to add explicit 405s across the action routes here.

.server-changes/sanitize-api-loader-action-leaks-sweep.md
```diff
@@ -0,0 +1,10 @@
...
+
+Each handler now wraps its body in try/catch, re-throws `Response` instances so auth helpers' `throw json(...)` / `throw redirect(...)` pass through unchanged, logs non-Response errors, and returns `{ error: "Internal Server Error" }` 500. For routes that already had an inner try/catch covering a service call but with auth/parsing outside the try (alertChannels, batches.results, deployments.finalize, several others), an outer try/catch is added so the inner typed-error handling is preserved. The `api.v1.authorization-code.ts` catch branch was returning `error.message` verbatim — switched to a generic body.
+
+Each route was verified locally with a synthetic-throw probe: inject `throw new Error("SYNTHETIC ...")` at the top of the catch'd try, curl the route with a dummy bearer token, confirm the response body is the generic shape and that the synthetic message is captured server-side via `logger.error`.
```
ericallam: The .server-changes files should be more like a high-level 1 or 2 sentence summary of the change. Imagine that it will eventually become a single item in the list of changes in a changelog. For example, this one should just be something like "Expands better API error messages across additional API endpoints" or something like that.

.server-changes/sanitize-loader-action-leaks.md
```diff
@@ -0,0 +1,8 @@
+---
+area: webapp
+type: fix
+---
+
+Wrap two loaders/actions that previously let thrown errors propagate to Remix's default 500 serializer, which writes `error.message` into the response body. When the underlying call (Prisma, etc.) fails, the raw error string was reaching API consumers — including the SDK, which surfaces it back to users via `TriggerApiError`. Each handler now catches non-Response errors, logs server-side, and returns a generic 500 body. `throw json(...)` / `throw redirect(...)` from auth helpers is re-thrown unchanged.
```
ericallam: Definitely no need for two .server-changes files in this pr

ericallam (changes requested):

.server-changes/sanitize-api-loader-action-leaks-sweep.md
d-cs: Collapsed both files into one and rewrote as a single changelog-style sentence in af7bbdec9.

.server-changes/sanitize-loader-action-leaks.md
d-cs: Removed the second file and trimmed this one to a single-sentence summary in af7bbdec9.

ericallam (approved):

PR #3686, opened 2026-05-21
feat(webapp): add RUNTIME_API_ORIGIN to decouple runner traffic from external origin

### LINKED ISSUE #2821: bug: OpenTelemetry Spans Not Displaying in Self-Hosted Instance Timeline

 **Self-hosted Instance:**                                                                                                                    │ │
- Trigger.dev Version: `main` (Docker image from 2024-12-26)                                                                                 │ │
 - Deployment: Docker Compose (official hosting/docker/webapp setup)                                                                          
 - Instance URL: trigger.devneurox.tech                                                                                                       │ │
 - Docker Images:                                                                                                                             │ │
 - Webapp: `ghcr.io/triggerdotdev/trigger.dev:main` (c3af7f05c1da)                                                                          │ │
  - Supervisor: `ghcr.io/triggerdotdev/supervisor:main` (d3d1d21ad176)

## Issue: OpenTelemetry Spans Not Displaying in Self-Hosted Instance Timeline

**Environment:**
- Trigger.dev SDK: v4.3.1
- Self-hosted instance: trigger.devneurox.tech
- Cloud instance: cloud.trigger.dev

**Problem Description:**

OpenTelemetry spans created with `logger.trace()` are **not visible** in the timeline UI on our self-hosted Trigger.dev instance, but they **work correctly** on the official cloud instance (cloud.trigger.dev).

**Configuration:**

`trigger.config.ts`:
```typescript
import { defineConfig } from "@trigger.dev/sdk";
import { HttpInstrumentation } from "@opentelemetry/instrumentation-http";
import { FetchInstrumentation } from "@opentelemetry/instrumentation-fetch";

export default defineConfig({
  project: "proj_iahascmunqsxfplmrepp",
  runtime: "node",
  logLevel: "log",
  telemetry: {
    instrumentations: [
      new HttpInstrumentation(),
      new FetchInstrumentation(),
    ],
  },
  // ...
});

Code Example:

const marketData = await logger.trace("bybit-api-calls", async (span) => {
  span.setAttribute("symbol", input.symbol);
  span.setAttribute("endpoints", "getTicker, getOpenInterest, getLongShortRatio, getFundingRate");
  return await bybitAPI.getMarketData(input.symbol);
});

Expected Behavior:

Timeline should show individual spans with hierarchy and timing.

Actual Behavior:

Self-Hosted: Timeline shows only one solid green bar - no spans visible
Cloud (cloud.trigger.dev): Timeline correctly shows all spans ✅
Steps to Reproduce:

Deploy same code with [logger.trace()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) spans to both self-hosted and cloud
Trigger task execution
Compare timeline view in dashboard
Additional Info:

Tasks execute successfully on both instances. Console logs show correct timing. Issue appears UI-specific to self-hosted.

Question: What self-hosted dashboard version is required for span visualization?

### Reproduction repo

N/A - Private repository. Can provide minimal reproduction example if needed.

### To reproduce

**Steps to reproduce:**

1. Configure Trigger.dev with OpenTelemetry instrumentations:
```typescript
telemetry: {
  instrumentations: [
    new HttpInstrumentation(),
    new FetchInstrumentation(),
  ],
}
2. Add [logger.trace() spans in your task code:
const result = await logger.trace("operation-name", async (span) => {
  span.setAttribute("key", "value");
  return await someOperation();
});

3 Deploy to self-hosted instance (trigger.devneurox.tech)
4 Deploy same code to cloud instance (cloud.trigger.dev)
5 Execute task on both instances
6 Compare timeline view in dashboard
Expected: Timeline shows individual spans with hierarchy and timing

Actual on self-hosted: Timeline shows only one solid green bar - no span breakdown visible

Actual on cloud: Timeline correctly displays all spans with proper hierarchy ✅

Note: Tasks execute successfully on both - issue is UI-only on self-hosted.

### Additional information

<img width="751" height="345" alt="Image" src="https://github.com/user-attachments/assets/6eb7b2a4-6265-4c92-a2bd-d6d72e3eaa02" />

<img width="647" height="418" alt="Image" src="https://github.com/user-attachments/assets/d506c4ab-7812-4d50-ba20-73e56df4a86c" />

### PR DESCRIPTION
**Pull request by ThullyoCunha**
Closes #2821

## ✅ Checklist

---

## Summary

The webapp publishes `API_ORIGIN` to runner pods as `TRIGGER_API_URL`, so runner-to-webapp traffic flows back through whatever URL is configured for external clients. Self-hosting behind a tracing-enabled gateway (Envoy, Istio, kgateway, ...) breaks the parent->child run link in trigger.dev's run-detail tree because the gateway's W3C `traceparent` rewrite on egress overwrites the SDK's `triggerAndWait()` span id. The webapp then writes that gateway-generated span id as the child run's `parentSpanId`, which never reaches the trigger event store, so the child renders as an orphan in the UI.

We hit this on our self-hosted v4 cluster running kgateway with tracing enabled (`spawnUpstreamSpan: true`). Reproduced with three rounds of SDK debug instrumentation (capturing the active span at `propagation.inject`, the wire `undici:request:headers` payload, and the value the webapp receives at the route level) plus a direct curl bypassing each hop until we isolated kgateway as the rewriter. Details on the investigation are in #2821 — several self-hosted users report the same symptom.

This PR splits the two concerns without sacrificing external auth/callbacks/UI flows that rely on the public `API_ORIGIN`:

- Set `RUNTIME_API_ORIGIN=http://<service>.<namespace>:<port>` (k8s) or `http://webapp:3000` (docker) to keep runner->webapp traffic on a cluster-internal hop that bypasses the gateway.
- Leave `API_ORIGIN` on the public URL so the dashboard, magic-link emails, waitpoint callbacks, and API `apiUrl` responses keep working for external clients.

The new env is optional and falls back to `API_ORIGIN`/`APP_ORIGIN`, so existing deployments are unaffected.

## Changes

- `apps/webapp/app/env.server.ts`: new optional `RUNTIME_API_ORIGIN` env.
- `apps/webapp/app/v3/environmentVariables/environmentVariablesRepository.server.ts`: prefer `RUNTIME_API_ORIGIN` when resolving `TRIGGER_API_URL`/`TRIGGER_STREAM_URL` for both dev and prod runner pods, falling back to the existing chain.
- `hosting/k8s/helm/values.yaml` + `templates/webapp.yaml`: expose `webapp.runtimeApiOrigin` (defaults to empty -> existing behavior).
- `hosting/docker/webapp/docker-compose.yml` + `hosting/docker/.env.example`: same opt-in for docker self-host.

---

## Testing

Validated on our self-hosted v4 staging cluster (kgateway + Envoy tracing).

**Before** (runners going through public URL, gateway rewrites `traceparent`):
- Parent runner SDK `triggerAndWait()` wrapper spanId: `070bcfdd63b42d2a` (confirmed via wire-level `undici:request:headers` debug)
- Webapp receives `traceparent` with spanId: `b8298ebb884ade7e` (gateway-rewritten)
- Child `TaskRun.parentSpanId = b8298ebb...` (orphan: never in event store)

**After** (with `apiOrigin` pointed at the in-cluster service):
- Parent SDK injects spanId `66fa71fda94ccdb9`
- Webapp receives spanId `66fa71fda94ccdb9` unchanged
- Child `TaskRun.parentSpanId = 66fa71fda94ccdb9` matches the parent's `triggerAndWait()` event in `TaskEvent` table
- Run-detail tree renders the child nested under the parent, matching SaaS behavior

End-to-end test task at https://github.com/meistrari/trigger-self-tests/blob/main/src/tasks/link-test.ts (parent does `linkTestChild.triggerAndWait(...)`).

---

## Changelog

- `webapp`: add optional `RUNTIME_API_ORIGIN` env to advertise a runner-only API origin separate from `API_ORIGIN`. Lets self-hosted operators route runner-to-webapp traffic cluster-internally, bypassing tracing-enabled gateways that rewrite the W3C `traceparent` header on egress and break parent-to-child run linkage in the trace tree. Optional and backward-compatible (falls back to existing `API_ORIGIN`/`APP_ORIGIN`).
- `hosting/k8s/helm` + `hosting/docker`: expose the new env via `webapp.runtimeApiOrigin` (Helm) and `RUNTIME_API_ORIGIN` (docker-compose).

---

## Screenshots

_N/A — no UI change. The visible effect is the run-detail tree correctly showing child task runs nested under their parent's `triggerAndWait()` span (matching trigger.dev SaaS behavior)._

💯

### CODE REVIEW

File: apps/webapp/app/v3/environmentVariables/environmentVariablesRepository.server.ts [hunk 1]
```diff
@@ -1143,11 +1173,11 @@ async function resolveBuiltInProdVariables(
     },
     {
       key: "TRIGGER_API_URL",
-      value: env.API_ORIGIN ?? env.APP_ORIGIN,
+      value: env.RUNTIME_API_ORIGIN ?? env.API_ORIGIN ?? env.APP_ORIGIN,
     },
     {
       key: "TRIGGER_STREAM_URL",
-      value: env.STREAM_ORIGIN ?? env.API_ORIGIN ?? env.APP_ORIGIN,
+      value: env.RUNTIME_API_ORIGIN ?? env.STREAM_ORIGIN ?? env.API_ORIGIN ?? env.APP_ORIGIN,
```
yashs33244, line 1187: Here `RUNTIME_API_ORIGIN` wins over an explicitly-set `STREAM_ORIGIN` for managed runners, making them mutually exclusive. Looks deliberate (and the reorder would re-expose the stream path to the gateway this bypasses), but a one-line comment on why stream follows the runner origin first would make the precedence easy to accept at a glance.

yashs33244 (commented):
Nice fix - I've hit the gateway-rewriting-traceparent problem on a self-hosted setup too, and routing the runner API base via an in-cluster origin is the right lever (the SDK reads `TRIGGER_API_URL` as its client base in `apiClientManager`, so this lands exactly on the `triggerAndWait` request carrying the traceparent). I like that it's optional and backward-compatible, the `(v) => v || undefined` transform neutralizes the empty `${RUNTIME_API_ORIGIN:-}` passthrough cleanly, and leaving dev/CLI runs on the existing chain is the correct call. The before/after span evidence in the description is great.

Two small things and one question:

- The `TRIGGER_STREAM_URL` precedence puts `RUNTIME_API_ORIGIN` ahead of `STREAM_ORIGIN`, so an operator who wants an in-cluster runner-API hop *and* a separate stream endpoint can't have both for managed runners. I see your `.env.example` note already documents the "keep it unset, use STREAM_ORIGIN" workaround, so this looks intentional - could you add a one-line note in the PR body / code on why stream follows the runner origin first? Worth being explicit since the obvious reorder (`STREAM_ORIGIN ?? RUNTIME_API_ORIGIN`) would re-expose the stream path to the same gateway, so the current order is defensible - just non-obvious.

- Might be worth a `RUNTIME_API_ORIGIN` row in `docs/self-hosting/env/webapp.mdx` (the Domains & ports table) - that's where self-hosters will look for it, and it's the one surface the PR doesn't touch yet.

Optional: a small unit test on `resolveBuiltInProdVariables` asserting the new precedence + the empty-string-falls-through behavior. Totally understand if you skip it given there's no existing harness for that path and you've verified end-to-end on a real cluster.

Either way this is a clean, well-scoped change - happy to see it.

ThullyoCunha:
Thanks for the thorough read @yashs33244. Addressed your two points in the latest push:

- Stream precedence rationale: added an explicit note in the resolveBuiltInProdVariables doc-comment spelling out why TRIGGER_STREAM_URL prefers RUNTIME_API_ORIGIN ahead of STREAM_ORIGIN — the obvious reorder (STREAM_ORIGIN ?? RUNTIME_API_ORIGIN) would route the stream path back through the same gateway and re-expose it to the rewrite. Documented the trade-off too: an operator wanting both an in-cluster runner hop and a separate external stream endpoint keeps RUNTIME_API_ORIGIN unset and uses STREAM_ORIGIN.

- Docs: added a RUNTIME_API_ORIGIN row to the Domains & ports table in docs/self-hosting/env/webapp.mdx.

On the unit test: happy to add one, but there's no existing harness for resolveBuiltInProdVariables and the behavior is verified end-to-end on a real cluster — leaving it out for now unless a maintainer prefers otherwise.

The branch is rebased on current main and green. @ericallam @matt-aitken — would love a maintainer pass whenever you have a moment; this is backward-compatible (no-op unless RUNTIME_API_ORIGIN is set) and unblocks self-hosters behind tracing gateways (#2821).

Issue #3519, opened 2026-05-04
bug: TypeError in Replication Service (v4.4.5)


Mathis001:
### Provide environment information

PS C:\Users\Chas\Documents\Github\AiAutomations\automations> npx envinfo --system --binaries

  System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 8.56 GB / 24.00 GB
  Binaries:
    Node: 24.11.1 - C:\Program Files\nodejs\node.EXE
    npm: 11.6.2 - C:\Program Files\nodejs\npm.CMD

### Describe the bug

# [BUG] TypeError: Cannot read properties of undefined (reading 'getTime') in ClickHouse Replication

### Description
This is a server-side bug within the self-hosted Docker infrastructure, specifically in the background replication service that moves data from PostgreSQL CDC (via logical replication) to ClickHouse. As such, it cannot be reproduced in a client-side environment like Stackblitz or CodeSandbox.

However, the issue is directly verifiable by inspecting the toSessionInsertArray and toTaskRunInsertArray utility functions in the apps/webapp source. The stack trace indicates the failure occurs here:

This appears to be a new bug in the data transformation logic, distinct from the permission issues discussed in #2726.

### Environment Details
- **Docker Image**: `ghcr.io/triggerdotdev/trigger.dev:v4.4.5` (also observed in `v4-beta`)
- **Deployment**: Docker Compose
- **PostgreSQL**: 16.2
- **ClickHouse**: 24.3 (LTS)
- **Relevant Env Vars**:
    - `RUN_REPLICATION_ENABLED=1`
    - `EVENT_REPOSITORY_DEFAULT_STORE=clickhouse`

### Error Logs
```text
{"error":{"message":"Cannot read properties of undefined (reading 'getTime')","stack":"TypeError: Cannot read properties of undefined (reading 'getTime')\n    at toSessionInsertArray (/triggerdotdev/apps/webapp/build/index.js:103386:23)\n    at /triggerdotdev/apps/webapp/build/index.js:103261:48\n    at Array.map (<anonymous>)\n    at /triggerdotdev/apps/webapp/build/index.js:103261:34\n    at /triggerdotdev/apps/webapp/build/index.js:12461:20\n    ...
```

### Reproduction repo

N/A

### Steps to Reproduce
1. Deploy Trigger.dev v4.4.5 with ClickHouse replication enabled.
[...]

PR #3697, opened 2026-05-22
fix: null-safe getTime() calls in replication services (#3519)

deepshekhardas:
Adds null-safe access to getTime() calls on createdAt/updatedAt in runs/sessions replication services. Fixes crash when CDC sends rows before timestamps are populated.


deepshekhardas:
Hi team, this fixes #3519 — replication service crash when CDC sends rows before timestamps are populated. Would appreciate a review and merge. Thanks!

Issue #3510, opened 2026-05-02
bug: runs.subscribeToBatch, runs.subscribeToRun sometimes throwing zod schema validation errors


argjent-serus:
### Provide environment information

  System:
    OS: macOS 26.3.1
    CPU: (12) arm64 Apple M4 Pro
    Memory: 1.69 GB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.0 - /Users/argjentsahiti/.nvm/versions/node/v22.13.0/bin/node
    npm: 10.9.2 - /Users/argjentsahiti/.nvm/versions/node/v22.13.0/bin/npm
    pnpm: 9.15.0 - /Users/argjentsahiti/.nvm/versions/node/v22.13.0/bin/pnpm
    bun: 1.2.19 - /Users/argjentsahiti/.bun/bin/bun

### Describe the bug

When using runs.subscribeToRun, and more frequently runs.subscribeToBatch, the SDK sometimes throws the following parsing error:

```
Error: Unable to parse shape: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "taskIdentifier"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "friendlyId"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [
      "createdAt"
    ],
    "message": "Required"
  }
]

```

The issue happens intermittently, but we have seen it occur across development, staging, and production environments.
We do not currently have reliable reproduction steps because the issue only happens sometimes. It appears to be intermittent and does not occur consistently for every run or batch.

Example run IDs where this occurred:
- run_cmolulohwagg00hmvhd9k8vw9
- run_cmoloagju75he0imxokkf9yyn

### Reproduction repo

.

### To reproduce

I’m not sure how to reproduce this consistently because the issue only happens sometimes. I haven’t found a reliable pattern yet. It appears to be intermittent, and I’m reporting it because I’ve seen it happen more than once.

Prateet-Github:
Hi! I’d like to take a look at this.

Since the issue is intermittent, I’ll start by tracing where the subscription payload is emitted and checking if partial data is being streamed before all required fields are populated.

it might be related to timing/race conditions during run initialization, but I’ll try to confirm and report back with findings.

[...]

PR #3699, opened 2026-05-22
fix: subscribeToBatch Zod stream crash on partial shape updates (#3510)

deepshekhardas:
When ElectricSQL sends partial shape updates during initial sync (where not all columns are present in every message), zodShapeStream would crash the entire ReadableStream by calling controller.error(). This killed the subscription for all downstream consumers.

Fixes the issue by logging a warning and skipping unparseable chunks instead of terminating the stream. The RowState in ReadableShapeStream correctly accumulates full row data across insert/update messages and only emits when up-to-date, but edge cases during reconnection or initial sync could emit incomplete rows.

### DIFF
--- a/packages/core/src/v3/apiClient/stream.ts
+++ b/packages/core/src/v3/apiClient/stream.ts
@@ -62,8 +62,10 @@ export function zodShapeStream<TShapeSchema extends z.ZodTypeAny>(
 
         if (result.success) {
           controller.enqueue(result.data);
-        } else {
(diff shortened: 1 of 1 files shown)


deepshekhardas:
Hi team, this fixes #3510 — subscribeToBatch/subscribeToRun stream crashes when ElectricSQL sends partial shape updates during initial sync. Stream now skips unparseable chunks instead of terminating. Would appreciate a review and merge. Thanks!

PR #3709, opened 2026-05-22
feat(webapp,run-engine,redis-worker): mollifier — burst-buffer to absorb trigger storms

d-cs:
## Summary

The mollifier sits in front of `engine.trigger` and diverts trigger storms into a Redis buffer when the per-env trigger rate exceeds a configurable threshold. A background drainer materialises buffered entries back into Postgres at a controlled rate. The customer's `trigger()` call still returns a valid `runId` synchronously; the run materialises in PG within a sub-second window in healthy operation.

The gate is a Lua sliding-window rate check (per-env). When tripped, it holds for `TRIGGER_MOLLIFIER_HOLD_MS` and every trigger during the hold goes to the buffer. The drainer is a fair round-robin over orgs → envs with configurable concurrency. Every read and mutate API that touches a run (retrieve, cancel, replay, reschedule, addTags, updateMetadata, realtime stream, etc.) gains a buffered-source branch so the buffered window is invisible to the customer — same response shape, same Zod schema, same dashboard rendering.

Default: `TRIGGER_MOLLIFIER_ENABLED=0`. Off completely unless explicitly opted in.

## Design

**The gate** (`apps/webapp/app/v3/mollifier/mollifierGate.server.ts`): per-env sliding window, Lua-atomic INCR + tripped-flag SETEX. Configured by `TRIGGER_MOLLIFIER_TRIP_WINDOW_MS`, `TRIGGER_MOLLIFIER_TRIP_THRESHOLD`, `TRIGGER_MOLLIFIER_HOLD_MS`. Globally short-circuited by `TRIGGER_MOLLIFIER_ENABLED` before the `evaluateGate` call so a non-mollifier deployment pays nothing per trigger.

**The buffer** (`packages/redis-worker/src/mollifier/buffer.ts`): Redis ZSET per env (`mollifier:queue:<envId>`) keyed by `createdAtMicros`, plus a per-run hash (`mollifier:entries:<runId>`) carrying the snapshot. Idempotency-lookup keys provide trigger-time dedup symmetric with PG's unique constraint. Entries persist until the drainer ACKs (with a 30s post-materialise grace) or FAILs them — no accept-time TTL, since silent eviction would lose runs without a customer-visible signal.

**The drainer** (`apps/webapp/app/v3/mollifier/mollifierDrainer.server.ts` + `@trigger.dev/redis-worker`'s `MollifierDrainer`): polls the buffer, replays each entry through `engine.trigger`. Handles the cancel-before-PG bifurcation via `engine.createCancelledRun` (no engine.trigger needed when a buffered cancel landed first). On non-retryable engine errors, writes a `SYSTEM_FAILURE` PG row via `engine.createFailedTaskRun` so customers see the failure in their dashboard — and `createFailedTaskRun` now emits `runFailed` so the alert pipeline picks the row up.

**Read-side fallbacks**: every public API route that retrieves a run gains a buffered-source branch. The realtime stream specifically holds the Electric SQL subscription open across the buffered window so `useRealtimeRun` doesn't see a 404 and bail — when the drainer materialises the PG row, Electric streams the INSERT to the client.

**The run span lands in the event store at trigger time**, before the gate divert decision. Buffered runs are visible in the trace view immediately rather than only after drain — important for `useRealtimeRun` and for trigger-and-wait flows where a parent's trace expects the child's span to exist.

**Observability**: `mollifier.decisions{outcome}` counter for gate decisions. `mollifier.stale_entries.current{envId}` gauge for entries sitting longer than `TRIGGER_MOLLIFIER_STALE_SWEEP_THRESHOLD_MS` (default 5min) — the alertable signal for an offline / falling-behind drainer. `mollifier.realtime_subscriptions.buffered{envId}` counter for visibility into customers hitting the buffered window with the realtime hook. All emit via the standard OTel meter pipeline.

**Customer-visible surface**: the API response from a mollified trigger carries a `mollifier.queued` notice. Otherwise the buffered window is invisible — same response shapes, same dashboard, same SDK behaviour. The list APIs (and dashboard runs list) are eventually consistent and do **not** surface buffered rows; that visibility is parked for a future global status-bar UX piece.

## Test plan

- [x] Per-feature unit + container tests across every commit (`apps/webapp/test/mollifier*.test.ts`, `internal-packages/run-engine/src/engine/tests/createFailedTaskRun.test.ts`, `packages/redis-worker/src/mollifier/buffer.test.ts`).
- [x] SDK response shape audit against a buffered run — exercises nine `apiClient` methods so zodfetch schemas validate against the buffered responses end-to-end. No drift surfaced.
- [x] Empirical probes across the realtime, cancel-before-PG, and drainer-cancel-bifurcation flows during development. The probes are not committed; the regressions they caught are pinned by the unit/container tests above.
- [ ] Reviewer: run the local dev stack with `TRIGGER_MOLLIFIER_ENABLED=1`, trigger a burst that exceeds the configured threshold, confirm the dashboard list + run-detail flow matches the non-buffered experience.


d-cs:
Closing as I've created split PR's for easier review

PR #3752, opened 2026-05-26
feat(redis-worker,webapp): mollifier buffer extensions + snapshot type

d-cs:
## Summary

Buffer-side data layer used by the rest of the mollifier phase-3 stack.

- `buffer.ts` gains entry inspection (`getEntry`), idempotency lookup (`lookupIdempotency`), in-place snapshot mutation (`mutateSnapshot`), and dwell tracking. All atomic via Lua.
- `mollifierSnapshot.server.ts`: shared `MollifierSnapshot` type plus (de)serialise helpers.
- Drops the entry-TTL config and its env var. The drainer is the recovery mechanism; an entry that survives the drainer should surface as a stale-sweep alert, not silently TTL away.

Adds methods to the buffer interface; nothing consumes them yet. Subsequent PRs in the stack wire trigger-time mollify, read-fallback, and mutation paths against this surface.

## Test plan

- [x] \`pnpm run typecheck --filter webapp\` passes
- [x] \`pnpm run test --filter @trigger.dev/redis-worker packages/redis-worker/src/mollifier/buffer.test.ts\` passes

### DIFF
--- a/packages/redis-worker/src/mollifier/buffer.ts
+++ b/packages/redis-worker/src/mollifier/buffer.ts
@@ -135,12 +135,10 @@ export class MollifierBuffer {
     const orgsKey = "mollifier:orgs";
     const nowMs = Date.now();
     const createdAt = new Date(nowMs).toISOString();
-    // Microsecond epoch. JS only has millisecond precision, so multiple
-    // accepts in the same ms share a score; ZSET ties resolve by member
-    // (runId) lex order, which is deterministic and acceptable for FIFO
-    // pop. The hash carries the same value as `createdAtMicros` so the
-    // listing helper (Phase E) can read a stable per-run timestamp
-    // without re-fetching the score.
+    // Microsecond epoch, stored as a hash field for dwell-time metrics
+    // (stale sweep, drainer dwell span). FIFO ordering comes from the
+    // LIST itself (LPUSH head / RPOP tail), not from this value — it is
+    // no longer a queue sort key.
     const createdAtMicros = nowMs * 1000;
     const idempotencyLookupKey =
       input.idempotencyKey && input.taskIdentifier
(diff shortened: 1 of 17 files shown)


packages/redis-worker/src/mollifier/buffer.ts
```diff
@@ -10,17 +10,66 @@ import { BufferEntry, BufferEntrySchema } from "./schemas.js";
...
+
+export type IdempotencyLookupInput = {
+  envId: string;
+  taskIdentifier: string;
+  idempotencyKey: string;
+};
+
+function makeIdempotencyLookupKey(input: IdempotencyLookupInput): string {
+  return `mollifier:idempotency:${input.envId}:${input.taskIdentifier}:${input.idempotencyKey}`;
+}
```
d-cs, line 80: Mirrored from CodeRabbit on the original PR (#3709, discussion r3290070699):

---

_⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_

**Encode composite key segments to prevent idempotency key alias collisions.**

Line 47 and Line 60 build Redis keys by raw `:` concatenation. If `taskIdentifier` and/or `idempotencyKey` contain `:`, different tuples can map to the same Redis key and dedupe the wrong run.

 

<details>
<summary>💡 Suggested fix</summary>

```diff
+function encodeKeyPart(value: string): string {
+  return Buffer.from(value, "utf8").toString("base64url");
+}
+
 function makeIdempotencyLookupKey(input: IdempotencyLookupInput): string {
-  return `mollifier:idempotency:${input.envId}:${input.taskIdentifier}:${input.idempotencyKey}`;
+  return `mollifier:idempotency:${encodeKeyPart(input.envId)}:${encodeKeyPart(input.taskIdentifier)}:${encodeKeyPart(input.idempotencyKey)}`;
 }
@@
 function makeIdempotencyClaimKey(input: IdempotencyLookupInput): string {
-  return `mollifier:claim:${input.envId}:${input.taskIdentifier}:${input.idempotencyKey}`;
+  return `mollifier:claim:${encodeKeyPart(input.envId)}:${encodeKeyPart(input.taskIdentifier)}:${encodeKeyPart(input.idempotencyKey)}`;
 }
```
</details>


Also applies to: 59-61

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@packages/redis-worker/src/mollifier/buffer.ts` around lines 46 - 48, The
Redis key builders (e.g., makeIdempotencyLookupKey and other key-construction
sites in this file that concatenate segments with ':') are vulnerable to
collisions when segments contain ':'; update these functions to encode each
variable segment (taskIdentifier, idempotencyKey, etc.) before concatenation
(use a safe reversible encoding such as encodeURIComponent or base64url) so the
colon separators remain unambiguous, and decode where needed when reading values
back; ensure you apply the same encoding consistently across all key-builder
functions in this file.
```

</details>

packages/redis-worker/src/mollifier/buffer.ts
```diff
@@ -10,17 +10,66 @@ import { BufferEntry, BufferEntrySchema } from "./schemas.js";
...
+// authoritative cross-store "this idempotency key is in flight or
+// resolved" pointer used by the trigger hot path
+// (`_plans/2026-05-21-mollifier-idempotency-claim.md`). Values:
+//   "pending"   → a trigger pipeline owns the key and hasn't published yet
+//   <runId>     → the winning trigger's runId (resolved)
+export const IDEMPOTENCY_CLAIM_PENDING = "pending";
+
+function makeIdempotencyClaimKey(input: IdempotencyLookupInput): string {
+  return `mollifier:claim:${input.envId}:${input.taskIdentifier}:${input.idempotencyKey}`;
+}
```
d-cs, line 98: Mirrored from CodeRabbit on the original PR (#3709, discussion r3290070707):

---

_⚠️ Potential issue_ | _🟠 Major_ | _🏗️ Heavy lift_

**Claim ownership is not protected; late release can erase another claimant’s state.**

Line 57 uses a shared `"pending"` marker and Line 403 performs unconditional `DEL`. If ownership changes after TTL expiry, an old caller can delete a new resolved claim, reopening duplicate execution paths.

 


Also applies to: 359-387, 399-404, 719-737

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@packages/redis-worker/src/mollifier/buffer.ts` around lines 57 - 61, The code
uses a shared constant IDEMPOTENCY_CLAIM_PENDING and performs unconditional DEL
which allows a stale claimant to delete a new owner’s claim; change the claim
model to generate a unique per-claim ownership token (e.g., random UUID) when
creating the claim (use makeIdempotencyClaimKey for the key and store the token
as the value), return/store that token with the claim owner, and replace
unconditional DEL with a conditional release that only deletes if the stored
value matches the owner’s token (implement via a Redis EVAL/Lua script or Redis
GET+WATCH/MULTI to compare-and-delete). Update all claim acquire/release call
sites to set the token on claim creation and pass it to the conditional release
so only the rightful owner can remove its claim.
```

</details>

packages/redis-worker/src/mollifier/buffer.ts
d-cs: Addressed in 02cfe1af1. Added an `encodeKeyPart` helper (base64url — no `:` in the alphabet) and applied it to both `idempotencyLookupKeyFor` and `makeIdempotencyClaimKey` so a customer-supplied `taskIdentifier` or `idempotencyKey` containing `:` can no longer alias another tuple onto the same Redis key. Coverage: `MollifierBuffer composite-key encoding (collision resistance)` describe block in `buffer.test.ts` — pins both the alias case (two accepts whose unencoded keys would collide both succeed and resolve to their own runIds) and the encoded-key shape (no `:` beyond the namespace separators).

packages/redis-worker/src/mollifier/buffer.ts
d-cs: Addressed in 02cfe1af1. Replaced the shared `"pending"` marker with a caller-supplied ownership token stored as `"pending:<token>"`. `publishClaim` and `releaseClaim` are now Lua compare-and-act (compare-and-set / compare-and-delete) against that marker, so a late release from a previous claimant whose TTL expired is a no-op rather than a stomp on a new owner's slot. The `readClaim` parser checks the `pending:` prefix to distinguish pending vs resolved without leaking the token to losers.

Token wiring is threaded through the trigger pipeline in #3753 (`ClaimOrAwaitOutcome.claimed` carries it, `ClaimedIdempotency` carries it, `triggerTask` passes it to publish/release).

Coverage in `buffer.test.ts`:
- claim/pending visibility (token-A vs token-B)
- wrong-token release is a no-op (owner's claim survives)
- wrong-token publish returns `false` and the slot stays pending
- explicit regression test for the hazard you described: claimant A claims with token A, slot TTLs out (simulated DEL), claimant B claims with token B, A's late release MUST be a no-op, B's claim survives intact and B can still publish.

packages/redis-worker/src/mollifier/schemas.ts
```diff
@@ -44,6 +48,24 @@ export const BufferEntrySchema = z.object({
   status: BufferEntryStatus,
   attempts: stringToInt,
   createdAt: stringToDate,
+  // Microsecond epoch matching the ZSET queue score. Stable across
+  // requeues — the score never moves once set at accept time.
+  createdAtMicros: stringToInt,
```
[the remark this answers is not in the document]
d-cs: Good catch — fixed in `decea6bec`. `createdAtMicros` now defaults to `"0"` like the other fields this PR added (`materialised`, `idempotencyLookupKey`, `metadataVersion`), so an entry written by an accept Lua predating the field (or surviving across the deploy that introduced it) still parses on pop instead of being silently dropped. Note that after the queue was reverted from a ZSET to a FIFO LIST, `createdAtMicros` is no longer a sort key — it's only a hash field for dwell metrics — so the default is purely about backward-compatible parsing. Added a schema test covering an entry that omits the field.

packages/redis-worker/src/mollifier/buffer.ts
```diff
@@ -10,17 +10,85 @@ import { BufferEntry, BufferEntrySchema } from "./schemas.js";
 
 export type MollifierBufferOptions = {
   redisOptions: RedisOptions;
-  entryTtlSeconds: number;
   logger?: Logger;
 };
```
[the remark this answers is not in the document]
d-cs: Already resolved on the current tip — all three `MollifierBuffer` constructions in `apps/webapp/test/mollifierTripEvaluator.test.ts` (lines 17/35/76) are `new MollifierBuffer({ redisOptions })` with no `entryTtlSeconds`, so there's no excess-property type error. This review looked at an earlier SHA. Resolving.

d-cs:
Addressed the CodeRabbit findings in `decea6bec`:

**Outside-diff (latest review):**
- **`lookupIdempotency` unconditional DEL → compare-and-delete.** The stale-lookup self-heal now clears the key only if it still holds the `runId` we observed, via a small `delMollifierKeyIfEquals` Lua. A concurrent `accept` that rebinds the key between our GET and DEL can no longer be clobbered.
- **`accept` returned a stale `existingRunId` without checking the entry exists.** `acceptMollifierEntry` now `EXISTS`-checks `mollifier:entries:<existing>` before honouring the binding; if the entry was evicted but the lookup survived, it rebinds to the new run instead of blocking the key indefinitely. Mirrors the `lookupIdempotency` self-heal. Covered by two new tests (rebind-on-eviction, and dedup still wins when the entry is live).
- **Requeue test naming nit.** Renamed to "requeued entry gets retry priority …"; the comment now states RPUSH-to-tail gives retry priority (pops ahead of newer items), explicitly not FIFO relative to the rest of the queue.

**Earlier reviews:**
- **`resetIdempotency` left the claim key behind** — already fixed earlier in this PR: `resetMollifierIdempotency` now `DEL`s the `mollifier:claim:*` slot alongside the lookup (test: "resetIdempotency also clears the pre-gate claim slot").
- **`listEntriesForEnv` parallel-fetch nit** — moot: `listForEnvWithWatermark` (the method it was asked to be consistent with) was removed, since the buffer no longer renders a paginated list view. `listEntriesForEnv` is now only used by the stale sweep with a bounded `maxCount`, so sequential fetch is fine.

The two Devin threads (createdAtMicros default, entryTtlSeconds) are resolved inline.

ericallam (approved):

ericallam (approved):

PR #3776, opened 2026-05-29
feat(webapp): runs live updating

kathiekiwi:
## Summary

The Runs list now updates live without requiring a page refresh. Status changes and other run fields are updated in place while runs are executing.

When new runs matching the current filters are created, a "New runs created" refresh button appears above the list. 

Root runs now show a live child-run status breakdown directly in the status tooltip.

### List live update

- Visible runs update in place while they are still running.
- A "New runs created" refresh button appears when new matching runs are detected.
- Polling stops when all visible runs have finished and a refresh button is already shown.
- Polling pauses when the browser tab is not visible.
- Runs list status updates and new-run detection share a single runs/live polling path.

### Child-status tooltip

- Root run tooltips now display a breakdown of child run statuses.
- Child statuses are loaded when the tooltip opens (after a 400ms hover delay).
- The tooltip stays up to date while child runs are still changing state.
- Handles cases where child runs continue running after their parent run has completed, or have not yet been created.

### Supporting changes

- Added hidden-tab awareness to polling.
- Added safeguards around polling inputs (`runIds` deduping and limits).

## Test plan

- [x] pnpm run typecheck --filter webapp passes
- [x] cd apps/webapp && pnpm run test ./test/presenters/mapRunToLiveFields.test.ts --run passes
- [x] cd apps/webapp && pnpm run test ./test/runsRepository.part2.test.ts --run -t "hasNewRuns" passes

### Manual smoke:

- [x] Active runs update without a page refresh.
- [x] A new matching run shows the refresh banner and the banner actions work as expected.
- [x] Root run tooltips show live child-status updates and stop polling once child runs settle.

### DIFF
--- a/apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.children-statuses.ts
+++ b/apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.children-statuses.ts
@@ -1,0 +1,110 @@
+import { type LoaderFunctionArgs } from "@remix-run/server-runtime";
+import { type TaskRunStatus } from "@trigger.dev/database";
+import { z } from "zod";
+import { $replica } from "~/db.server";
+import { clickhouseFactory } from "~/services/clickhouse/clickhouseFactoryInstance.server";
+import { loadProjectEnvironmentFromRequest } from "~/services/loadProjectEnvironmentFromRequest.server";
+import { RunsRepository } from "~/services/runsRepository/runsRepository.server";
+import { runIdsQueryParam } from "~/utils/searchParams";
+
+const SearchParamsSchema = z.object({
+  runIds: runIdsQueryParam,
+});
+
+const ROOT_CREATED_AT_SAFETY_MARGIN_MS = 5 * 60 * 1000;
+
+type RootRun = { id: string; friendlyId: string; createdAt: Date };
+type ChildStatusEntry = { status: TaskRunStatus; count: number };
+type GroupedChildStatus = {
+  rootRunId: string;
+  status: TaskRunStatus;
+  count: number;
+};
+
+function mapGroupedStatusesToFriendlyIds(
+  grouped: GroupedChildStatus[],
+  roots: RootRun[]
+): Map<string, ChildStatusEntry[]> {
+  const rootFriendlyIdById = new Map(roots.map((run) => [run.id, run.friendlyId]));
+  const statusesByFriendlyId = new Map<string, ChildStatusEntry[]>();
+
+  for (const item of grouped) {
+    const friendlyId = rootFriendlyIdById.get(item.rootRunId);
+    if (!friendlyId) continue;
+
+    const existing = statusesByFriendlyId.get(friendlyId) ?? [];
+    existing.push({
+      status: item.status,
+      count: item.count,
+    });
+    statusesByFriendlyId.set(friendlyId, existing);
+  }
+
+  return statusesByFriendlyId;
+}
+
+function childrenStatusesResponseForRunIds(
+  runIds: string[],
+  statusesByFriendlyId: Map<string, ChildStatusEntry[]>
+) {
+  return {
+    runs: runIds.map((friendlyId) => ({
+      friendlyId,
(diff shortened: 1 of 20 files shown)


apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.live.ts
```diff
@@ -0,0 +1,108 @@
...
+              baseCostInCents: true,
+            },
+            where: {
+              projectId: project.id,
+              runtimeEnvironmentId: environment.id,
+              friendlyId: {
+                in: runIds,
+              },
+            },
+          })
```
matt-aitken: I think we should use the RunsRepository for this. Replace any Postgres `taskRun` findMany queries with the run repository instead. That means it will use ClickHouse in production for us.

You should be able to use `runsRepository.listRuns()` and pass an array in for `runId`. You'll need to pass the org, project, and environment ids too.

apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.children-statuses.ts
```diff
@@ -0,0 +1,115 @@
...
+      createdAt: true,
+    },
+    where: {
+      projectId: project.id,
+      runtimeEnvironmentId: environment.id,
+      friendlyId: {
+        in: runIds,
+      },
+    },
+  });
```
matt-aitken: I think we should use the RunsRepository for this. Replace any Postgres taskRun findMany queries with the run repository instead. That means it will use ClickHouse in production for us.

You should be able to use runsRepository.listRuns() and pass an array in for runId. You'll need to pass the org, project, and environment ids too.

apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.live.ts
```diff
@@ -0,0 +1,108 @@
...
+            from: Math.max(filters.from ?? 0, newRunsSince + 1),
+            to: filters.to,
+            rootOnly: filters.rootOnly,
+            batchId: filters.batchId,
+            runId: filters.runId,
+            bulkId: filters.bulkId,
+            queues: filters.queues,
+            machines: filters.machines,
+            errorId: filters.errorId,
+          });
```
matt-aitken: To get the 99+ you  could use `runsRepository.listRunIds()` with a `size: 100`. If you get 100 items then it's 99+. I'm not worried about the performance of this.

matt-aitken (changes requested):

apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.live.ts
kathiekiwi: Performance should be fine, it's about when we stop polling - 1 will stop immediately, 100 will be hanging for some time

apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.live.ts
kathiekiwi: Applied 👍

apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.live.ts
kathiekiwi: Applied 👍

apps/webapp/app/routes/resources.orgs.$organizationSlug.projects.$projectParam.env.$envParam.runs.children-statuses.ts
kathiekiwi: Applied 👍

matt-aitken (approved):

PR #3829, opened 2026-06-03
Env vars page performance fix

kathiekiwi:
## Summary

This PR improves performance across the Environment Variables page.

### Targeted value loading

- load only the non-secret (environmentId, key) pairs required by the page. Secret values continue to be redacted in the UI.

### SSR windowing + virtualization

- SSR-render only the first 50 rows
- hydrate those rows
- virtualize the remaining dataset client-side
- search is now URL-driven during SSR, ensuring deep links such as `?search=DATABASE_URL`

### Lightweight 'Create' flow

- 'Create' page no longer loads the full Environment Variables dataset.

## Results

Large projects no longer render thousands of rows during SSR.
Example (~11k rendered rows):

Metric | Before | After
-- | -- | --
Document size | ~150 MB | ~5 MB
SSR rows | ~11k | 50
Browser DOM rows | Thousands | ~26–38

## Testing

Automated

- [x] pnpm run test ./test/environmentVariablesRepository.test.ts
- [x] pnpm run test ./test/environmentVariablesEnvironments.test.ts 
- [x] pnpm run test ./test/EnvironmentVariablesPresenter.test.ts

Manual

- [x] Environment Variables list loads without browser freezes
- [x] Search works via URL parameters
- [x] Secret values remain redacted
- [x] Non-secret values render correctly
- [x] Create page loads successfully
- [x] Create/edit/delete flows continue to work
- [x] Virtualized table preserves column alignment and sticky headers


apps/webapp/app/presenters/v3/environmentVariablesEnvironments.server.ts
```diff
@@ -0,0 +1,77 @@
...
+        select: {
+          userId: true,
+        },
+      },
+    },
+    where: {
+      project: {
+        slug: projectSlug,
+      },
+      archivedAt: null,
```
matt-aitken: Will this query be slow? There's an index on projectId we could use if we have the id to use instead.

matt-aitken: I just saw this is how we already do it, I wonder if this is part of why the page is slow on production where we have a ton of environments.

matt-aitken: We should switch to passing the project id into this function and use that here

matt-aitken (changes requested):

apps/webapp/app/presenters/v3/environmentVariablesEnvironments.server.ts
kathiekiwi: done!

matt-aitken (approved):

PR #3830, opened 2026-06-04
fix(webapp): validate packet storage paths

kathiekiwi:
## Summary

Packet-relative paths were used to construct object-store keys and presigned URLs without validation. Crafted paths containing traversal segments (for example `../`) could escape the intended `packets/{projectRef}/{envSlug}/` storage prefix.

This PR adds packet path validation before key construction and presigning. Invalid paths are rejected before reaching either object-store client implementation, ensuring consistent behavior regardless of the underlying storage configuration.

Presign API routes now return `400 Bad Request` for invalid packet paths instead of `500 Internal Server Error`.

### Changes

* Added `assertSafePacketRelativePath()`.
* Rejects empty paths, leading `/`, backslashes (`\`), empty path segments (`foo//bar`), `.` and `..` path segments
* Validation is enforced for packet upload, downloads, presign requests
* Invalid packet paths now raise `ServiceValidationError` with status `400`.
* Packet presign routes return:
  ```json
  { "error": "Invalid packet storage path" }
  ```
  with HTTP `400 Bad Request`.
* Other presign failures continue to return `500 Internal Server Error`.

#### Automated

- [x] * `pnpm run typecheck --filter webapp` passes
- [x] * `cd apps/webapp && pnpm run test ./test/objectStore.test.ts --run` passes

#### Manual smoke

- [x] Valid packet paths (for example `run_123/payload.json`) continue to presign and upload/download normally.
- [x] Presign requests using traversal paths:

  * `../file.json`
  * `../../other-env/file.json`
  * `foo/../bar.json`
  * `/absolute/path.json`

  return `400 Bad Request` with:

  ```json
  { "error": "Invalid packet storage path" }
  ```

- [x]  Server-side upload and download paths containing traversal segments are rejected before object-store key construction.

### DIFF
--- a/.server-changes/validate-packet-storage-paths.md
+++ b/.server-changes/validate-packet-storage-paths.md
@@ -3,16 +3,4 @@ area: webapp
 type: fix
 ---
 
-Validate packet-relative storage paths before building object-store keys or presigned URLs. Rejects:
-
-- an empty path
-- absolute paths (leading `/`)
-- backslashes (`\`)
-- empty path segments (e.g. `foo//bar`, leading or trailing `/`)
-- `.` path segments (e.g. `.`, `foo/./bar`)
-- `..` path segments (path traversal, e.g. `../file`, `foo/../bar`)
-- percent-encoded `.` / `..` segments (e.g. `%2e%2e`, `%2E%2E`, `%2e.`)
-
-After segment checks, paths are normalized with the same URL pathname resolution used by `Aws4FetchClient`, and the full object-store key must remain under `packets/{projectRef}/{envSlug}/` after that normalization.
-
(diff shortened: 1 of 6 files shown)


nicktrn (approved):

lissy93 (commented):
Yay, thanks for fixing!! 🎉

Did you want to update the security report in [GHSA-888c-px7m-736v](https://github.com/triggerdotdev/trigger.dev/security/advisories/GHSA-888c-px7m-736v)? Since these two issues are fixed, the chain is effectively broken 😇
- url.pathname traversal in presign - fixed :)
- unvalidated wildcard in key - fixed :)
- missing per-resource authorization GET loader - not fixed, but not exploitable anymore :)

Issue #3539, opened 2026-05-08
Vouch Request - Lissy93


lissy93:
### Why do you want to contribute?

I found a bug / security issue. I would like to fix it (in a private fork, according to your / the default security policy).

### Prior contributions or relevant experience

I don't have any 😢

But I have been working on similar projects since 2025, contributing to open source since 2015, and coding since 2005.

I'm also a maintainer of several large projects, so you can trust that I am a real human, I will not spam, and I know how to code. My GH is https://github.com/Lissy93

lissy93:
Hey @ericallam - any chance you cold take a look at my vouch request 🙏

For some more context, I specialise in self-hosting, security and open source ecosystems

I would like to be vouched because:
- There's a few small bugs I would like to help fix :)

Relevant experience:
- OpenUK honours list, [2024](https://openuk.uk/community/openuk-honours-lists/2024-honours-list/) and [2025](https://openuk.uk/honours/)
- GitHub AI Accelerator member, [source](https://github.com/open-source/accelerator)
- Maintainer of several large open source projects, [source](https://gitstar-ranking.com/Lissy93). including:
  - Dashy, 10M Docker pulls, 180 contributors, 25k stars, 9 years of maintenance 
  - Web-Check, 10k daily users, 33k stars, profitable for 3 years

PR #3842, opened 2026-06-05
fix(webapp): scope team-member removal to the caller's org

lissy93:
Closes #nothing

## ✅ Checklist

---

## Testing
- `pnpm run typecheck --filter webapp` passes
- Traced the only caller (`_app.orgs.$organizationSlug.settings.team/route.tsx`): the rendered `memberId` always comes from the org-scoped loader, so every legitimate request (self-leave or removing a same-org member) still matches and succeeds
- A cross-org `memberId` now resolves to no record; the delete throws P2025, which the route already catches and returns as a 400, so no crash

---

## Changelog

Scope `removeTeamMember` deletes to the caller's organization, fixing a cross-org member-deletion IDOR

Previously, a member of any org could delete a row from any other org, if they knew their id. And they do know their ID, since `OrgMember.id` is rendered into a hidden form input on the team-settings page.

Note that I previously opened this as a private PR about a month ago, [here](https://github.com/triggerdotdev/trigger.dev-ghsa-58mc-m3qq-6qrh/pull/1). But I think this is safe to open publicly, since it doesn't reveal the other steps needed to exploit the full account take over (which I documented in [GHSA-58mc-m3qq-6qrh](github.com/triggerdotdev/trigger.dev/security/advisories/GHSA-58mc-m3qq-6qrh).

I submitted a vouch request in #3539 - could you take a look please?

---

## Screenshots

N/A — server-side authorization fix

💯

### DIFF
--- a/.server-changes/remove-team-member-org-scope.md
+++ b/.server-changes/remove-team-member-org-scope.md
@@ -1,0 +1,6 @@
+---
+area: webapp
+type: fix
+---
+
+Scope the `removeTeamMember` delete to the resolved organization so a member can only be deleted from the org they belong to. Previously the delete was keyed by `OrgMember.id` alone, letting a privileged caller in one org delete members of an unrelated org by id (cross-org IDOR).
--- a/apps/webapp/app/models/member.server.ts
+++ b/apps/webapp/app/models/member.server.ts
@@ -76,6 +76,7 @@ export async function removeTeamMember({
   return prisma.orgMember.delete({
     where: {
       id: memberId,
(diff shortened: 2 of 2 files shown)


lissy93:
> This project requires that pull request authors are vouched, and you are not in the list of vouched users.
> This PR will be closed automatically. See https://github.com/triggerdotdev/trigger.dev/blob/main/CONTRIBUTING.md for more details.

I did, about a month ago, in #3539 - but am still waiting 😭

PR #3848, opened 2026-06-05
fix(webapp): stop API auth failures leaking the controller's raw error string

d-cs:
## Summary

API auth failures returned the auth controller's raw `error` string verbatim in the response body. That string isn't guaranteed to be client-safe — a controller can surface an internal infrastructure error (e.g. a database connection failure) as the auth error, which then reaches the API client and is echoed by the SDK into the run view. This removes that channel: the client message is now derived purely from the HTTP status, and the controller's original string is logged server-side only.

## Design

The fix sits at the single auth chokepoint every authenticated API request funnels through — the apiBuilder bearer bridge and the PAT path. Both now route failures through `sanitizeAuthFailure`, which maps the status to a fixed client-safe message (`401 → "Invalid credentials"`, `403 → "Forbidden"`) and drops the controller's prose. Status codes, response body shape, and machine-readable fields are unchanged; only the human-readable message text changes. Granular auth hints (e.g. preview-branch header guidance) are flattened as a deliberate trade-off — full detail stays in server logs.

This is provenance-based rather than content-based: it holds regardless of what any current or future auth controller returns, so the same class of leak can't reappear through a different message string. It closes a gap left by the earlier 5xx/exception-focused passes in #3536 and #3664, which couldn't reach a value *returned* on a 4xx through the auth channel.

## Test plan

- [x] Unit test: the leaked-message fixture is reduced to the status-derived message
- [x] e2e: bridge route returns the fixed message on auth failure (confirmed RED on pre-fix code, GREEN after)
- [x] Full auth e2e suites pass (`api-auth`, `auth-api`, `auth-cross-cutting`)


ericallam (approved):

d-cs:
Closing as auth reads come from replica - this is unnecessary

PR #3866, opened 2026-06-08
ci: allow forks to override published container image namespace

d-cs:
## Summary

The container publish workflows hardcoded `ghcr.io/triggerdotdev/...` as the image destination. As a result, a fork that builds on push-to-`main` (or on the worker publish tags) would attempt to push to — and attest — the upstream packages rather than its own, which fails on permissions and is surprising besides.

This makes the image destination configurable via a single `IMAGE_REGISTRY` repository variable, while leaving the upstream defaults byte-identical:

- **Single source of truth** (`publish.yml`): a `resolve-registry` job resolves the target registry namespace once — `IMAGE_REGISTRY` repository variable, defaulting to `ghcr.io/${{ github.repository_owner }}` — and passes it down to every publish job as an `image_registry` input. So a fork publishes to its own namespace automatically with no configuration.
- **Webapp** (`publish-webapp.yml`): the image now lives at `<registry>/<repo-name>` (e.g. `ghcr.io/<owner>/trigger.dev`). The provenance attestation and the downstream Trivy scan follow the same computed repo via the `image_repo` workflow output.
- **Workers** (`publish-worker.yml`, `publish-worker-v4.yml`): build under `<registry>/<worker-name>`. They keep a `vars.IMAGE_REGISTRY || ghcr.io/<owner>` fallback so they still resolve correctly on their direct `infra-*` / `re2-*` push triggers (which bypass the parent workflow).

A single `IMAGE_REGISTRY` namespace variable now governs both webapp and workers (the earlier `WEBAPP_IMAGE_REPO` full-path override is dropped, removing the full-path/namespace asymmetry). When `IMAGE_REGISTRY` is unset, every resolved image name is exactly what it is today, so there is no change for this repo.

## Test plan

- [x] `actionlint` passes on all four workflows
- [ ] On merge, confirm the webapp publish still pushes `ghcr.io/triggerdotdev/trigger.dev:main` + the commit-SHA tag (defaults unchanged)


d-cs:
Addressed the asymmetry @devin-ai-integration flagged in 7645e6d02: the worker workflows (`publish-worker.yml`, `publish-worker-v4.yml`) now default the namespace to `format('ghcr.io/{0}', github.repository_owner)` instead of the hardcoded `ghcr.io/triggerdotdev`, mirroring the webapp's fork-friendly default. Upstream resolves byte-identically (owner is `triggerdotdev`), so forks now get zero-config worker publishing too.

.github/workflows/publish.yml
myftija, line 54: Another option would be to resolve the target registry here on the parent workflow and pass it down to the jobs.

myftija (approved):

.github/workflows/publish.yml
d-cs: Done in 271521703 — `publish.yml` now resolves the registry once in a `resolve-registry` job and passes it down to all three publish jobs as an `image_registry` input.

While here I unified on a single `IMAGE_REGISTRY` namespace variable for webapp + workers (the webapp image is now `<registry>/<repo-name>`), which also retires the separate `WEBAPP_IMAGE_REPO` var and the full-path/namespace asymmetry. The child workflows keep the `vars.IMAGE_REGISTRY || ghcr.io/<owner>` fallback so the workers still resolve on their direct `infra-*`/`re2-*` push triggers. Upstream defaults are byte-identical (`ghcr.io/triggerdotdev/...`).

d-cs: Followed up in beb539181 — dropped the dedicated job to skip its runner startup. (A workflow-level `env:` can't be referenced from a reusable workflow's `with:` block, so the resolution expression is inlined directly into each job's `image_registry` input instead.) Same result, no extra runner.

d-cs: Better solution in c564e9d26 — defined the expression once on `publish-webapp` via a YAML anchor (`&image_registry`) and reused it in the worker jobs via aliases (`*image_registry`). Single source of truth, no dedicated resolve job, no runner startup. (TIL GitHub Actions has supported YAML anchors since 2025-09-18 — only merge keys are still unsupported.)

myftija (approved):

PR #3933, opened 2026-06-12
feat(sdk): add triggerConfig support to chat.headStart()

saasjesus:
## Summary

- Adds optional `triggerConfig` to `chat.headStart()` and `chat.openSession()`, threading it through `openHandoverSession` into `createSession`.
- Merges customer trigger options (tags, queue, machine, maxAttempts, maxDuration, region, lockToVersion) using the same assembly pattern as `chat.createStartSessionAction()`, auto-prepending `chat:{chatId}` to tags.
- Top-level `idleTimeoutInSeconds` still wins over `triggerConfig.idleTimeoutInSeconds`; `handover-prepare` trigger and `chatId` in `basePayload` cannot be overridden by customer config.
- Fixes a latent tag-cap bug in `createChatStartSessionAction`: `SessionTriggerConfig.tags` allows at most 5 (per schema), but the preload path was slicing to 10 — now both session entry points cap at 5.

## Why this matters

`openHandoverSession` is the session-entry path on turn 0 for head-start chats. Because `createSession` is idempotent on `(env, externalId = chatId)`, that first call fixes the session's `triggerConfig` for its whole lifetime — a later `createStartSessionAction` cannot retroactively add tags to a session head-start already created.

## Test plan

- [x] `cd packages/trigger-sdk && pnpm test src/v3/chat-server.test.ts src/v3/createStartSessionAction.test.ts --run`
- [x] `chat-server.test.ts`: merged tags (`chat:chat-1` first) and queue on `createSession`
- [x] `createStartSessionAction.test.ts`: tag prepend + cap at 5
- [ ] Manual: `chat.headStart({ triggerConfig: { tags, queue } })` and verify handover-prepare run tags/queue in dashboard

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/chat-headstart-trigger-config.md
+++ b/.changeset/chat-headstart-trigger-config.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/sdk": patch
+---
+
+Add `triggerConfig` support to `chat.headStart()` so handover-prepare runs inherit tags, queue, and other session trigger options like `chat.createStartSessionAction()`.
--- a/packages/trigger-sdk/src/v3/ai.ts
+++ b/packages/trigger-sdk/src/v3/ai.ts
@@ -9710,8 +9710,8 @@ function createChatStartSessionAction<TChat extends AnyTask = AnyTask>(
     // run-list filter by chat works without the customer having to wire it
     // up. Mirrors the browser-mediated `TriggerChatTransport.doStart` path.
     const userTags = params.triggerConfig?.tags ?? options?.triggerConfig?.tags ?? [];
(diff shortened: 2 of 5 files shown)


ericallam:
Thanks for this! We're folding it into #3963, which adds `chat.headStart` handover support for the `customAgent` and `createSession` backends, where this `triggerConfig` threading sits naturally alongside. Your commits are credited as co-author there, and it picks up the review's parity follow-up (`createStartSessionAction` now forwards `maxDuration`/`region`/`lockToVersion` too). Closing in favor of #3963.

PR #3949, opened 2026-06-15
feat(plugins): add SSO plugin contract to @trigger.dev/plugins

0ski:


.changeset/sso-plugin-contract.md
```diff
@@ -0,0 +1,5 @@
+---
+"@trigger.dev/plugins": patch
+---
+
+Add the SSO plugin contract (`SsoController`, `SsoPlugin`, domain types, error unions). Vendor-neutral surface for self-service SSO setup, login routing, and JIT provisioning — the cloud implementation lives outside the package; OSS deployments get a no-op fallback that returns `no_sso` from `decideRouteForEmail`.
```
matt-aitken: This package is now private, so we don't need a changeset for the plugin now.

packages/plugins/src/index.ts
```diff
@@ -18,3 +18,31 @@ export type {
...
+  SsoWebhookError,
+  SsoWebhookEvent,
+} from "./sso.js";
+
+export { SSO_FLOWS } from "./sso.js";
+
+// Convenience re-exports — gives plugin authors (and the cloud workspace
+// link) one import surface without reaching into @trigger.dev/core
+// directly. Both helpers live in core; this is purely a forwarder.
+export { sanitizeBranchName, isValidGitBranchName } from "@trigger.dev/core/v3/utils/gitBranch";
```
matt-aitken: Remove these, it caused issues and we can just use directly from core now

matt-aitken (changes requested):

.changeset/sso-plugin-contract.md
0ski: Nice, ok removed

packages/plugins/src/index.ts
0ski: sounds good, removed

matt-aitken (approved):

PR #4105, opened 2026-07-02
feat(sdk): add bulk replay to api and sdk

carderne:
## Summary

Adds SDK and API support for run bulk actions. You can now create bulk cancel or replay actions from `@trigger.dev/sdk` using run IDs or the same filters as `runs.list()`, then retrieve, list, poll, or abort the action by its `bulk_` handle.

Tests, docs, changesets added.

## Design

The dashboard bulk action service now accepts structured filters instead of reading directly from a dashboard request, so the dashboard and API share the same creation path. Replay actions created through the API are attributed with the existing `api` trigger source, while dashboard-created actions keep `dashboard`.

The SDK exposes the new surface under `runs.bulk.*`, including `targetRegion` for replay region overrides and cursor pagination for listing bulk actions.

## Filters and runIds

Nuance on filters. If `filter` is provided, it MUST have at least one key. This is to remove the footgun of passing no filter and selecting all runs.

```typescript
   { action: "cancel", runIds: ["run_1"] } // valid
   { action: "cancel", runIds: [] } // invalid, min(1)
   { action: "cancel", filter: { status: "FAILED" } } // valid
   { action: "cancel", filter: {} } // invalid
   { action: "cancel", filter: {}, runIds: ["run_1"] } // invalid
```

### DIFF
--- a/apps/webapp/app/routes/api.v1.bulk-actions.$bulkActionId.ts
+++ b/apps/webapp/app/routes/api.v1.bulk-actions.$bulkActionId.ts
@@ -1,0 +1,39 @@
+import { json } from "@remix-run/server-runtime";
+import { z } from "zod";
+import { $replica } from "~/db.server";
+import { ApiBulkActionPresenter } from "~/presenters/v3/ApiBulkActionPresenter.server";
+import { createLoaderApiRoute } from "~/services/routeBuilders/apiBuilder.server";
+
+const ParamsSchema = z.object({
+  bulkActionId: z.string(),
+});
+
+export const loader = createLoaderApiRoute(
+  {
+    params: ParamsSchema,
+    corsStrategy: "none",
+    authorization: {
+      action: "read",
+      resource: () => ({ type: "runs" }),
+    },
+    findResource: async (params, auth) => {
+      return $replica.bulkActionGroup.findFirst({
+        select: { id: true },
+        where: {
+          friendlyId: params.bulkActionId,
+          environmentId: auth.environment.id,
+        },
+      });
+    },
+  },
+  async ({ params, authentication }) => {
+    const presenter = new ApiBulkActionPresenter();
+    const bulkAction = await presenter.retrieve(authentication.environment.id, params.bulkActionId);
+
+    if (!bulkAction) {
+      return json({ error: "Bulk action not found" }, { status: 404 });
+    }
+
+    return json(bulkAction);
+  }
+);
--- a/apps/webapp/app/routes/api.v1.bulk-actions.$bulkActionId.ts
+++ b/apps/webapp/app/routes/api.v1.bulk-actions.$bulkActionId.ts
@@ -1,7 +1,10 @@
 import { json } from "@remix-run/server-runtime";
 import { z } from "zod";
 import { prisma } from "~/db.server";
-import { ApiBulkActionPresenter } from "~/presenters/v3/ApiBulkActionPresenter.server";
+import {
+  apiBulkActionObject,
+  bulkActionSelect,
+} from "~/presenters/v3/ApiBulkActionPresenter.server";
 import { createLoaderApiRoute } from "~/services/routeBuilders/apiBuilder.server";
 
 const ParamsSchema = z.object({
@@ -16,10 +19,10 @@ export const loader = createLoaderApiRoute(
       action: "read",
       resource: () => ({ type: "runs" }),
     },
+    // Read from primary so create -> retrieve/poll doesn't 404 on replica lag.
     findResource: async (params, auth) => {
-      // Read from primary so create -> retrieve/poll doesn't 404 on replica lag.
       return prisma.bulkActionGroup.findFirst({
-        select: { id: true },
+        select: bulkActionSelect,
         where: {
           friendlyId: params.bulkActionId,
           environmentId: auth.environment.id,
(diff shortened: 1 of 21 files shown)


apps/webapp/app/routes/api.v1.bulk-actions.ts
```diff
@@ -0,0 +1,129 @@
...
+        organizationId: authentication.environment.organizationId,
+        projectId: authentication.environment.projectId,
+        environmentId: authentication.environment.id,
+        userId: authentication.actor?.sub ?? null,
+        action: body.action,
+        title: body.name,
+        region: body.targetRegion,
+        filters: body.runIds
+          ? { runId: body.runIds }
+          : bulkActionFilterToRunListFilters(body.filter),
```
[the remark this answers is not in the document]
carderne: fixed in [d789912](https://github.com/triggerdotdev/trigger.dev/pull/4105/commits/d789912c68e80c1029123bf1539de35ec0ca9ddb)

packages/trigger-sdk/src/v3/runs.ts
```diff
@@ -278,6 +293,139 @@ function cancelRun(
...
+): ApiPromise<CreateBulkActionResponseBody> {
+  const apiClient = apiClientManager.clientOrThrow();
+
+  const $requestOptions = mergeRequestOptions(
+    {
+      tracer,
+      name: "runs.bulk.cancel()",
+      icon: "runs",
+      attributes: {
+        ...flattenAttributes(options as Record<string, unknown>, "bulkAction"),
```
carderne, line 308: With AI/scripting, API users could plausibly dump thousands of runIds in here that we pump to otel, maybe we want to truncate/omit these?

apps/webapp/app/routes/api.v1.bulk-actions.$bulkActionId.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const loader = createLoaderApiRoute(
+  {
+    params: ParamsSchema,
+    corsStrategy: "none",
+    authorization: {
+      action: "read",
+      resource: () => ({ type: "runs" }),
+    },
+    // Read from primary so create -> retrieve/poll doesn't 404 on replica lag.
```
carderne: Uses primary to avoid read-your-write issue... but used in SDK `poll()` method... what do we prefer here?

carderne: changed to replica

packages/trigger-sdk/src/v3/runs.ts
carderne: this is now capped to 500

apps/webapp/app/v3/services/bulk/BulkActionV2.server.ts
```diff
@@ -35,22 +48,68 @@ export type ProcessToCompletionResult = {
...
 };
 
+// How recently a PENDING replay must have made progress to still count against
+// the per-environment concurrency limit. Every processed batch bumps the
+// group's `updatedAt`, so a live replay keeps a fresh heartbeat for its whole
+// life no matter how long it runs, while a replay whose job has exhausted its
+// retries (and stopped making progress) ages out and frees its slot. This is
+// wide enough to cover the worst-case gap between batches for a healthy replay
+// that is retrying.
+const REPLAY_INFLIGHT_WINDOW_MS = 30 * 60 * 1000;
```
carderne, line 61: not created as an env var as the value is tied to how often bulk actions heartbeat the updatedAt value so doesn't really make sense to change.

apps/webapp/app/v3/services/bulk/BulkActionV2.server.ts
```diff
@@ -35,22 +48,68 @@ export type ProcessToCompletionResult = {
...
+    // Concurrency guard for replays.
+    // The seek is backed by the (environmentId, status, type) index; the
+    // `updatedAt` window is applied on top so we only count replays that are
+    // actually still making progress. A replay whose job has died stops bumping
+    // `updatedAt` and drops out of the count, so it can't permanently hold a
+    // slot. Aborting a replay (dashboard or API) clears its slot immediately.
+    if (input.action === "replay") {
+      const maxConcurrentReplays = env.BULK_ACTION_MAX_CONCURRENT_REPLAYS;
+      const inFlightReplays = await this._replica.bulkActionGroup.count({
+        where: {
```
carderne, line 91: This checks:
- If there are any existing bulk actions on this environment that are PENDING (these three are on the index)
- AND they have been updatedAt in the last 30 minutes, to prevent stale/inactive bulk actions from blocking new ones. This is not on the index but set will be small by this point

Not 100% sure about this, but thought likely to cause confusion otherwise...

apps/webapp/app/v3/services/bulk/BulkActionV2.server.ts
```diff
@@ -479,28 +542,26 @@ export class BulkActionService extends BaseService {
...
-    to: _to,
-  } = timeFilters({
-    period: filters.period,
-    from: filters.from,
-    to: filters.to,
+  const { period } = timeFilters({
+    period: frozenFilters.period,
+    from: frozenFilters.from,
+    to: frozenFilters.to,
   });
```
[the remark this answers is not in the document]
carderne: updated docs

ericallam (approved):

PR #4110, opened 2026-07-02
feat(cli,core): add opt-in dev-only telnet log streaming

0ski:
Stream dev logs over a local telnet/TCP socket. `trigger dev` mirrors its terminal output on port 6767 by default (override with --telnet-logs-port or TRIGGER_DEV_TELNET_LOGS_PORT, 0 disables). webapp, supervisor, and coordinator each expose an opt-in stream gated on a per-service *_TELNET_LOGS_PORT env var. New @trigger.dev/core/v3/telnetLogServer module (localhost-only, backpressure-safe, plain-text) plus optional static Logger.onLog / SimpleStructuredLogger.onLog sinks.

Then you (or your agent) can use `nc` to connect and filter out the stream.
image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,8 @@ DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres?schema=publi
 # See: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#fields:~:text=the%20shadow%20database.-,directUrl,-No
 DIRECT_URL=${DATABASE_URL}
 REMIX_APP_PORT=3030
+# Dev-only: stream the webapp's logs over a local telnet/TCP socket (nc localhost 6767). Uncomment to enable.
+# WEBAPP_TELNET_LOGS_PORT=6767
 APP_ENV=development
 APP_ORIGIN=http://localhost:3030
 ELECTRIC_ORIGIN=http://localhost:3060
--- a/apps/supervisor/.env.example
+++ b/apps/supervisor/.env.example
@@ -15,3 +15,6 @@ OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:3030/otel
 # Optional settings
 DEBUG=1
 TRIGGER_DEQUEUE_INTERVAL_MS=1000
+
+# Dev-only: stream this process's logs over a local telnet/TCP socket (nc localhost 6769). Uncomment to enable.
+# SUPERVISOR_TELNET_LOGS_PORT=6769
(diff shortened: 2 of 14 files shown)


.env.example
```diff
@@ -8,6 +8,8 @@ DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres?schema=publi
 # See: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#fields:~:text=the%20shadow%20database.-,directUrl,-No
 DIRECT_URL=${DATABASE_URL}
 REMIX_APP_PORT=3030
+# Dev-only: stream the webapp's logs over a local telnet/TCP socket (nc localhost 6767). Unset to disable.
+WEBAPP_TELNET_LOGS_PORT=6767
```
nicktrn: should this be commented out by default?

apps/supervisor/.env.example
```diff
@@ -14,4 +14,7 @@ OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:3030/otel
 
 # Optional settings
 DEBUG=1
-TRIGGER_DEQUEUE_INTERVAL_MS=1000
\ No newline at end of file
+TRIGGER_DEQUEUE_INTERVAL_MS=1000
+
+# Dev-only: stream this process's logs over a local telnet/TCP socket (nc localhost 6769). Unset to disable.
+SUPERVISOR_TELNET_LOGS_PORT=6769
```
nicktrn: should this be commented out by default? what if we don't consume it and there's lots of logs

apps/coordinator/.env.example
```diff
@@ -1,4 +1,6 @@
 HTTP_SERVER_PORT=8020
 PLATFORM_ENABLED=true
 PLATFORM_WS_PORT=3030
-SECURE_CONNECTION=false
\ No newline at end of file
+SECURE_CONNECTION=false
+# Dev-only: stream this process's logs over a local telnet/TCP socket (nc localhost 6770). Unset to disable.
+COORDINATOR_TELNET_LOGS_PORT=6770
```
nicktrn: same

.server-changes/coordinator-telnet-logs.md
nicktrn, line 1: when you get rid of coordinator changes also remove this

nicktrn (approved):

0ski:
@nicktrn applied your feedback, thanks! 🙏

PR #4137, opened 2026-07-03
fix(webapp): db seed script

carderne:
## Summary

`pnpm run db:seed` failed with `SyntaxError: The requested module
'./app/models/organization.server' does not provide an export named
'createOrganization'`, even though that export exists. Renaming the seed entry
point from `seed.mts` to `seed.ts` runs it as CommonJS and fixes the failure.
No seed logic changes.

## Root cause

`seed.mts` is an ES module, but the server modules it imports (`.ts` files, and
no package declares `"type": "module"`) resolve as CommonJS. tsx compiles those
to CommonJS using esbuild's getter-based export shape
(`Object.defineProperty(exports, name, { get })`), which Node's
`cjs-module-lexer` does not detect when it links the ESM importer. The named
exports look absent, so linking throws before any code runs.

The seed only ever needed the `.mts` extension: it has no top-level `await` and
no `import.meta`. Running it as `.ts` keeps the whole import chain CommonJS to
CommonJS and never crosses the lexer boundary. Verified the seed runs to
completion after the change.

Possibly caused by Node 22 upgrade?

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #4135
- <kbd>&nbsp;1&nbsp;</kbd> #4137 👈


1stvamp (approved):
lgtm

PR #4153, opened 2026-07-04
fix(run-store): route caller-passed read clients to the owning store's primary

d-cs:
## The bug

When the run-ops routing store is active (both run-ops DB URLs configured), `RoutingRunStore` accepted a caller-supplied `client` on its read methods but **dropped it**, so every routed read fell back to the owning sub-store's `readOnlyPrisma` (the read **replica**).

The run engine's hot paths pass the **writer/tx** into these reads deliberately, to get read-your-writes. On the dequeue path the engine writes the `QUEUED` snapshot to the primary and reads it back milliseconds later; with the client dropped that read hit the replica. On a single-DB install (or local/CI, where `$replica` falls back to the primary) this is invisible, but against a genuinely separate, lagging replica under write load the read returns a **stale or missing** snapshot — surfacing as `TASK_DEQUEUED_INVALID_STATE` and `No execution snapshot found for TaskRun …`, i.e. runs failing on dequeue.

## The fix

Honor the caller's read-consistency intent, but map it to the **owning sub-store's own primary** (the caller's client is bound to the control-plane DB, which is the wrong database for a NEW-resident row):

- Add `readonly primaryReadClient: ReadClient` to the `RunStore` interface; `PostgresRunStore` returns its writer, `RoutingRunStore` throws (a router has no single primary).
- `#ownPrimary(store, client)` = `client === undefined ? undefined : store.primaryReadClient` — client passed → owning primary; `undefined` → replica (byte-identical to before).
- Applied across the **whole class** of routed reads (`findLatestExecutionSnapshot`, `findExecutionSnapshot`, `findManyExecutionSnapshots`, `findSnapshotCompletedWaitpointIds`, `findRun`/`findRunOrThrow`, the waitpoint read family, batch reads, `findTaskRunAttempt`) plus the cross-DB relation-hydration chain.
- `readYourWrites` now keys on client **presence** rather than writer *identity*, which also fixes tx clients (they lack `$transaction` at runtime) being silently downgraded.

Single-DB / self-host behavior is unchanged (no routing store is built).

## Testing

- New deterministic regression harness (`runOpsStore.routedReadPrimary.test.ts`): two **physically distinct** Postgres containers — writer→DB-A (rows), replica→DB-B (empty = unbounded lag) — so there's no `replica == primary` aliasing to mask the bug. Fails before, passes after, for snapshot reads, fan-out `findRuns`, batch friendlyId, and waitpoint reads; the NEW-arm case asserts a control-plane client resolves to the NEW store's own primary (never forwarded verbatim).
- Full run-store suite: 191 tests pass. Run-engine dequeue/snapshot/waitpoint suites: 30 pass. `@internal/run-store`, `@internal/run-engine`, and `webapp` typecheck clean.

### DIFF
--- a/.server-changes/fix-run-store-routing-read-your-writes.md
+++ b/.server-changes/fix-run-store-routing-read-your-writes.md
@@ -1,0 +1,6 @@
+---
+area: webapp
+type: fix
+---
+
+Fix run store routing dropping the caller's read client, which downgraded read-your-writes reads (execution snapshots, waitpoints, batches) to the read replica and could fail dequeues under replica lag
--- a/apps/webapp/app/db.server.ts
+++ b/apps/webapp/app/db.server.ts
@@ -8,6 +8,7 @@ import {
   type PrismaTransactionOptions,
 } from "@trigger.dev/database";
 import { RunOpsPrismaClient } from "@internal/run-ops-database";
+import { markReadReplicaClient } from "@internal/run-store";
 import invariant from "tiny-invariant";
 import { z } from "zod";
 import { env } from "./env.server";
@@ -170,7 +171,11 @@ export const prisma = singleton("prisma", () =>
 
 export const $replica: PrismaReplicaClient = singleton("replica", () => {
   const replica = getReplicaClient();
-  return replica ? captureInfrastructureErrors(tagDatasource("replica", replica)) : prisma;
+  // Brand ONLY a real replica so the run-store routing layer keeps replica reads off the primary.
+  // No replica configured → fall back to the writer `prisma`, which must stay UNBRANDED.
+  return replica
+    ? markReadReplicaClient(captureInfrastructureErrors(tagDatasource("replica", replica)))
+    : prisma;
 });
 
 export type RunOpsClients = { writer: PrismaClient; replica: PrismaReplicaClient };
@@ -254,9 +259,14 @@ const runOpsTopology: RunOpsTopology = singleton("runOpsTopology", () => {
         captureInfraErrorsRunOps(
           tagDatasourceRunOps("writer", buildRunOpsWriterClient({ url, clientType }))
         ),
+      // Brand the run-ops replica (only built for a real replica URL) so routed replica reads stay
+      // off the primary. When no replica URL is set, selectRunOpsTopology reuses the writer here —
+      // which this callback never touches, so the writer stays unbranded.
       buildNewReplica: (url, clientType) =>
-        captureInfraErrorsRunOps(
-          tagDatasourceRunOps("replica", buildRunOpsReplicaClient({ url, clientType }))
+        markReadReplicaClient(
+          captureInfraErrorsRunOps(
+            tagDatasourceRunOps("replica", buildRunOpsReplicaClient({ url, clientType }))
+          )
         ),
     }
   );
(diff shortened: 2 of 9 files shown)


d-cs:
Thanks both — addressed in cc72e87c7.

## The factual dispute first (Devin's MAJOR): does `$replica` have `$transaction` at runtime?

**Yes.** Verified empirically against the generated Prisma 6.14 client: a base `PrismaClient` exposes `$transaction` as a function, and it survives `.$extends()` (single and double — exactly how `$replica` is built via `captureInfrastructureErrors(tagDatasource("replica", replica))`). So `PrismaReplicaClient` is only a *compile-time* `Omit<PrismaClient, "$transaction">`; the runtime object still has the method.

Consequence: the **pre-PR** `isWriterClient` check (`typeof x.$transaction === "function"`) ALSO returned true for `$replica`, so those four callers were *already* being escalated to the primary before this PR — this specific change was not a new regression for them. BUT the underlying point is right and worth fixing properly: a caller-passed replica should never be escalated to the primary. It also means `$transaction`-presence can't discriminate writer from replica at all (both have it; only a transaction client lacks it), so the old check was doubly wrong — it *failed* to escalate transaction clients while *always* escalating replicas.

## 1. Replica reads no longer escalate to the primary (Devin MAJOR)

Since writer and replica are structurally identical at runtime, I brand replica handles at construction (`markReadReplicaClient`, a `Symbol.for` marker in run-store) and have the router read the brand:
- writer / tx (incl. the engine's dequeue re-read) -> owning store's **primary** (read-your-writes preserved)
- branded replica, or no client -> owning store's **replica** (scaling preserved)

Applied in `db.server.ts` to `$replica` and the run-ops new replica, branding **only genuine replicas** (never the writer, and never the `prisma` fallback when no replica is configured). `readYourWrites` and `#ownPrimary` now escalate only when the client is present **and not** a branded replica. All four flagged callers (spans, trace, sessionRunManager via `$replica`; runEngineHandlers via the read-through `runOpsNewReplica`/`runOpsLegacyReplica`) now read from a replica again.

New test (RED before / GREEN after): a branded replica keeps a routed read on the owning store's replica (misses the lagging replica), while an unbranded writer escalates and finds the fresh row.

## 2. `forWaitpointCompletion` replica probe (Devin minor)

It **is** write-after-read: it selects the store a subsequent `updateManyWaitpoints` lands on. Its two `findWaitpoint` probes now read each store's **primary** (`store.primaryReadClient`), mirroring `#resolveWaitpointStore(onPrimary)`. New test (RED before / GREEN after): under replica lag the owner still resolves correctly instead of mis-routing to the id-shape default and stranding the run.

## 3. Fragile raw SQL in the new test (CodeRabbit)

Replaced the raw `INSERT INTO "_completedWaitpoints" ("A","B")` with the Prisma relation API (`taskRunExecutionSnapshot.update({ data: { completedWaitpoints: { connect: { id } } } })`), so a relation/column rename fails at compile time instead of silently seeding nothing.

## Verification
- run-store: full suite 193/193 green; both new tests confirmed RED on the pre-fix logic, GREEN after.
- run-engine: waitpoint / snapshot / dequeue / residency suites green.
- `typecheck --filter webapp`, `format`, `lint` all clean.

ericallam (approved):

PR #4155, opened 2026-07-05
fix(ci): retry prisma generate to survive Windows EPERM engine-DLL race

d-cs:
## Problem

The `e2e / CLI v3 tests (windows - npm)` job intermittently dies during dependency setup at the `@trigger.dev/database` prisma generate step, before any test runs:

```
@trigger.dev/database:generate: Error: EPERM: operation not permitted, rename
  '...\node_modules\.pnpm\prisma@6.14.0_...\node_modules\prisma\query_engine-windows.dll.node.tmp5744'
  -> '...\node_modules\.pnpm\prisma@6.14.0_...\node_modules\prisma\query_engine-windows.dll.node'
Failed: @trigger.dev/database#generate
```

## Root cause

Confirmed a structural concurrency race, not a random transient. `@trigger.dev/database` and `@internal/run-ops-database` pin the same prisma version, so pnpm gives them a single shared package instance in the store. `turbo run generate` runs their `generate` scripts concurrently, and both `prisma generate` processes race to write the shared `query_engine-windows.dll.node`. On Windows the loser fails with `EPERM` on the `.tmp -> .dll.node` rename because the file was just written/held by the winner.

The CI logs show both `:generate` tasks executing at the same timestamp against the identical shared store path — one finishes, the other EPERMs. In the run that surfaced this, the Windows-pnpm job passed while Windows-npm failed on the same commit, which is the signature of a timing-dependent race.

## Fix

Wrap `prisma generate` in a small cross-platform node script that retries on transient filesystem contention (EPERM/EBUSY/EACCES) with a short backoff. On the retry the engine binary is already present and unlocked, so it succeeds. Non-transient errors (e.g. schema errors) fail fast without retrying.

- Fixing it in the two package `generate` scripts covers every invocation path (all `pnpm run generate` CI jobs plus local dev) without editing many workflow files.
- On non-Windows the first attempt always succeeds, so it is a zero-cost no-op — no serialization, no slowdown to Linux CI.

## Validation

- Verified locally on macOS: the wrapper runs `prisma generate` successfully (exit 0) and does not retry on hard errors.
- The added comment on the e2e generate step brings this PR into the CLI-v3 path filter so the Windows job actually runs here and exercises the fix on the real runner.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -42,6 +42,8 @@ jobs:
       - name: 📥 Download deps
         run: pnpm install --frozen-lockfile --filter trigger.dev...
 
+      # Prisma clients generate concurrently and share one engine binary in the
+      # pnpm store; the generate script retries the transient Windows EPERM race.
       - name: 📀 Generate Prisma Client
         run: pnpm run generate
 
--- a/internal-packages/database/package.json
+++ b/internal-packages/database/package.json
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "clean": "rimraf dist",
-    "generate": "prisma generate",
+    "generate": "node ../../scripts/retry-prisma-generate.mjs",
     "db:migrate:dev:create": "prisma migrate dev --create-only",
     "db:migrate:deploy": "prisma migrate deploy",
     "db:push": "prisma db push",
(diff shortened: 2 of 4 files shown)


d-cs:
Folded into #4154 (cherry-picked `ba796ecfd`) — keeping the Windows CI hardening in the run-ops PR. Closing this standalone PR.

PR #4162, opened 2026-07-05
fix(run-engine): route read-your-writes to the owning store's primary in RoutingRunStore

d-cs:
## Incident fix — run-ops read-your-writes hole

`RoutingRunStore` dropped the caller's writer/tx client on three read-your-writes reads, defaulting to the owning store's **replica** (or, on a single DB, an off-transaction connection). Callers that pass the writer to read back state they just wrote — to beat replica lag — got a stale miss. Result: time-based **DATETIME waitpoints** (`delay` / `wait.until` / `runs.reschedule` / retry backoff) were never armed or were torn down. Surfaces only with the run-ops split active + real read-replicas (the recent replica + pgbouncer rollout is what turned the latent replica-default reads into genuinely lagging reads).

### The 3 sites
Each now routes to the owning store's `*OnPrimary` variant when a writer client is passed (`isWriterClient`), mirroring the existing `findRun`→`findRunOnPrimary` / `findWaitpoint`→`findWaitpointOnPrimary` pattern:
- `findLatestExecutionSnapshot` — reschedule + retry-backoff snapshot read
- `countPendingWaitpoints` — wait blocking check
- `findManyTaskRunWaitpoints` — `continueRunIfUnblocked` un-block read

Adds the three `*OnPrimary` methods to `RunStore` (interface) + `PostgresRunStore`. **Plain reads (no writer) still hit the replica** — no read-offload regression (guarded by a test).

### Test
`runOpsStore.snapshotReadAfterWrite.test.ts` — real split topology (`heteroRunOpsPostgresTest`, prisma14=legacy / prisma17=new) with a lagging-replica proxy. Proves red→green for both LEGACY (cuid) and NEW (ksuid) residencies, plus a guard that plain reads don't escalate to primary.

### Verification
- Fails without the fix (`expected null not to be null` on the writer-client read, both residencies), passes with it — independently reproduced (stash fix → red → restore → green).
- Full `@internal/run-store` suite green (190 tests), incl. existing `readAfterWrite` + `snapshots`.

### Follow-ups (separate)
Same-class gaps to triage: `findWaitpoint` caller in `batchSystem`, and caller-side `tx`-threading in `streamBatchItems.server.ts` / `batchTriggerV3` `tryCompleteBatchV3`.

### DIFF
--- a/internal-packages/run-store/src/PostgresRunStore.ts
+++ b/internal-packages/run-store/src/PostgresRunStore.ts
@@ -1418,6 +1418,16 @@ export class PostgresRunStore implements RunStore {
     });
   }
 
+  // Read-after-write on the OWNING store's PRIMARY: a caller that just wrote a snapshot in this
+  // request re-reads it here so replica lag (findLatestExecutionSnapshot's default) can't miss it.
+  async findLatestExecutionSnapshotOnPrimary(
+    runId: string
+  ): Promise<Prisma.TaskRunExecutionSnapshotGetPayload<{
+    include: { completedWaitpoints: true; checkpoint: true };
+  }> | null> {
+    return this.findLatestExecutionSnapshot(runId, this.prisma as ReadClient);
+  }
+
   /**
    * Dedicated-schema replacement for the legacy `include: { completedWaitpoints: true }` snapshot
    * read. The relation doesn't exist on the subset schema, so we resolve the linked waitpoint ids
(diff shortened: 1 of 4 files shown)


d-cs:
Closing — duplicates the already-merged #4153 (same read-your-writes → owning-store-primary routing fix), and that class is already deployed. Not the cause of the waitpoint incident.

PR #4165, opened 2026-07-06
feat(sdk,core,webapp): offload large batch payloads to object storage

1stvamp:
## Summary

`batchTrigger` and `batchTriggerAndWait` (and the by-id and by-task variants) now offload any per-item payload over 128KB to object storage before sending, the same way single `trigger`/`triggerAndWait` already do since [#3785](https://github.com/triggerdotdev/trigger.dev/pull/3785). A batch of large items no longer inflates the request body past the API limit.

## Demo

A live local run: `batchTriggerAndWait` of 5 items × 300KB (1.5MB total). Each item offloads to object storage, so the receiver run rows hold a 65-byte `application/store` pointer instead of the 300KB body, and every item round-trips (received == sent).

batch large-payload offload demo

## Design

Both the array and streaming batch paths funnel through `executeBatchTwoPhase`, so offloading happens once there: each item is measured, then offloaded through the existing `conditionallyExportPacket` when it crosses 128KB, with bounded concurrency so a big batch doesn't fire an unbounded number of presigned PUTs.

Because items are offloaded before the request, SDK batches arrive as small `application/store` references, so the server-side inline offload during item ingest (parallelised in [#3777](https://github.com/triggerdotdev/trigger.dev/pull/3777)) mostly no longer fires for them.

Every trigger and item also carries its pre-offload serialised size as `options.payloadSize`. The trigger span records that value, so an offloaded payload shows its real size instead of the size of the small object-store reference (previously the span measured the reference).

### DIFF
--- a/.changeset/large-batch-payloads-and-size.md
+++ b/.changeset/large-batch-payloads-and-size.md
@@ -1,0 +1,6 @@
+---
+"@trigger.dev/sdk": patch
+"@trigger.dev/core": patch
+---
+
+Large batch payloads now offload to object storage instead of riding inline in the trigger request. `batchTrigger` and `batchTriggerAndWait` (and the by-id and by-task variants) offload any per-item payload over 128KB before sending, the same way single `trigger` and `triggerAndWait` already do, so a big batch no longer blows past the API body limit.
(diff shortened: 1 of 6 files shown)


1stvamp:
@coderabbitai, on your two findings:

1. Major (wrap offload in `BatchTriggerError`): done. Offload failures now throw with a new `offload` phase (`92343d5`).
2. Nitpick (unit-test the offload branch via `vi.fn()`): holding. Stubbing conflicts with AGENTS.md's "never mock" rule, and the branch is already covered end-to-end by the PR demo.

matt-aitken (approved):

PR #4166, opened 2026-07-06
fix(webapp): record real payload size on the trigger span

1stvamp:
## Summary

When a payload is offloaded to object storage the trigger span only saw the small `application/store` reference, so its `size` attribute measured the reference rather than the payload. It now prefers the caller-reported pre-offload size (`options.payloadSize`) when present, so the span reflects the real payload size. Inline payloads are unchanged.

Stacked on [#4165](https://github.com/triggerdotdev/trigger.dev/pull/4165), which adds the `payloadSize` option this reads.

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #4166 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #4165

### DIFF
--- a/.changeset/large-batch-payloads-and-size.md
+++ b/.changeset/large-batch-payloads-and-size.md
@@ -1,0 +1,6 @@
+---
+"@trigger.dev/sdk": patch
+"@trigger.dev/core": patch
+---
+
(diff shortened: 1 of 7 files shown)


1stvamp:
Folding this into #4165. With the payload-size DB column dropped, this is a one-line span change that only surfaces the value #4165 introduces, so it belongs in the same PR.

PR #4200, opened 2026-07-09
fix(webapp): downgrade retryable directory-sync effect failures to warn

0ski:
Directory-sync effects are idempotent and the accounts-webhook worker retries
the whole event, so a single failed attempt (typically a role assignment
losing a serializable race during a backfill burst) is self-healing rather
than alert-worthy. Tag those thrown errors with logLevel "warn" so the worker
logs at warn instead of error, keeping them visible for triage without paging.

### DIFF
--- a/apps/webapp/app/services/directorySyncEffects.server.ts
+++ b/apps/webapp/app/services/directorySyncEffects.server.ts
@@ -11,6 +11,18 @@ import { createPlatformNotification } from "~/services/platformNotifications.ser
 
 const LAST_OWNER_NOTIFICATION_TITLE = "Directory Sync: last Owner protected";
 
+// Directory-sync effects are idempotent and the accounts-webhook worker retries
+// the whole event (maxAttempts: 5). A single failed attempt is therefore an
+// expected, self-healing condition rather than an alert-worthy error — the most
+// common case is a role assignment losing a serializable race during a backfill
+// burst (the plugin already retries that internally; the worker retry mops up
+// the rest). Tag the thrown error with `logLevel: "warn"` so the worker logs it
+// at warn instead of error (see redis-worker `processItem`), keeping it visible
+// for triage without paging as an error on every transient retry.
+function retryableEffectError(message: string): Error {
+  return Object.assign(new Error(message), { logLevel: "warn" as const });
+}
+
 // Raise a user-scoped, deduped notification when the directory tried to
 // remove the org's last Owner. We keep the member and tell the Owner what to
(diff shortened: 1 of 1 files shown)


apps/webapp/app/services/directorySyncEffects.server.ts
```diff
@@ -11,6 +11,18 @@ import { createPlatformNotification } from "~/services/platformNotifications.ser
...
+// the whole event (maxAttempts: 5). A single failed attempt is therefore an
+// expected, self-healing condition rather than an alert-worthy error — the most
+// common case is a role assignment losing a serializable race during a backfill
+// burst (the plugin already retries that internally; the worker retry mops up
+// the rest). Tag the thrown error with `logLevel: "warn"` so the worker logs it
+// at warn instead of error (see redis-worker `processItem`), keeping it visible
+// for triage without paging as an error on every transient retry.
+function retryableEffectError(message: string): Error {
+  return Object.assign(new Error(message), { logLevel: "warn" as const });
+}
```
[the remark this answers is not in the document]
0ski: Intended behavior. The failure this targets is the transient serializable-conflict retry during a backfill burst (and the cloud-side `setUserRole` now retries 40001/40P01 internally, so most never reach this layer) — exactly what we do not want paging as an error.

Permanent failures are not invisible: they still exhaust retries and land in the DLQ, which every redis-worker exposes via the monitored `redis_worker.queue.dead_letter_size` gauge. So escalation of a genuinely stuck effect happens on DLQ depth, independent of log level — which is the right surface for it. Per-attempt error escalation would need the worker to thread attempt/terminal-ness into the log level, and it is not worth that complexity given the DLQ gauge already covers the case.

apps/webapp/app/services/directorySyncEffects.server.ts
```diff
@@ -11,6 +11,18 @@ import { createPlatformNotification } from "~/services/platformNotifications.ser
 
 const LAST_OWNER_NOTIFICATION_TITLE = "Directory Sync: last Owner protected";
 
+// Directory-sync effects are idempotent and the accounts-webhook worker retries
```
ericallam, line 14: Way too many comments here, we have to be careful about comment bloat

ericallam (approved):

apps/webapp/app/services/directorySyncEffects.server.ts
0ski: yup, let me trim them down

PR #4208, opened 2026-07-09
fix(webapp): schedule run-ops mint-kind flips to avoid duplicate runs

d-cs:
## Summary

Flipping an organization's run-ops mint kind (which database new runs are minted on) took effect independently per process as each cached value expired. For a window after a flip, two concurrent triggers using the same idempotency key could mint on different databases, where the per-database unique constraint can't dedupe them, producing a duplicate run.

## Fix

The flip is now a deterministic wall-clock cutover. The admin feature-flag routes stamp the previously-effective kind and a flip timestamp onto the organization; the mint read resolves the old kind until `flippedAt + grace` (default 90s, chosen to outlast the caches a flip drains through), after which every process crosses to the new kind together. During the window all processes agree on a single database, so a concurrent same-key collision lands on one database and the existing unique-constraint retry returns the one idempotent run. A same-target re-save carries the in-flight stamp forward, so it can't slide the cutover.

### DIFF
--- a/apps/webapp/app/routes/admin.api.v1.orgs.$organizationId.feature-flags.ts
+++ b/apps/webapp/app/routes/admin.api.v1.orgs.$organizationId.feature-flags.ts
@@ -1,9 +1,12 @@
 import type { ActionFunctionArgs, LoaderFunctionArgs } from "@remix-run/server-runtime";
 import { json } from "@remix-run/server-runtime";
+import type { Prisma } from "@trigger.dev/database";
 import { z } from "zod";
+import { env } from "~/env.server";
 import { prisma } from "~/db.server";
 import { requireAdminApiRequest } from "~/services/personalAccessToken.server";
 import { controlPlaneResolver } from "~/v3/runOpsMigration/controlPlaneResolver.server";
+import { stampMintKindFlip } from "~/v3/runOpsMigration/mintFlipGrace";
 import { validatePartialFeatureFlags } from "~/v3/featureFlags";
 
 const ParamsSchema = z.object({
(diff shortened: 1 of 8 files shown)


apps/webapp/app/v3/runOpsMigration/mintFlipGrace.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    const existing = {
+      runOpsMintKind: "runOpsId",
+      runOpsMintKindPrev: "cuid",
+      runOpsMintKindFlippedAt: "not-a-date",
+    };
+    const outgoing = { runOpsMintKind: "runOpsId" };
+    const result = stampMintKindFlip(existing, outgoing, T, GRACE_MS);
+
+    expect(result.runOpsMintKind).toBe("runOpsId");
+  });
```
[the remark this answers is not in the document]
d-cs: Good call. Added in 8d92ce9ae: the test now asserts the malformed `flippedAt` is carried forward verbatim AND that it resolves inert (`Date.parse` yields `NaN`, so `effectiveMintKind` returns the target kind), so it actually validates the "no effective stamp" behavior rather than just the kind.

apps/webapp/app/routes/admin.api.v1.orgs.$organizationId.feature-flags.ts
```diff
@@ -82,18 +85,23 @@ export async function action({ request, params }: ActionFunctionArgs) {
...
-      ...(existingFlags.success ? existingFlags.data : {}),
-      ...validationResult.data,
-    };
+    const mergedFlags = stampMintKindFlip(
+      existingFlags.success ? existingFlags.data : {},
+      {
+        ...(existingFlags.success ? existingFlags.data : {}),
+        ...validationResult.data,
+      },
+      Date.now(),
```
d-cs, line 94: **issue:** Use control plane DB time

apps/webapp/app/routes/admin.api.v2.orgs.$organizationId.feature-flags.ts
```diff
@@ -119,6 +121,17 @@ export async function action({ request, params }: ActionFunctionArgs) {
...
     featureFlags = validationResult.data;
+
+    const existingOrg = await prisma.organization.findFirst({
+      where: { id: organizationId },
+      select: { featureFlags: true },
+    });
+    featureFlags = stampMintKindFlip(
+      existingOrg?.featureFlags as Record<string, unknown> | null | undefined,
+      featureFlags,
+      Date.now(),
```
d-cs, line 132: **suggestion:** use control plane db time

PR #4223, opened 2026-07-10
fix(docs): openapi labels for different bulk api variants

carderne:
Replace Option 1 Option 2 etc with labelled variants.


kathiekiwi (approved):
👍 👍

PR #4239, opened 2026-07-13
chore(webapp): remove unused dependencies and add a knip check

ericallam:
## Summary

Removes 25 unused production dependencies from the webapp (plus 4 orphaned `@types/*` packages and two dead files whose only importers were those packages), and adds a `knip:deps` script so unused dependencies can be found systematically from now on.

## Details

The removed packages are leftovers that nothing in the webapp imports. They were found with [`knip`](https://github.com/webpro-nl/knip), added here as a dev dependency with a `knip:deps` script (`knip --production --dependencies`) and a `knip.json` config. knip's production analysis is workspace-aware and catches dependencies reachable only through otherwise-dead code, which a plain grep can't: both removed deps' only importer files, `SimpleSelect.tsx` and `ulid.server.ts`, were themselves unused.

`non.geist` is ignored in the config: it's a font imported for its side effects via `tailwind.css`, invisible to knip. `socket.io-client` also reports as unused for the webapp but is handled in a separate PR.


nicktrn (approved):

d-cs (approved):

ericallam:
Closing in favor of folding these dependency removals into #4236, where the code that makes them unused is removed. Keeping it to one PR.

PR #4260, opened 2026-07-14
fix(webapp): survive asset hash rotation across rolling deploys

kathiekiwi:
### Problem

Webapp HTML references content-hashed /build assets, and each Docker image contains exactly one build with a hard 404 for unknown hashes. During a rolling deploy, a client holding HTML from the old build may request old asset hashes from a replica running the new image, causing missing styles or failed chunk loads.

The page should recover automatically once a compatible build becomes available, without reload loops or unnecessary interruptions during normal deployments.

### What changed

- Build changes alone do nothing — no polling, no automatic reloads.
- If a CSS or JavaScript asset fails to load, a recovery overlay is shown immediately.
- While the server still reports the same build, the client polls for a newer build using exponential backoff (up to ~60s). As soon as a newer build is detected, the page reloads automatically.
- If no newer build appears within the timeout, recovery falls back to a manual Reload action.
- If recovery still fails after the automatic reload, the client stops retrying and displays a final recovery screen instead of entering a reload loop.
- Recovery preserves form values and scroll position across the automatic reload.


.server-changes/stale-deploy-asset-recovery.md
```diff
@@ -0,0 +1,10 @@
+---
```
carderne, line 1: Nit: changeset seems quite long for this?

carderne (approved):

.server-changes/stale-deploy-asset-recovery.md
kathiekiwi: Machines like the sound of their voice

nicktrn (changes requested):
We need to find a different solution for this that doesn't involve creating Docker images with assets from two different builds. Lots of additional complexity to debug when things go wrong.

.server-changes/stale-deploy-asset-recovery.md
```diff
@@ -0,0 +1,6 @@
...
+area: webapp
+type: fix
+---
+
+Fix intermittent unstyled/broken pages during rolling deploys, caused by stale HTML requesting `/build` asset hashes that 404 on the new image. Documents now default to `Cache-Control: no-cache`, and all responses carry an `X-Build-Id` header. A working page is never touched: when a Remix navigation reveals a new build, it becomes a full document load. Only real breakage (a `/build` asset 404 or failed chunk import) triggers recovery: an overlay goes up, the client polls the new `/build-version` endpoint and reloads once the server reports a different build (form fields and scroll are snapshotted and restored across the reload); if versions never diverge or the reload budget is spent, the overlay offers a manual reload.
```
nicktrn:
```suggestion
Fix pages occasionally loading unstyled or failing to load during deploys. The dashboard now detects this and reloads to recover automatically, or prompts you to reload if it can't.
```

nicktrn (approved):
lgtm, small comment

.server-changes/stale-deploy-asset-recovery.md
nicktrn: It's a bit painful to get these right. We need to force the existing instructions into context maybe. Current locations:
- .server-changes/README.md
- AGENTS.md - section "Changesets and Server Changes", points at above

Claude makes these far too long, they should be short, largely non-technical, user-facing.

kathiekiwi: yeahhh 🤖

PR #4261, opened 2026-07-14
fix(webapp): prevent duplicate envs from provision race

carderne:
Fixes TRI-12078

## Summary

Prevents concurrent environment setup requests from creating duplicate Staging and Preview environments.

## Fix

Adds database-enforced uniqueness for root Staging and Preview environments.

If two requests race, the losing request loads the environment created by the winner and continues successfully instead of creating a duplicate or returning an error.


internal-packages/database/prisma/migrations/20260714140000_runtime_environment_unique_staging_preview_root/migration.sql
```diff
@@ -0,0 +1,4 @@
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "RuntimeEnvironment_projectId_type_staging_preview_root_key"
```
carderne, line 1: Do not merge until this has been backfilled.

matt-aitken (approved):

PR #4278, opened 2026-07-17
revert(webapp): stale-deploy asset recovery (#4260)

nicktrn:
Reverts #4260.

The client-side stale-asset recovery it introduced is net-negative during normal deploys:

- Its `fetch` interception treats any `?_data=` request (Remix loader **and** action traffic) as a navigation and, on a build-id mismatch, `location.assign`es the tab to the fetched URL. During a rolling deploy an open dashboard tab can be hard-navigated to a raw data URL (e.g. a `/resources/*` loader), losing unsaved input.
- Any transient `/build` asset error (a network blip, an extension, an unrelated failed dynamic import) blanks the page behind a full-screen overlay for ~60s before offering a manual reload.
- It serialized form field values to `sessionStorage` to restore them across the reload.

This returns the webapp to the pre-#4260 baseline as a fast, low-risk step — the diff is an exact inverse of #4260 (nothing else has touched those files since). A simpler reload-only recovery, plus a load-balancer stickiness change that addresses the root cause, will follow as separate PRs.


nicktrn:
Superseded by #4280 — redone as a standard `git revert` of #4260 (plain git, canonical revert commit) instead of the hand-rolled version here.

PR #4304, opened 2026-07-20
feat(sdk): return lastEventId from writeTurnComplete and typed capture result

matt-aitken:
## Summary

Two ergonomic additions for custom chat-agent loops that own the turn loop (`chat.customAgent`, `chat.createSession`, and the hand-rolled primitives).

`chat.writeTurnComplete()` now resolves to `{ lastEventId }`, the resume cursor for the start of the next turn. A custom loop can persist it straight from the task instead of round-tripping it back from the client after the turn ends. The value was already produced internally by the turn-complete write; the public wrapper simply discarded it.

`chat.pipeAndCapture()` no longer throws when a stream is stopped or fails. It now resolves to a `PipeAndCaptureResult` carrying any partial `message` captured before the stop or failure, a typed `status` (`"complete" | "aborted" | "error"`), and the `error` on failure. Previously a failed stream threw and the partial was lost, and an abort was captured only when the AI SDK happened to fire `onFinish` in time.

```ts
const { message, status, error } = await chat.pipeAndCapture(result, { signal });
if (message) conversation.addResponse(message);
if (status === "error") logger.error("turn failed", { error });

const { lastEventId } = await chat.writeTurnComplete();
await db.chats.update(chatId, { lastEventId });
```

## Design

`pipeAndCapture` wraps the pipe in a `try/catch` and classifies the outcome from the abort signal (a stop drains the source stream cleanly rather than throwing) versus a thrown error. It also races the `onFinish` capture against a timeout so a hard stop that prevents `onFinish` from firing can't hang the caller. This mirrors the capture path `chat.agent` already uses internally.

The `finishReason` from `onFinish` is surfaced too, since it was already captured on the built-in path.

The internal `turn.complete()` helper keeps its existing contract: it still returns `UIMessage | undefined`, still throws on a genuine stream failure, and still discards output on a full run cancel.

## Breaking change

`chat.pipeAndCapture` previously resolved to `UIMessage | undefined`. Call sites now read `.message` off the result. This is a young, low-level API; the docs examples are updated in this PR.

### DIFF
--- a/.changeset/chat-custom-agent-capture.md
+++ b/.changeset/chat-custom-agent-capture.md
@@ -1,0 +1,22 @@
+---
+"@trigger.dev/sdk": patch
+---
+
+Custom chat agent loops get two ergonomic wins for owning the turn loop.
+
+`chat.writeTurnComplete()` now returns the `lastEventId` of the turn-complete record, so you can persist the next turn's resume cursor straight from the task instead of round-tripping it back from the client.
+
+```ts
+const { lastEventId } = await chat.writeTurnComplete();
+await db.chats.update(chatId, { lastEventId });
+```
+
+`chat.pipeAndCapture()` no longer throws when a stream is stopped or fails. It now returns a `PipeAndCaptureResult` whose `message` holds any partial output captured before the stop or failure, alongside a typed `status` (`"complete" | "aborted" | "error"`) and, on failure, the `error`. Read the message off the result:
+
+```ts
+const { message, status, error } = await chat.pipeAndCapture(result, { signal });
+if (message) conversation.addResponse(message);
+if (status === "error") logger.error("turn failed", { error });
+```
+
+Note: `pipeAndCapture` previously resolved to `UIMessage | undefined`. Update call sites to read `.message` from the returned result.
--- a/.changeset/chat-custom-agent-capture.md
+++ b/.changeset/chat-custom-agent-capture.md
@@ -4,11 +4,11 @@
 
 Custom chat agent loops get two ergonomic wins for owning the turn loop.
 
-`chat.writeTurnComplete()` now returns the `lastEventId` of the turn-complete record, so you can persist the next turn's resume cursor straight from the task instead of round-tripping it back from the client.
+`chat.writeTurnComplete()` now returns the turn boundary's resume cursors (`lastEventId` for the output stream and `sessionInEventId` for the input stream), so you can persist them straight from the task instead of round-tripping them back from the client.
 
 ```ts
-const { lastEventId } = await chat.writeTurnComplete();
-await db.chats.update(chatId, { lastEventId });
+const { lastEventId, sessionInEventId } = await chat.writeTurnComplete();
+await db.chats.update(chatId, { lastEventId, sessionInEventId });
 ```
 
 `chat.pipeAndCapture()` no longer throws when a stream is stopped or fails. It now returns a `PipeAndCaptureResult` whose `message` holds any partial output captured before the stop or failure, alongside a typed `status` (`"complete" | "aborted" | "error"`) and, on failure, the `error`. Read the message off the result:
--- a/docs/ai-chat/pending-messages.mdx
+++ b/docs/ai-chat/pending-messages.mdx
@@ -179,10 +179,14 @@ for (let turn = 0; turn < 100; turn++) {
     stopWhen: stepCountIs(15),
   });
 
-  const { message } = await chat.pipeAndCapture(result);
+  const { message, status, error } = await chat.pipeAndCapture(result);
   sub.off();
 
+  // Keep any partial output, then branch on the outcome: pipeAndCapture no
+  // longer throws, so rethrow a real failure and don't complete a failed turn.
   if (message) await conversation.addResponse(message);
+  if (status === "error") throw error;
+
   await chat.writeTurnComplete();
 }
 ```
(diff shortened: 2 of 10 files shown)


packages/trigger-sdk/src/v3/ai.ts
```diff
@@ -8875,19 +8912,51 @@ async function pipeChatAndCapture(
...
+  }
 
-  return captured;
+  // `onFinish` fires even on abort, carrying the partial — but a hard stop can
+  // prevent it from firing at all, so race it against a timeout to avoid
+  // hanging the caller. Mirrors chat.agent's capture path.
+  await Promise.race([onFinishPromise, new Promise<void>((r) => setTimeout(r, 2_000))]);
+
+  return {
+    message: captured,
```
gtremper, line 9042: Thanks for updating the API here! I ran a review with Codex, and it had this feedback:

> Could we also assert in the hard-error test that the emitted "partial" text is present in result.message? That fixture emits a text delta and then calls controller.error(), but never invokes the supplied onFinish. Since captured is only assigned inside onFinish, it looks like the timeout could return message: undefined even though partial output was streamed. The abort path appears covered; this question is specifically about preserving partial output on the hard-error path.

packages/trigger-sdk/src/v3/ai.ts
```diff
@@ -8828,35 +8828,72 @@ function createStopSignal(): {
...
+ * await db.chats.update(chatId, { lastEventId });
  * ```
  */
-async function chatWriteTurnComplete(options?: { publicAccessToken?: string }): Promise<void> {
-  await writeTurnCompleteChunk(undefined, options?.publicAccessToken);
+async function chatWriteTurnComplete(options?: {
+  publicAccessToken?: string;
+}): Promise<{ lastEventId?: string }> {
+  const result = await writeTurnCompleteChunk(undefined, options?.publicAccessToken);
+  return { lastEventId: result?.lastEventId };
```
gtremper: One more thought while this return type is changing: would it make sense to also return sessionInEventId? That would let a raw loop correlate the turn boundary with the exact session.in record it committed, and it’s already computed for the session-in-event-id header and exposed through the client’s turn-completed event. It isn’t a blocker since Trigger owns input-cursor recovery, but having both cursors server-side would be useful for acknowledgement and out-of-sync detection.

packages/trigger-sdk/src/v3/ai.ts
matt-aitken: Good catch, thanks. Fixed in a6d0ddb: pipeAndCapture now buffers chunks as they flow to the pipe and, if onFinish never fired (the transport-failure path abandons the stream before it finishes), reconstructs the partial from the buffer via readUIMessageStream. The error-path test now asserts the partial text survives. This makes pipeAndCapture slightly more thorough than chat.agent, which stays onFinish-only; happy to extract a shared helper for parity later if we want it.

packages/trigger-sdk/src/v3/ai.ts
matt-aitken: Added in a6d0ddb: writeTurnComplete() now returns { lastEventId, sessionInEventId }. sessionInEventId is the committed .in consume cursor (the same value already written to the session-in-event-id header), so a raw loop can persist both turn-boundary cursors from the task.

packages/trigger-sdk/src/v3/test/test-session-handle.ts
```diff
@@ -147,9 +153,7 @@ export class TestSessionOutputChannel extends SessionOutputChannel {
         }
         resolveDone();
       }
-    })().catch(() => {
-      resolveDone();
-    });
+    })();
```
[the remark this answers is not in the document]
matt-aitken: Fixed in a6d0ddb: the stream setup (ensureReadableStream/getReader) now runs inside the guarded block, so a throw there resolves the completion signal instead of leaving waiters hung.

.changeset/chat-custom-agent-capture.md
```diff
@@ -0,0 +1,22 @@
+---
+"@trigger.dev/sdk": patch
+---
```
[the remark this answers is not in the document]
matt-aitken: Deliberate decision: the changed return shape is on a young, low-level primitive (chat.pipeAndCapture) that nobody using chat.agent hits, so we're shipping it as a patch.

docs/ai-chat/pending-messages.mdx
```diff
@@ -179,10 +179,10 @@ for (let turn = 0; turn < 100; turn++) {
     stopWhen: stepCountIs(15),
   });
 
-  const response = await chat.pipeAndCapture(result);
+  const { message } = await chat.pipeAndCapture(result);
   sub.off();
 
-  if (response) await conversation.addResponse(response);
+  if (message) await conversation.addResponse(message);
```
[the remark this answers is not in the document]
matt-aitken: Applied in a6d0ddb: the example now branches on status and rethrows on error before writeTurnComplete().

packages/trigger-sdk/src/v3/ai.ts
matt-aitken: On chat.agent specifically we're leaving it as-is, because it already preserves the partial on a hard failure through a stronger mechanism than an in-process buffer. It rethrows transport errors instead of swallowing them, and on the next boot it reconstructs the partial from whatever reached session.out (replaySessionOutTail). That survives full process death / OOM, not just a catchable pipe error. pipeAndCapture needs the in-line buffer precisely because it's the raw primitive where the caller owns persistence and there's no automatic replay. Adding the same fallback to chat.agent would mean catching and swallowing transport errors, which disables that recovery path, so we'd rather not touch the proven loop.

myftija (approved):

PR #4387, opened 2026-07-27
feat(core,sdk): support additional environment API keys

carderne:
## Summary

Additional environment API keys can use SDK APIs that require public access tokens. The SDK detects the additional-key format and asks the Trigger.dev server to mint scoped tokens instead of attempting to sign them locally.

Root environment keys retain their existing local-signing behavior. Trigger and batch clients also prefer server-issued tokens returned in response headers while preserving compatibility with older servers.

## Deployment notes

This package update is safe to publish before servers expose additional key creation. Existing root keys continue to use the current path, while an additional key used with an older server fails with an actionable upgrade error.

## Follow-ups

- [ ] Publish the updated `@trigger.dev/core` and `@trigger.dev/sdk` packages before enabling additional key creation.
- [ ] Record the first published SDK version for the dashboard warning and public documentation.

### DIFF
--- a/.changeset/friendly-keys-batch-tokens.md
+++ b/.changeset/friendly-keys-batch-tokens.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/core": patch
+---
+
+Use server-issued public access tokens for batch operations so environment-scoped API keys can read batch results.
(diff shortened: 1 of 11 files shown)


matt-aitken (approved):
Let's update the format so it's `sk` instead of `ak`. Otherwise this LGTM

PR #4396, opened 2026-07-27
perf(run-ops-database): index BatchTaskRun for the batches list on the dedicated schema

ericallam:
## Summary

The batches list page orders by `(createdAt DESC, id DESC)`, which is why [#4361](https://github.com/triggerdotdev/trigger.dev/pull/4361) added a matching index on `BatchTaskRun`. That index only landed in `@trigger.dev/database`.

The dedicated run-ops database has its own migration history, so it never received the index. `BatchListPresenter` reads both databases and merges, so for environments whose batches live in the dedicated database the page kept falling back to a scan and in-memory sort, which is the exact behaviour #4361 set out to fix.

## Fix

Adds the index to the run-ops schema with its own migration. `CREATE INDEX CONCURRENTLY IF NOT EXISTS`, so it is a no-op where the index already exists and still records its ledger row.

The second half is the interesting part. Because the two packages own separate migration histories, a run-graph schema change has to be authored twice, and nothing made the miss visible: the run-ops status check truthfully reports "up to date" against its own history, so the apply step just skips.

`schemaParity.test.ts` compares the physical shape of every model the run-ops schema declares against its counterpart in `@trigger.dev/database`: scalar fields with their attributes, plus `@@index`, `@@unique`, `@@id` and `@@map`. Relation navigation fields are excluded, since the run-ops schema deliberately drops relations that would cross a database boundary while keeping the scalar FK column. A field counts as a relation when its type resolves to a model name, which keeps enum-typed columns in scope.

Two models are listed as run-ops-only: `CompletedWaitpoint` and `WaitpointRunConnection`, both explicit FK-free replacements for a control-plane implicit many-to-many, since an implicit m2m carries a foreign key that cannot resolve across databases. The test also asserts that exception list is exhaustive, so a new unpaired model fails rather than being silently skipped.

Confirmed the guard actually fails: reverting the index turns `BatchTaskRun` red with the missing `@@index` named in the diff.

### DIFF
--- a/internal-packages/run-ops-database/prisma/migrations/20260727120000_add_batch_task_run_created_at_dashboard_index/migration.sql
+++ b/internal-packages/run-ops-database/prisma/migrations/20260727120000_add_batch_task_run_created_at_dashboard_index/migration.sql
@@ -1,0 +1,1 @@
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "BatchTaskRun_runtimeEnvironmentId_createdAt_id_idx" ON "public"."BatchTaskRun"("runtimeEnvironmentId", "createdAt" DESC, "id" DESC);
--- a/internal-packages/run-ops-database/prisma/schema.prisma
+++ b/internal-packages/run-ops-database/prisma/schema.prisma
@@ -827,6 +827,7 @@ model BatchTaskRun {
   @@index([dependentTaskAttemptId])
   // This is for the batch list dashboard page
   @@index([runtimeEnvironmentId, id(sort: Desc)])
+  @@index([runtimeEnvironmentId, createdAt(sort: Desc), id(sort: Desc)])
 }
 
 enum BatchTaskRunStatus {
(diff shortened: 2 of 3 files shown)


d-cs (approved):

ericallam:
Re the CodeRabbit nitpick about co-locating this with the primary schema: keeping it in `@internal/run-ops-database` on purpose.

`internal-packages/database` has no test runner at all, no vitest config, no vitest devDependency, no `test` script and no existing test files, so moving it there means standing up test infrastructure in a package that has none. Here it rides the existing `pnpm run test` and the `internal` CI job.

It is also the right side of the dependency: the control-plane schema is the source of truth and this package is the one carrying the obligation to keep up, so the assertion belongs with the schema that has to follow.

Worth noting it does run on the change that would introduce the bug. Both the `internal` and `webapp` path filters in `pr_checks.yml` include `internal-packages/**`, so editing the control-plane schema triggers this test and `runops-guard` together.

PR #4404, opened 2026-07-28
fix(webapp): stop logging full batch item contents in batchTriggerV3

carderne:
## Summary

The idempotency grouping diagnostic now records task identifiers and item counts rather than full batch items. This keeps the log useful for troubleshooting without including payloads or options.

## Performance notes
Should be better than baseline.


apps/webapp/app/v3/services/batchTriggerV3.server.ts
```diff
@@ -43,6 +43,29 @@ function chunkArray<T>(items: T[], size: number): T[][] {
   }
   return chunks;
 }
+
+// Summarizes a task-identifier -> items grouping for logging, without including the items
+// themselves (each item is `{ task, payload, options }`, and `options.metadata` is arbitrary
+// customer data that must never be written to logs whole).
+export function summarizeItemsByTask(itemsByTask: Record<string, unknown[]>): {
```
ericallam: This adds unnecessary ELU pressure on the server, especially if debug logging is turned off, as the summarzeItemsByTask is called no matter what. Better to just remove the log altogether, or don't do a loop here

ericallam (changes requested):

apps/webapp/app/v3/services/batchTriggerV3.server.ts
carderne: deleted the whole log.

ericallam (approved):

PR #4476, opened 2026-08-02
docs: improve retry discoverability in Errors & Retrying guide

Rohan170603:
## Summary

The **Errors & Retrying** guide demonstrates retry configuration using an OpenAI example, but it doesn't explicitly explain that retries restart the entire `run()` function rather than resuming from the point of failure. This can lead readers to unintentionally repeat non-idempotent operations, such as API calls, emails, or payment requests.

Although this behavior is documented elsewhere, it isn't surfaced from this guide, making it easy to overlook.

## Changes

- Added a warning callout explaining that retries restart the entire `run()` function.
- Linked readers to the idempotency guidance and Durable execution example for preventing repeated side effects.
- Added a note to **Combining tasks** explaining that retrying a parent task also re-triggers child tasks unless an idempotency key is used.

## Testing

- Ran the documentation locally using:
  ```bash
  pnpm --filter "*docs*" dev
  ```
- Verified the **Errors & Retrying** page renders correctly.
- Confirmed the new warning and additional guidance appear as expected.

## Checklist

- [x] I have followed the project's contributing guidelines.
- [x] The change is documentation only.
- [x] I verified the documentation locally before submitting.

### DIFF
--- a/docs/errors-retrying.mdx
+++ b/docs/errors-retrying.mdx
@@ -28,6 +28,19 @@ This task will retry 10 times with exponential backoff.
 
 <OpenaiRetry />
 
+<Warning>
+  Retrying re-runs the entire `run()` function from the start — it does not resume from
+  where the error was thrown. In the example above, if `openaiTask` fails for any reason
+  *after* a successful call to `openai.chat.completions.create()`, the next attempt calls
+  OpenAI again, and you're billed for both calls.
+
+  This applies to any side effect inside `run()` that isn't idempotent, such as sending an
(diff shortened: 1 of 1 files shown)


matt-aitken (approved):

Rohan170603:
Hi @matt-aitken,

Thanks a lot for vouching me!

It looks like this PR was automatically closed before I was vouched. Since I'm now on the vouched list, would you prefer that I reopen this PR or create a new one from the same changes?

Happy to do whichever you prefer. Thanks again!

PR #4484, opened 2026-08-03
test(observability-map): add an unguarded sensitive canary route

1stvamp:
Throwaway. Case 2 of 3 in an end-to-end check of the observability-map CI
comment: adding a sensitive route with no auth guard should be reported as a
regression and should reach the top of FIX FIRST.

Locally verified: new entry, sensitive, score 0, fails auth-boundary,
request-context and audit-trail. Global stays 19, because one route in 412
cannot move the rounded mean; the signal is the row, not the headline.

Do not merge. Will be closed once the CI behaviour is observed.

---
This is **part 3 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #4485
- <kbd>&nbsp;3&nbsp;</kbd> #4484 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #4483
- <kbd>&nbsp;1&nbsp;</kbd> #4455

### DIFF
--- a/apps/webapp/app/routes/api.v1.obsmap-canary.tokens.ts
+++ b/apps/webapp/app/routes/api.v1.obsmap-canary.tokens.ts
@@ -1,0 +1,10 @@
+// SCORER TEST FIXTURE. A throwaway route used to verify that the observability-map CI
+// comment reports a regression when an unguarded sensitive route is added. Never merge.
+import { json } from "@remix-run/server-runtime";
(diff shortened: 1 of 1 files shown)


1stvamp:
Test complete. All four states of the delta rule confirmed on the real runner, plus the stale-comment gap this exposed (fixed in 042c9c3) and its reconcile path. Closing; branch deleted.

PR #4485, opened 2026-08-03
test(observability-map): name the environment on the queues page failure logs

1stvamp:
Throwaway. Case 1 of 3 in an end-to-end check of the observability-map CI
comment: a real improvement should be reported as one.

The two logger.warn calls on the queues page log { error } with no tenant,
while environment is already in scope. Adding environmentId is exactly the
change the tool asks for.

Locally verified: that route goes 0 -> 50, request-context fail -> pass.
Global stays 19, because one route in 412 cannot move the rounded mean.

Do not merge. Will be closed once the CI behaviour is observed.

---
This is **part 4 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #4485 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #4484
- <kbd>&nbsp;2&nbsp;</kbd> #4483
- <kbd>&nbsp;1&nbsp;</kbd> #4455

### DIFF
--- a/apps/webapp/app/db.server.ts
+++ b/apps/webapp/app/db.server.ts
@@ -1,3 +1,4 @@
+// SCORER TEST FIXTURE: non-route edit, outside every watched path. Never merge.
 import {
   Prisma,
   PrismaClient,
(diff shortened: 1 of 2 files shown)


1stvamp:
Test complete. All four states of the delta rule confirmed on the real runner, plus the stale-comment gap this exposed (fixed in 042c9c3) and its reconcile path. Closing; branch deleted.

PR #4499, opened 2026-08-04
fix: prevent nanosecond timestamp overflow in event repository

AkhilTrivediX:
Resolves #3292

Fixes IEEE 754 precision loss that was occurring when multiplying epoch milliseconds by 1,000,000 in float-land before casting to `BigInt`.

The fix properly casts the epoch milliseconds to `BigInt` before multiplying by `BigInt(1_000_000)`, completely preventing the ~256ns errors in timestamp values.

This affects:
- `getNowInNanoseconds()`
- `calculateDurationFromStart()`
- `recordRunDebugLog()`
- Retry event recording in `runEngineHandlers.server.ts`


AkhilTrivediX:
Just realized someone else had already submitted a fix for this (#3381) and is just waiting for a vouch as well. Apologies for the duplicate!

PR #4558, opened 2026-08-10
feat(deployments): install deployed image base packages from a pinned Debian snapshot

myftija:
Every project's image contained a near-duplicate base-packages layer: `apt-get` against the live Debian archive bakes build-time archive state, so identical Containerfiles produced different ~30MB blobs and worker nodes cached one per project. The generated Containerfile now installs base packages from a pinned snapshot.debian.org archive and scrubs apt's timestamped files, making the layer a pure function of the base image and the pinned date: every project converges on one identical blob that nodes cache once.

Verified by building on two independent cold-cache builders, which produced identical digests for every layer.

## Notes

- Default packages now install before user `instructions` and `pkgs`, and user packages go in a separate sorted RUN, so customized projects still share the default layer. Instructions run with ca-certificates and git present and inherit the pinned sources.
- Base package versions freeze at the snapshot date and move forward with CLI releases; a test fails when the pin is over 90 days old, and the build asserts the base image codename matches the pinned suite with an error naming the escape hatch.
- `TRIGGER_BUILD_SKIP_APT_SNAPSHOT=1` falls back to the live archive. Documented in the aptGet extension docs, which now also note that version pins resolve against the snapshot.
- Known limits: cold builds' archive dependency moves from the main Debian CDN to the best-effort snapshot.debian.org service (immutable and heavily CDN-cached, and mitigated by apt retries and the escape hatch, but a weaker SLA); builders that can only reach a private Debian mirror through `image.instructions` need the escape hatch plus their own sources; layers created by extensions' own apt-get calls remain per-project.

### DIFF
--- a/.changeset/deterministic-apt-layer.md
+++ b/.changeset/deterministic-apt-layer.md
@@ -1,0 +1,5 @@
+---
+"trigger.dev": patch
+---
+
+Deployed images now install base system packages from a pinned Debian snapshot archive, making that layer byte-identical across projects and builds. Worker nodes running many projects cache one copy of it instead of a near-duplicate per project, speeding up image pulls. Set TRIGGER_BUILD_SKIP_APT_SNAPSHOT=1 to fall back to the live package archive.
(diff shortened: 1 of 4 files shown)


myftija:
Closing in favor of #4581 and #4602.

PR #4572, opened 2026-08-11
feat: surface cron windows in webapp, cli, sdk

carderne:
## Summary

Adds execution-window product surfaces for both declarative and imperative schedules.

- Declarative schedules can set `window` through `schedules.task()`, with support for whole-minute, hour, and percentage values.
- Imperative schedules can create, update, clear, and inspect windows through the API and dashboard.
- Schedule API responses preserve `nextRun` as the nominal CRON time and expose `nextRunEffectiveAt` as the stable assigned time.
- The dashboard displays configured windows alongside assigned upcoming-run times.
- Deploy output summarizes declarative schedules and suggests adding a wider window when the default 60-second placement range is used.

## Design

Window validation remains authoritative on the server and ensures each window is compatible with the schedule cadence. Omitting a window uses the default 60-second range, while explicit zero-duration windows remain supported.

Deployment summaries are derived from the deployment's stored task metadata, so they reflect the declarations associated with that deployment.


apps/webapp/app/v3/scheduleWindow.server.ts
```diff
@@ -59,20 +67,73 @@ export function formatScheduleWindow({
...
+      schedulePhase: phase,
+      window,
+    }).effectiveAt,
+  }));
+}
+
 export function validateScheduleWindowSyntax(
-  window: ScheduleWindow | undefined
+  window: string | undefined
 ): { valid: true } | { valid: false; message: string } {
```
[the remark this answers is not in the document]
carderne: ratio is already at 100% and not going back

ericallam (approved):

packages/core/src/v3/schemas/api.ts
```diff
@@ -1079,7 +1079,10 @@ export const ScheduleObject = z.object({
   generator: ScheduleGenerator,
   timezone: z.string(),
   window: ScheduleWindow.optional(),
+  /** The next nominal CRON time. */
```
nicktrn, line 1082: one does not simply capitalise cron

nicktrn (approved):

packages/core/src/v3/schemas/api.ts
carderne: haha I had a hard timing getting agent to stop doing this in the UI also

PR #4786, opened 2026-08-26
fix(webapp): integrations page follow-ups — build settings visibility, onboarding spinner, dead code

0ski:
Stacked on #4784. Clears TRI-13488, the follow-ups found while fixing James's two tickets.

## `fix`: show build settings when the GitHub app is disabled

The page wrapped Git settings, the Vercel section **and** build settings in one `githubAppEnabled` guard, so with the GitHub app off it rendered an empty container.

Item 3 turned out narrower than the ticket described. The Vercel section genuinely depends on GitHub — it can't sync environment variables or link deployments without a connected repo, which is what `VercelGitHubWarning` says — so it stays gated. Build settings don't: they also apply to CLI deploys run with `--native-build-server`, exactly as the section's own description states. They now render regardless.

## `fix`: stop the Vercel onboarding modal spinning forever

`computeInitialState` starts in `loading-projects` whenever the org has a Vercel integration but no onboarding data yet, and the effect that escapes it waits for `availableProjects !== undefined`. When `getOnboardingData` returns `null` — it does that on any thrown error, and when the org integration row is missing — nothing ever arrives.

The empty-array case self-resolves (`[] !== undefined`), so this is specifically the null case. The route can tell "still loading" from "loaded nothing" because its fetcher always requests `?vercelOnboarding=true`; it now passes that down and the modal explains the failure with a retry and a link to check the integration's access on Vercel.

## `fix`: match preview environment on slug, not type

`VercelSettingsPresenter` asked for a `PREVIEW`-typed environment with no parent filter, so any preview *branch* row satisfied it — branches are `PREVIEW` rows too. `GitHubSettingsPresenter` and `ProjectSettingsService` ask via the `"preview"` slug, which only the branchable parent carries, so two presenters feeding the same page could disagree. All four checks now match on slug.

## `chore`: remove the remaining dead code

Continues the sweep from the parent branch:

- The `"installing"` `OnboardingState` is unproducible — no `setState` call yields it — so its redirect effect, switch arm, `isLoadingState` conjunct and the `vercelAppInstallPath` import it was the only user of are all dead.
- `(state as string) !== "completed"` sits in a branch where TypeScript has already narrowed `"completed"` out; the cast is what let it compile.
- `hideSectionToggles` was only ever passed alongside `layout="settings"` but only read inside `layout="card"` blocks, so it could never take effect. Removed the prop entirely.
- Unused bindings and the helpers only they referenced: `envSlugLabel`, `_formatSelectedEnvs`, `_CompleteOnboardingForm`, `_handleFinishOnboarding`, and the rest.

No behaviour change in that commit.

## Not included

Per Oskar, the three overlapping modal-open effects in `settings.integrations/route.tsx` are left alone — they're defensive against a close-then-reopen race, and untangling them is a behavioural risk with no user-visible payoff.

## Verification

`pnpm run typecheck --filter webapp`, `pnpm run lint` and `pnpm run knip` are clean, and the four Vercel/project-settings test files pass (39 tests). The one `vite.config.ts` duplicate-`allowedHosts` error in my local output is pre-existing and comes from an unrelated branch in my workspace, not this diff.

No changeset or `.server-changes/` note: the build-settings visibility fix and the spinner fix only bite when the GitHub app is disabled or the Vercel integration has broken, and the rest is internal.

refs TRI-13488

### DIFF
--- a/apps/webapp/app/components/integrations/VercelBuildSettings.tsx
+++ b/apps/webapp/app/components/integrations/VercelBuildSettings.tsx
@@ -49,7 +49,6 @@ type BuildSettingsFieldsProps = {
    * the pin status is unknown — distinct from "not set". */
   currentTriggerVersionFetchFailed?: boolean;
   /** Hide the section-level master toggles for "Pull env vars" and "Discover new env vars". */
-  hideSectionToggles?: boolean;
   showAtomicDeployments?: boolean;
   layout?: "settings" | "card";
 };
@@ -68,7 +67,6 @@ export function BuildSettingsFields({
   onAutoPromoteChange,
   currentTriggerVersion,
   currentTriggerVersionFetchFailed,
-  hideSectionToggles,
   showAtomicDeployments = true,
   layout = "card",
 }: BuildSettingsFieldsProps) {
@@ -222,7 +220,7 @@ export function BuildSettingsFields({
           <div className="mb-2">
             <div className="flex items-center justify-between">
               <Label>Pull env vars before build</Label>
-              {!hideSectionToggles && availableEnvSlugs.length > 1 && (
+              {availableEnvSlugs.length > 1 && (
                 <Switch
                   variant="small"
                   checked={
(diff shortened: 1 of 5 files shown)


0ski:
Folded into #4784 — these four commits now live on that branch, so everything ships as a single PR.

PR #4789, opened 2026-08-26
feat(core,sdk,webapp): version skew protection for chat.agent sessions

0ski:
## Summary

A chat session now carries an external deployment id, so every run it schedules lands on the deployment that matched the app build that started the conversation: the first run, each continuation after an idle suspend, and each recovery after a crash. A chat started by one release of your app keeps talking to the agent build that release shipped with, however many turns that takes.

The id is resolved wherever the session is started, with the same precedence `trigger()` already uses: an explicit `externalDeploymentId`, then `configure()`, then `TRIGGER_EXTERNAL_DEPLOYMENT_ID`, then your platform's commit SHA when `TRIGGER_AUTOMATIC_SKEW_VERSION_PROTECTION` is `1`. **If none of those yields a value, no id is sent and nothing changes**: the chat's runs go to the current version exactly as they do today, and nothing parks.

There is no chat-specific setup either way, so an app already pinning its task runs gets pinned chats with no code change:

```ts
// app/actions.ts
"use server";
import { chat } from "@trigger.dev/sdk/ai";
import type { myChat } from "@/trigger/chat";

export const startChatSession = chat.createStartSessionAction<typeof myChat>("my-chat");
```

Chats need the same two halves as tasks and no more: a deployment carrying an id, and an app that sends the same one. Pass `externalDeploymentId: null` on `triggerConfig` to opt one chat out while the rest of the app stays pinned.

## Pinning to an id you choose

Discovery covers the common case, a commit SHA your platform already exposes. To name the deployment yourself instead, deploy the agent under that id and give your app the same value:

```bash
npx trigger.dev@latest deploy --external-id "$RELEASE_ID"
```

```ts
// app/actions.ts
export const startChatSession = chat.createStartSessionAction<typeof myChat>("my-chat", {
  triggerConfig: { externalDeploymentId: process.env.RELEASE_ID },
});
```

`sessions.start({ triggerConfig: { externalDeploymentId } })` and `AgentChat`'s `triggerConfig` take the same field. Setting `TRIGGER_EXTERNAL_DEPLOYMENT_ID` in the app's environment does the same job without touching code, and an explicit value wins over it.

## Waiting on a deployment

A run pinned to a deployment that has not landed yet parks rather than failing, and messages sent meanwhile are stored and answered once it arrives. That is a strict improvement on `lockToVersion`, which throws on session create and is swallowed on the append path, leaving the chat hung with no run at all. `PENDING_VERSION` is non-final, so the parked run is reused rather than re-triggered.

The session-create and `.in/append` responses now report `pendingVersion`, and the transport emits a matching event, so the wait can read as a deploy in progress instead of a stall:

```tsx
// app/components/chat.tsx
const transport = useTriggerChatTransport<typeof myChat>({
  task: "my-chat",
  accessToken: ({ chatId }) => mintChatAccessToken(chatId),
  startSession: ({ chatId, clientData }) => startChatSession({ chatId, clientData }),
  onEvent: (event) => {
    if (event.type === "run-pending-version") setDeploying(true);
    if (event.type === "first-chunk") setDeploying(false);
  },
});
```

## Escaping the pin

`chat.requestUpgrade()` re-applied the stored pin, which made upgrading away from a pinned version impossible. It now clears the session's id and persists that, so the next continuation cannot bounce back, and takes a target when you want to move to a specific deployment instead:

```ts
// trigger/chat.ts
chat.requestUpgrade({ externalDeploymentId: clientData.commitSha });
```

`lockToVersion` is deliberately untouched: it still wins where both are set, and `requestUpgrade()` still cannot escape it.

Docs: a new Chat sessions section on the version skew protection page, plus notes on the version-upgrades pattern page.

### DIFF
--- a/.changeset/agentchat-forward-trigger-config.md
+++ b/.changeset/agentchat-forward-trigger-config.md
@@ -1,0 +1,5 @@
+---
+"@trigger.dev/sdk": patch
+---
+
+Fixed `AgentChat` silently ignoring `maxDuration`, `region` and `lockToVersion` when they were set on its `triggerConfig`. They are now applied to the session's runs.
--- a/.changeset/chat-agent-version-skew-protection.md
+++ b/.changeset/chat-agent-version-skew-protection.md
@@ -1,0 +1,15 @@
+---
+"@trigger.dev/core": patch
+"@trigger.dev/sdk": patch
+---
+
+Chat sessions now stay on the deployment that matched the app build that started them, so a conversation keeps talking to the agent version its release shipped with across every turn, idle suspend and recovery. The id is resolved wherever you start the session, exactly as it is for `trigger()`, so a chat picks up whatever your app already sends when it triggers a task, with no chat-specific setup. If your app sends no id, nothing changes: chats run on the current version as they do today.
+
+```ts
+// Opt a single chat out of pinning:
+export const startChatSession = chat.createStartSessionAction<typeof myChat>("my-chat", {
+  triggerConfig: { externalDeploymentId: null },
+});
+```
+
+Messages sent while a chat waits on a deployment that is still building are stored and answered once it lands, and the transport emits a `run-pending-version` event so your UI can say so. `chat.requestUpgrade()` now clears the session's pin so the handoff can reach a new version, and accepts `{ externalDeploymentId }` to move to a specific one.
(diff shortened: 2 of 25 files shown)


0ski:
Closing, as I'm going to create a new PR

PR #4791, opened 2026-08-26
feat(sdk): surface a parked agent run to Head Start callers

0ski:
## Summary

A Head Start turn runs across two processes: step 1 in your warm server, step 2 in the agent run. When that run is pinned to a deployment that has not landed yet, it parks, so step 1 streams normally and step 2 arrives late with nothing saying why. `pendingVersion` now reaches every Head Start shape.

The transport learns it from the `headStart` POST and emits the existing event with a new `head-start` source:

```tsx
const transport = useTriggerChatTransport<typeof myChat>({
  task: "my-chat",
  headStart: "/api/chat",
  accessToken: ({ chatId }) => mintChatAccessToken(chatId),
  startSession: ({ chatId, clientData }) => startChatSession({ chatId, clientData }),
  onEvent: (event) => {
    if (event.type === "run-pending-version") setDeploying(true);
    if (event.type === "first-chunk") setDeploying(false);
  },
});
```

The detached flow has no open browser connection to emit into, so it returns the flag:

```ts
const { chatId, pendingVersion, completion } = await chat.startHeadStart({ agentId, chatId, messages, run });
after(completion);
return Response.json({ chatId, deploying: pendingVersion });
```

`chat.handover`'s session handle exposes it too, for callers that build their own response.

## Design

The handler sets `X-Trigger-Chat-Pending-Version: 1` only when the run is parked, next to the `X-Trigger-Chat-*` headers the transport already reads to hydrate. A chat that is not pinned, or is pinned to a live deployment, gets exactly the response it got before.

Nothing about the wait itself changes, and it is gentler than it looks. The handover signal is written to `session.in`, which is durable; a fresh run subscribes from the start of that stream rather than the live tail; and the agent's `idleTimeoutInSeconds` is consumed inside the run at turn 0, so a parked run does not burn its idle budget while parked. A parked turn resumes at step 2 once the deployment lands.

Stacked on [#4789](https://github.com/triggerdotdev/trigger.dev/pull/4789), which adds the session pin and the `run-pending-version` event this extends.


0ski:
Closing, as I'm going to create a new PR

PR #4799, opened 2026-08-27
feat(sdk): ttl for session-triggered runs; expire undequeued dashboard-agent turns

kathiekiwi:
Chat-server sessions can now set a ttl on the runs they trigger, and the dashboard agent uses it: turn runs that no worker dequeues expire after 2 minutes instead of lingering PENDING (where they could later be picked up with an expired user-actor token).

- @trigger.dev/core: optional `ttl` on `SessionTriggerConfig` (same typing as task trigger options).
- @trigger.dev/sdk: both session trigger-config consumers forward it; absent = unchanged behavior.
- webapp: `sessionRunManager` passes it through; only the dashboard-agent path sets it (2m). Expiry is safe for live turns — ttl only affects PENDING, unlocked runs.

### DIFF
--- a/.changeset/chat-session-run-ttl.md
+++ b/.changeset/chat-session-run-ttl.md
@@ -1,0 +1,6 @@
+---
+"@trigger.dev/core": patch
+"@trigger.dev/sdk": patch
+---
+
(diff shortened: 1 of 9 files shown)


kathiekiwi:
Folding this into #4796 — keeping the dashboard-agent work in one PR.

Issue #4963, opened 2026-09-20
Vouch request: kaiizer777 — fix for #4939 (MCP / schema-to-json z.date crash)


kaiizer777:
### Why do you want to contribute?

I recently set up trigger.dev locally and explored the codebase. I investigated and resolved #4939, where converting Zod 4 schemas containing `z.date()` to JSON Schema threw `Date cannot be represented in JSON Schema` because `unrepresentable` was not configured in `convertZod4Schema`.

I implemented the fix with `unrepresentable` mapping date nodes to `{ type: "string", format: "date-time" }` (matching Zod 3 output), added comprehensive unit tests covering date schemas and wrappers across Zod 4 and minimum Zod 3 permalink, added a patch changeset, and verified that all tests (both in `@trigger.dev/schema-to-json` and `trigger.dev` MCP tests) pass cleanly with 0 lint/formatting errors.

PR is prepared at #4962 (branch `fix/zod4-date-json-schema`). I would love to be vouched so the PR can be reviewed and I can continue tackling more issues across the repo!

### Prior contributions or relevant experience

Experienced with TypeScript/JavaScript monorepos, full-stack architecture, and open source development.

PR #4965, opened 2026-09-20
fix(cli): handle catalog: protocol in dependency update check (#3905)

### LINKED ISSUE #3905: bug: CLI crashes with "Invalid comparator: catalog:" when @trigger.dev/* deps use workspace catalogs (bun/pnpm catalog: protocol)

  System:
    OS: Windows 11 10.0.26200
    CPU: (8) x64 Intel(R) Core(TM) Ultra 7 258V
    Memory: 9.90 GB / 31.55 GB
  Binaries:
    Node: 22.20.0 
    npm: 10.9.3
    bun: 1.3.14
    Deno: 2.7.14

### Describe the bug

Running `trigger dev` (or `deploy`, `preview`, `workers build` — anything that runs the embedded update check) crashes immediately when `@trigger.dev/*` dependencies in package.json use the `catalog:` version protocol (bun and pnpm workspace catalogs):

```
TypeError: Invalid comparator: catalog:
    at Comparator.parse (.../semver/classes/comparator.js:41:13)
    ...
    at Module.minVersion (.../semver/ranges/min-version.js:8:11)
    at .../trigger.dev/dist/esm/commands/update.js:71:42
```

### Reproduction repo

See Repo Info Below

### To reproduce

In a bun (or pnpm) workspace, define a catalog in the root package.json and reference it from the app that runs the CLI:

```jsonc
// root package.json
"workspaces": {
  "packages": ["apps/*"],
  "catalog": {
    "@trigger.dev/sdk": "4.4.6",
    "trigger.dev": "4.4.6"
  }
}
```

```jsonc
// apps/trigger/package.json
"dependencies": {
  "@trigger.dev/sdk": "catalog:"
},
"devDependencies": {
  "trigger.dev": "catalog:"
}
```

Then run `trigger dev`.

### Additional information

`getTriggerDependencies()` in `packages/cli-v3/src/commands/update.ts` explicitly skips `workspace:` specifiers, but has no equivalent skip for `catalog:`:

https://github.com/triggerdotdev/trigger.dev/blob/7a54a843e2e9af36515daacb48369b7b1ff3f05c/packages/cli-v3/src/commands/update.ts#L308-L310

When the `tryResolveTriggerPackageVersion()` fallback fails to resolve the installed version from node_modules, the raw `"catalog:"` string is kept as `dep.version` and flows into `semver.minVersion()` in `getVersionMismatches()`, which throws (it's wrapped in neither a try/catch nor a `semver.validRange` guard):

https://github.com/triggerdotdev/trigger.dev/blob/7a54a843e2e9af36515daacb48369b7b1ff3f05c/packages/cli-v3/src/commands/update.ts#L116

## Suggested fix

Skip `catalog:` the same way `workspace:` is skipped:

```ts
if (version.startsWith("workspace") || version.startsWith("catalog:")) {
  continue;
}
```

(Optionally also harden `getVersionMismatches()` — e.g. guard the `semver.minVersion()` call with `semver.validRange(dep.version)` or a try/catch — so any other non-semver specifier degrades to a warning instead of a crash.)

Note: skipping these specifiers entirely does mean catalog users lose the version-mismatch check. A nicer fix would be to keep relying on `tryResolveTriggerPackageVersion()` (which already resolves the real installed version) and only skip when resolution fails — but never letting `catalog:` reach `semver` is the minimal correct change.

## Workaround

`trigger dev --skip-update-check` (and the same flag on `deploy`) bypasses the crashing code path.

kaiizer777:
## Overview

Fixes #3905

In bun and pnpm monorepos utilizing workspace catalogs, dependencies in `package.json` are specified with catalog protocols (e.g. `"@trigger.dev/sdk": "catalog:"` or `"catalog:<catalog-name>"`).

When running CLI commands such as `trigger dev`, `trigger deploy`, or `trigger update`, `updateTriggerPackages()` checks for version mismatches:
1. `getTriggerDependencies()` previously only skipped dependencies starting with `"workspace"`, but did not filter out `"catalog:"` specifiers. If local resolution failed to resolve the local installed package path, `"catalog:"` was retained as the dependency version.
2. In `getVersionMismatches()`, `semver.minVersion(dep.version)` was called directly on `"catalog:"`, which threw an unhandled `TypeError: Invalid comparator: catalog:`, crashing the CLI process immediately.

## Changes

- In `packages/cli-v3/src/commands/update.ts`:
  - Updated `getTriggerDependencies()` to skip `catalog:` dependencies in addition to `workspace` dependencies.
  - Defensively guarded `getVersionMismatches()` with `semver.validRange(dep.version)` and wrapped `semver.minVersion(dep.version)` in a `try...catch` block to gracefully skip non-standard or unresolvable version specifiers without throwing.
  - Exported `Dependency`, `getVersionMismatches`, and `getTriggerDependencies` for direct unit testing.
- Added unit tests in `packages/cli-v3/src/commands/update.test.ts` verifying that `catalog:` and `workspace:` protocols are handled cleanly without throwing, and that `getVersionMismatches()` never crashes on non-semver strings.
- Added patch changeset for `trigger.dev`.

## Note

A vouch request is already open at #4963.


kaiizer777:
Thanks @devin-ai-integration for the thorough review! Both points have been addressed in commit `3a1fe728b`:

1. **Concrete Catalog Version Enforcement & Metadata Retention**:
   - `getTriggerDependencies()` now retains `rawVersion` while attempting resolution of the concrete installed version via `tryResolveTriggerPackageVersion()`. If resolved (e.g. `4.5.0`), the concrete version is used for mismatch detection and downgrade checks, ensuring catalog dependencies don't bypass version enforcement.
   - If resolution fails for an uninstalled/missing package, `catalog:` dependencies gracefully skip downstream crash-inducing evaluation.

2. **Unknown Protocol Mismatch Filtering**:
   - `getVersionMismatches()` now filters out unknown protocols and non-semver specifiers (`!semver.validRange(dep.version) && !semver.valid(dep.version)`), preventing unknown protocols from triggering false mismatches in CI or interactive updates.

3. **Manifest Preservation & Actionable Guidance**:
   - `mutatePackageJsonWithUpdatedPackages()` now inspects `rawVersion` and avoids overwriting `"catalog:"` specifiers in `package.json`.
   - Emits actionable guidance warning the user to update their workspace catalog configuration instead of mangling package manifest references.

All 7 Vitest tests are passing, formatted with `oxfmt`, and validated with `oxlint`.

